PR #334, opened 2023-02-25
Simplify Dockerfile UI using bitnami/nginx V2

alfespa17:
The container for the UI was running as "root" user.

image

The Dockerfile now is using the bitnami/nginx image and is running as user 1001

image

image

The env-config.js should be mount as a volume to load the parameters needed for the react application instead of running the script to generate the file

Example using docker-compose:

```yaml
version: "3.8"
services:
  terrakube-ui:
    image: terrakube-ui:latest
    container_name: terrakube-ui
    volumes:
      - ./env-config.js:/app/env-config.js
    ports:
      - 8080:8080 
```

image

Adding a custom "my_server_block.conf" to fix the react application when using the bitnami/nginx image

```
server {
  listen 0.0.0.0:8080;
  location / {
    root   /app;
    index  index.html index.htm;
    try_files $uri $uri/ /index.html;
  }
}
```


alfespa17:
https://github.com/AzBuilder/terrakube-helm-chart/issues/46

PR #499, opened 2023-09-17
Fix Terraform registry authentication logic

alfespa17:
This will fix the issue with the registry and require authentication to use the following endpoints used by the terraform CLI when using the module protocol.

Search version endpoint:
```bash
{{terrakubeRegistry}}/terraform/modules/v1/{{registryModuleOrg}}/{{registryModuleName}}/{{registryModuleProvider}}
```

Get module.zip URL:
```bash
{{terrakubeRegistry}}/terraform/modules/v1/{{registryModuleOrg}}/{{registryModuleName}}/{{registryModuleProvider}}/{{registryModuleVersion}}
```

Searching without token will return 401 error:
image

image

Searching using a token:
image

image

Using the following will return 401
```terraform
module "iam" { 
  source = "8075-azbuilder-terrakube-cebsv4s4ehg.ws-us104.gitpod.io/aws/iam/aws" 
  version = "v5.30.0" 
  # insert required variables here 
}
```
image

After terraform login
```bash
terraform login 8075-azbuilder-terrakube-cebsv4s4ehg.ws-us104.gitpod.io
Terraform will request an API token for 8075-azbuilder-terrakube-cebsv4s4ehg.ws-us104.gitpod.io using OAuth.

This will work only if you are able to use a web browser on this computer to
complete a login process. If not, you must obtain an API token by another
means and configure it in the CLI configuration manually.

If login is successful, Terraform will store the token in plain text in
the following file for use by subsequent commands:
    /home/user/.terraform.d/credentials.tfrc.json

Do you want to proceed?
  Only 'yes' will be accepted to confirm.

  Enter a value: yes

Terraform must now open a web browser to the login page for 8075-azbuilder-terrakube-cebsv4s4ehg.ws-us104.gitpod.io.

If a browser does not open this automatically, open the following URL to proceed:
    https://5556-azbuilder-terrakube-cebsv4s4ehg.ws-us104.gitpod.io/dex/auth?scope=openid+profile+email+offline_access+groups&client_id=example-app&code_challenge=fynTeeOaNpx6na0tnC4DJy6WaTNq9QnLwjxzWnsEl-g&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A10000%2Flogin&response_type=code&state=cce3bade-0f07-8d6b-45cf-d320efeb9337

Terraform will now wait for the host to signal that login was successful.


---------------------------------------------------------------------------------

Success! Terraform has obtained and saved an API token.

The new API token will be used for any future Terraform command that must make
authenticated requests to 8075-azbuilder-terrakube-cebsv4s4ehg.ws-us104.gitpod.io.
```
With authentication:
```
terraform init

Initializing the backend...
Initializing modules...
Downloading 8075-azbuilder-terrakube-cebsv4s4ehg.ws-us104.gitpod.io/aws/iam/aws 5.30.0 for iam...
- iam in .terraform/modules/iam

Initializing provider plugins...

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

```

### DIFF
--- a/registry/src/main/java/org/terrakube/registry/configuration/authentication/dex/DexWebSecurityAdapter.java
+++ b/registry/src/main/java/org/terrakube/registry/configuration/authentication/dex/DexWebSecurityAdapter.java
@@ -43,6 +43,7 @@ public class DexWebSecurityAdapter extends WebSecurityConfigurerAdapter {
                         .antMatchers("/actuator/**").permitAll()
                         .antMatchers("/terraform/modules/v1/download/**").permitAll()
                         .antMatchers(HttpMethod.OPTIONS, "/**").permitAll()
+                        .anyRequest().authenticated()
                 )
                 .oauth2ResourceServer(oauth2 -> {
                     AuthenticationManagerResolver<HttpServletRequest> authenticationManagerResolver = RegistryAuthenticationManagerResolver


jderieg:
Hi @alfespa17,
I just updated our install to the 2.16.0-beta.2 rev so I could test this. I did indeed get the 401 to our registry when I _don't_ have the token in my credentials.tfrc.json file, so that's great. It's no longer open to the world.
However, I also got the 401 after I did 'terraform login'.

I determined that the reason I'm still getting the error is because we have different URL's set for the API and Registry (similar to what's shown in your docs: https://docs.terrakube.io/getting-started/deployment/ingress-configuration). E.g., 'terrakube-api.example.com' and 'terrakube-reg.example.com'. So when terraform login writes it out to the credentials file, it writes it as 'terrakube-api.example.com'. This causes the registry lookup to continue to fail with 401.

If I manually edit my credentials.tfrc.json file and add another section with the exact same token, but "terrakube-reg.example.com", it works fine. This is an ok workaround, but it would be better if we didn't have to manually edit that credentials file to put in the reg URL with the exact same key that's already in there for the api.

alfespa17:
Hello @jderieg  I am glad that it work for you.

You should be able to run "terraform login REGISTRY" as long as you don't have any token for the registry in the credentials.tfrc.json, once the token in credentials.tfrc.json expire you will start to getting 401 errors, the token expiration time will depend on your Dex configuration.

I get the same error from time to time so I usually delete the token from the credentials.tfrc.json to run the "terraform login" again or create a [PAT](https://docs.terrakube.io/user-guide/organizations/api-tokens) with a 7 days duration, this should work.

You should be able to run Terraform login to the API /Registry because they expose two different type of services, you could check the following endpoints:

Registry expose support to handle modules and providers.
https://REGISTRY_URL/.well-known/terraform.json
image


API expose support to run the "cloud block and remote state" as you can see in this picture
https://API_URL/.well-known/terraform.json
image

This is the reason you will have to tokens inside credentials.tfrc.json, because each endpoint expose different services.

jderieg:
Also, in our Github Actions workflows, we set up Terraform like this to use Terrakube... I'm not sure how this will work with the registry if Terrakube has to have the proper URL set for the token. Seems like it should just use one API key for all authentication, whether it be for the API or the Registry.

```
      # Set up TF on runner
      - name: Setup Terraform
        uses: hashicorp/setup-terraform@v2
        with:
          terraform_version: ${{ env.TF_VERSION }}
          cli_config_credentials_token: ${{ secrets.TERRAKUBE_TOKEN }}
          cli_config_credentials_hostname: "terrakube-api.example.com"
```

alfespa17:
> Also, in our Github Actions workflows, we set up Terraform like this to use Terrakube... I'm not sure how this will work with the registry if Terrakube has to have the proper URL set for the token (I don't think it will). Seems like it should just use one API key for all authentication, whether it be for the API or the Registry.
> 
> ```
>       # Set up TF on runner
>       - name: Setup Terraform
>         uses: hashicorp/setup-terraform@v2
>         with:
>           terraform_version: ${{ env.TF_VERSION }}
>           cli_config_credentials_token: ${{ secrets.TERRAKUBE_TOKEN }}
>           cli_config_credentials_hostname: "terrakube-api.example.com"
> ```

I guess that Github action is manually adding the credentials.tfrc.json, but for terrakube you will need two tokens inside the file

- 1 token if you are using the REGISTRY
- 1 token if you are using "cloud block/remote state"

In that case maybe you can edit the file in your pipeline so you can have 2 tokens

alfespa17:
If you check the source code for the Github action here.

https://github.com/hashicorp/setup-terraform/blob/main/lib/setup-terraform.js#L96

You will see that is creating the terraform.rc file in the home directory with the credentials

alfespa17:
Maybe you could include an additional step in your pipeline with some small script like this:

```
echo "credentials \"TERRAKUBE_REGISTRY\" {" >> $HOME/.terraform.rc
echo "  token = \"xxxxxx\"" >> $HOME/.terraform.rc
echo "}" >> $HOME/.terraform.rc
```

jderieg:
OK, that's an ugly workaround, but that may be the only option.

alfespa17:
Another option could be to create a custom fork for https://github.com/hashicorp/setup-terraform that generate the file with the two tokens

This will require more work.

jderieg:
Yeah, that's probably the route I'll take

PR #516, opened 2023-09-28
fix hard-coded hostname in configuration details

jderieg:
In the 'configuration details' help box on a registry module, there is a hard-coded value of 'app.terrakube.io' for the registry hostname.
This change sets the hostname to the value defined in Terrakube for the Registry URL.


alfespa17 (changes requested):
Hello @jderieg I think you are missing one "}" after the hostname like this:

```
{new URL(window._env_.REACT_APP_REGISTRY_URI).hostname}
```

alfespa17:
Thank you @jderieg 

It looks better now.

image

PR #652, opened 2023-12-28
SSH keys with VCS connection

### LINKED ISSUE #640: Using VCS and SSH connections together for the same workspace
### Feature description 💡

Hello @alfespa17, @jcanizalez;

I would like to briefly summarize what I need. While creating the Workspace, I connected the Terraform repo to Terrakube via the VCS GitHub App. However, in the Terraform repo, the private module pulls my modules by making an SSH connection to the repo. Unfortunately, this is not available in the current Terrakube structure, you have to choose one of the two. However, this is available in Terraform Cloud. Is it possible to add this feature to Terrakube?

For example Terraform Cloud screenshots:
image
image

### Anything else?

_No response_

alfespa17:
This PR will allow to select a SSH key in the workspace setting, this key will be use to download terraform modules using ssh connection.

Adding a new option inside the workspace setting to select which SSH key should be use to download terraform modules like the following example:

```terraform
module "test" {
  source = "git@github.com:alfespa17/simple-terraform.git"
}
```

image

This option can also be use with workspaces with "ssh connection" by default it will be using the SSH key that was selected when creating the workspace but can be used if you want to use a different ssh key for downloading modules.

Fix #640

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
+++ b/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
@@ -27,6 +27,7 @@ public class ExecutorContext {
     private boolean refreshOnly;
     private boolean showHeader;
     private String accessToken;
+    private String moduleSshKey;
     private HashMap<String, String> environmentVariables;
     private HashMap<String, String> variables;
 }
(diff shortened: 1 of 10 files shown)


ilkerispir:
@alfespa17, first all of thank you very much for this feature. I wonder when version 2.19 will be released? And also when will you upgrade Helm Chart version?

alfespa17:
Hello @ilkerispir we are planning to release 2.19 by the end of January, we still want to add  a couple of feature to that release as you can see [here](https://github.com/AzBuilder/terrakube/milestone/20)

I will release a beta if you wan to test it, and you can override the version in the helm chart, we appreciate any feedback

ilkerispir:
@alfespa17, I would be very happy if you released the beta version, I will also upgrade the app version in Helm chart.

alfespa17:
> @alfespa17, I would be very happy if you released the beta version, I will also upgrade the app version in Helm chart.

You could try 2.19.0-beta.3

PR #713, opened 2024-02-15
Redis SSL improvement (+postgres fix)

rogerxaic:
This PR

- add user parameter and possibility to use SSL server with Redis
- changes liquibase defaultSchema parameter to use DatasourceSchema as does the PostgreSQL datasource

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/streaming/StreamingConfiguration.java
+++ b/api/src/main/java/org/terrakube/api/plugin/streaming/StreamingConfiguration.java
@@ -1,13 +1,76 @@
 package org.terrakube.api.plugin.streaming;
 
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.security.KeyStore;
+import java.security.SecureRandom;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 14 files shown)


alfespa17:
Thank you for your contribution @rogerxaic 👍 , I will check the PR later today.

api/src/main/java/org/terrakube/api/plugin/streaming/StreamingConfiguration.java
```diff
@@ -1,27 +1,81 @@
...
     @Bean
-    JedisConnectionFactory jedisConnectionFactory(StreamingProperties streamingProperties) {
-        RedisStandaloneConfiguration redisStandaloneConfiguration = new RedisStandaloneConfiguration(streamingProperties.getHostname(), streamingProperties.getPort());
-        redisStandaloneConfiguration.setPassword(streamingProperties.getPassword());
-        JedisConnectionFactory jedisConFactory = new JedisConnectionFactory(redisStandaloneConfiguration);
+    JedisConnectionFactory jedisConnectionFactory(StreamingProperties props, SSLSocketFactory sslSocketFactory) {
+        RedisStandaloneConfiguration redisStandaloneConfiguration = new RedisStandaloneConfiguration(
+                props.getHostname(), props.getPort());
+        redisStandaloneConfiguration.setPassword(props.getPassword());
+        redisStandaloneConfiguration.setUsername(props.getUsername());
```
dcopestake, line 58: @alfespa17 @rogerxaic Won't this change cause a problem with the default Redis deployment because it means a username will be specified all the time and cause the `RedisHealthIndicator` to fail? This is in relation to this: https://github.com/AzBuilder/terrakube/pull/713#issuecomment-1961261563

PR #746, opened 2024-02-22
Fix default logic to connect to redis

### LINKED ISSUE #733: Registry Cache is failing
### Bug description 🐞

Registry using Redis cache is failing with:

```
2024-02-19 23:04:33.926 ERROR 14854 --- [           main] o.terrakube.api.rs.module.GitTagsCache   : localhost 6379 default password123456
```

The issue is related to this change https://github.com/AzBuilder/terrakube/pull/713

### Steps to reproduce

Open gitpod and check the registry for any organization

image

### Expected behavior

Load all the modules from redis cache correctly

### Example repository

_No response_

### Anything else?

_No response_

alfespa17:
Fix #733


alfespa17:
Hello @rogerxaic I did an small change in the logic that you defined [here](https://github.com/AzBuilder/terrakube/pull/713),
The default Terrakube installation is using [Bitnami Redis Container](https://github.com/bitnami/containers/blob/main/bitnami/redis/README.md) the container only support REDIS_PASSWORD without a username, so I have to remove the "default" username because it was breaking the registry in the UI. 

I think the code to use Redis with sslSocketFactory should continue working but maybe you can take a look when you have some free time because I am not really sure how to test the sslSocketFactory feature correctly

rogerxaic:
Hi @alfespa17, thanks for the heads up! I'll check if that logic works for my use case 😃

PR #751, opened 2024-02-23
Added Terrakube Version Number in UI

jcanizalez:
image


jderieg:
I'm still not seeing the version number at the bottom of the page after upgrade:

PR #960, opened 2024-06-11
fix: the endpoint of api importer

klinux:
With slash in the end of the endpoint, terrakube api returns status code 404


alfespa17:
Thank you, let me create a new release with you fix

PR #1006, opened 2024-06-24
feat: add multiple branch support for VCS

akozhuharov:
This will allow processing of multiple branches using regex matching in the list of branches

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/vcs/WebhookService.java
+++ b/api/src/main/java/org/terrakube/api/plugin/vcs/WebhookService.java
@@ -82,8 +82,16 @@ public class WebhookService {
 
                     // if the webhook is a valid request we can create a new job
                     if(webhookResult.isValid()){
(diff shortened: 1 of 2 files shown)


alfespa17:
Thank you @akozhuharov

alfespa17:
Hello @akozhuharov I missed something you will need to "override" the branch so the request can run in the correct branch there is a field to do that.

https://github.com/AzBuilder/terrakube/blob/81ca458886df6acfcdf3e2bee9bac2800447aae4/api/src/main/java/org/terrakube/api/rs/job/Job.java#L68

https://github.com/AzBuilder/terrakube/blob/81ca458886df6acfcdf3e2bee9bac2800447aae4/api/src/main/java/org/terrakube/api/plugin/vcs/WebhookService.java#L108

akozhuharov:
Added in https://github.com/AzBuilder/terrakube/pull/1010 @alfespa17

PR #1010, opened 2024-06-24
fix: add override branch on job

akozhuharov:
As per https://github.com/AzBuilder/terrakube/pull/1006


alfespa17:
Thank you @akozhuharov

Issue #961, opened 2024-06-12
Allow to configure OpenTofu custom versions and binaries endpoint


FelipeAFV:
### Feature description 💡

Just as you can now create a custom versions endpoint for Terraform, it would be helpful to have the ability to define a custom endpoint for OpenTofu releases. This would allow you to explicitly specify which versions are available in the Terrakube app. Additionally, you would need to define the value in the Terrakube Chart to specify the URL of the custom endpoint, similar to the terraformReleasesUrl value.

alfespa17:
I will add it for the next release, I can add this in the following weeks

alfespa17:
This feature can be tested using version 2.22.0-beta.2

PR #1029, opened 2024-07-03
Allow to configure custom OpenTofu releases URL

gespinozat:
Fixes issue #961


alfespa17:
You will also need to update the executor component to add the custom url so it can download the binary from a different endpoint. 

https://github.com/AzBuilder/terrakube/blob/b810df6ed924ebcf950ae9b7d057ecb66a99aa6a/executor/src/main/resources/application.properties#L8

We are using our terraform Java library the code that added the necessary code was in the following pr where a new property was added as you can see here:

https://github.com/AzBuilder/terraform-spring-boot/pull/45/files

alfespa17:
I have some free time today I will add the missing part for the executor

PR #1082, opened 2024-07-16
feat: #107 send status check back to Github

### LINKED ISSUE #1069: Possible to support commit status on Github
### Feature description 💡

https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28

It would be nice to have the commit status pushed back with a URL to the run

### Anything else?

_No response_

stanleyz:
https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28

Github allows external checks to mark commits with set states, this change does that.

Additional changes:

1. Add a JobVia enum class to unify the value of the `via` field
2. Add a GithubCommitStatus enum class to convert from JobStatus to allowed Github commit status.

fix #1069

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java
+++ b/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java
@@ -13,9 +13,11 @@ import org.terrakube.api.plugin.scheduler.job.tcl.model.Flow;
 import org.terrakube.api.plugin.scheduler.job.tcl.model.FlowType;
 import org.terrakube.api.plugin.scheduler.job.tcl.model.ScheduleTemplate;
 import org.terrakube.api.plugin.softdelete.SoftDeleteService;
+import org.terrakube.api.plugin.vcs.provider.github.GitHubWebhookService;
 import org.terrakube.api.repository.*;
 import org.terrakube.api.rs.job.Job;
 import org.terrakube.api.rs.job.JobStatus;
+import org.terrakube.api.rs.job.JobVia;
 import org.terrakube.api.rs.job.step.Step;
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
@@ -57,6 +59,8 @@ public class ScheduleJob implements org.quartz.Job {
     ScheduleJobService scheduleJobService;
 
     RedisTemplate redisTemplate;
+    
+    GitHubWebhookService gitHubWebhookService;
 
     @Transactional
     @Override
@@ -75,6 +79,7 @@ public class ScheduleJob implements org.quartz.Job {
                 redisTemplate.delete(String.valueOf(job.getId()));
                 log.warn("Deleting Job Context {} from Quartz", PREFIX_JOB_CONTEXT + job.getId());
                 updateJobStepsWithStatus(job.getId(), JobStatus.failed);
+                updateJobStatusOnVcs(job, JobStatus.unknown);
                 removeJobContext(job, jobExecutionContext);
             } catch (Exception e) {
                 log.error(e.getMessage());
@@ -110,6 +115,7 @@ public class ScheduleJob implements org.quartz.Job {
                         completeJob(job);
                         redisTemplate.delete(String.valueOf(job.getId()));
                         updateJobStepsWithStatus(job.getId(), JobStatus.notExecuted);
+                        updateJobStatusOnVcs(job, JobStatus.completed);
                     }
                     break;
                 case approved:
@@ -122,6 +128,7 @@ public class ScheduleJob implements org.quartz.Job {
                     redisTemplate.delete(String.valueOf(job.getId()));
                     removeJobContext(job, jobExecutionContext);
                     ephemeralExecutorService.deleteEphemeralJob(job);
+                    updateJobStatusOnVcs(job, JobStatus.completed);
                     break;
                 case cancelled:
                 case failed:
@@ -129,6 +136,7 @@ public class ScheduleJob implements org.quartz.Job {
                     redisTemplate.delete(String.valueOf(job.getId()));
                     log.info("Deleting Failed/Cancelled/Rejected Job Context {} from Quartz", PREFIX_JOB_CONTEXT + job.getId());
                     updateJobStepsWithStatus(job.getId(), JobStatus.failed);
+                    updateJobStatusOnVcs(job, JobStatus.failed);
                     removeJobContext(job, jobExecutionContext);
                     ephemeralExecutorService.deleteEphemeralJob(job);
                     break;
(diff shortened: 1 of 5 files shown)


stanleyz:
@alfespa17 can you please review? Thanks.

This is only partially working however because the below line re-sets the `commitId` to empty from legit value, is there any particular reason that I should not touch that? With that logic updated, this PR should work as expected.

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/executor/ExecutorJobImpl.java#L45

alfespa17:
The commit id is retrieved here

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/workspace/SetupWorkspaceImpl.java#L147

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/workspace/SetupWorkspaceImpl.java#L229 

So the value that in this place should be valid:

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/executor/ExecutorJobImpl.java#L45

I can see the value here:

image

Not sure when you see that value as invalid

stanleyz:
Yeah, I saw that appear in the UI, but I don't see `SetupWorkspaceImpl.getCommitId` being called for webhook workflow though, without much digging, can you please explain a bit how the `SetupWorkspaceImpl.getCommitId` is getting called?

The commitId is valid for webhook workflow at least when it reaches `ExecutorJobImpl`. It was set from here:

https://github.com/AzBuilder/terrakube/blob/81864245517958c356ecbf009901db5c20315be5/api/src/main/java/org/terrakube/api/plugin/vcs/WebhookService.java#L125

then handled to Executor from `ScheduleJob.java`

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java#L156

But it was reset to empty in the `ExecutorJobImpl.java`

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/executor/ExecutorJobImpl.java#L45

I am thinking to change the logic to try `ExecutorJobImpl.getCommitId` only if the existing commitId is empty or null.

stanleyz:
Actually, I see where the `SetupWorkspace` is called now.  

It is called in `ExecutorJobImpl` line 39 but reset to empty on line 45 for webhook workflow, weird my debugger didn't catch this just now.

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/executor/ExecutorJobImpl.java#L39-L45

alfespa17:
Basically, when the executor clones the repository locally to run terraform it gets the latestcommitHash using the below code

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/workspace/SetupWorkspaceImpl.java#L128

```
private void getCommitId(File gitCloneFolder, String commitId) {
        RevCommit latestCommit = null;
        try {
            if (commitId == null) {
                latestCommit = Git.init().setDirectory(gitCloneFolder).call().
                        log().
                        setMaxCount(1).
                        call().
                        iterator().
                        next();
                String latestCommitHash = latestCommit.getName();
                log.info("Commit Id: {}", latestCommitHash);
                String commitInfoFile = String.format("%s/commitHash.info", gitCloneFolder.getCanonicalPath());
                log.info("Writing commit id to {}", commitInfoFile);
                FileUtils.writeStringToFile(new File(commitInfoFile), latestCommitHash, Charset.defaultCharset());
            } else {
                String commitIdFile = String.format("%s/commitHash.info", gitCloneFolder.getCanonicalPath());
                FileUtils.writeStringToFile(new File(commitIdFile), commitId, Charset.defaultCharset());
            }

        } catch (GitAPIException | IOException e) {
            log.error(e.getMessage());
        }
    }
```

alfespa17:
> Actually, I see where the `SetupWorkspace` is called now.
> 
> It is called in `ExecutorJobImpl` line 39 but reset to empty on line 45 for webhook workflow, weird my debugger didn't catch this just now.
> 
> https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/executor/src/main/java/org/terrakube/executor/service/executor/ExecutorJobImpl.java#L39-L45
> 
> But all in all, don't think that logic is right for webhook workflow at least? Happy to be corrected though.

That is correct I think it should take the information from the webhook commit information instead of getting the information locally

stanleyz:
Right, I see the issue now. The below runs into exception when fetching the code from Git if people use branch name that contains slashes, for example `feat/test1111`, which was what I tend to use, thus the below code didn't create the `commitHash.info` and that causes the the `getCommitId` from `ExecutorJobImpl` to fail.

```
private void getCommitId(File gitCloneFolder, String commitId) {
        RevCommit latestCommit = null;
        try {
            if (commitId == null) {
                latestCommit = Git.init().setDirectory(gitCloneFolder).call().
                        log().
                        setMaxCount(1).
                        call().
                        iterator().
                        next();
                String latestCommitHash = latestCommit.getName();
                log.info("Commit Id: {}", latestCommitHash);
                String commitInfoFile = String.format("%s/commitHash.info", gitCloneFolder.getCanonicalPath());
                log.info("Writing commit id to {}", commitInfoFile);
                FileUtils.writeStringToFile(new File(commitInfoFile), latestCommitHash, Charset.defaultCharset());
            } else {
                String commitIdFile = String.format("%s/commitHash.info", gitCloneFolder.getCanonicalPath());
                FileUtils.writeStringToFile(new File(commitIdFile), commitId, Charset.defaultCharset());
            }

        } catch (GitAPIException | IOException e) {
            log.error(e.getMessage());
        }
    }
```

The bug is introduced here, we should support branch name that contains slashes, which is a common practise nowadays.

https://github.com/AzBuilder/terrakube/blob/e4de8fc6e946162903c0ad7cba7f79b9116e4d37/api/src/main/java/org/terrakube/api/plugin/vcs/provider/github/GitHubWebhookService.java#L49

Additionally, if the commitId is correctly configured already in line 39 of `ExecutorJobImpl`, why do we need to have an additional `getCommitId` again on line 45?

So I'd propose you test this PR, literally it works if there branch name doesn't contain slashes, which I'd look to handle it in a separate PR.

alfespa17:
I don't remember exactly why we check the commit id in line 45 again, I will have to check the executor code I did that part a long time ago.

I will test the PR tomorrow morning, thank you for your help.

alfespa17:
This is working correctly, there is one small issue but I will fix it in a different pull request.

image

I like that when you click the details button is redirected to the job in the ui

PR #1090, opened 2024-07-18
fix: Run from UI failed when workspace is set with multiple branches

stanleyz:
This change addresses it by using the first branch in the list as the one for any runs from UI.


stanleyz:
@alfespa17 free free to change it, I am not quite sure this is the right place to check and update the branch. 

Additionally there should some documentation or UI change to make sure users understand how to use multiple branches.

alfespa17:
Hello @stanleyz you can update the documentation in this page.

https://github.com/AzBuilder/docs/blob/0a3b5db3b3d466ce7aa2acef7f7d25419b780516/user-guide/workspaces/creating-workspaces.md?plain=1

stanleyz:
yeah, I have given it a whirl. 

https://github.com/AzBuilder/docs/pull/10

alfespa17:
Thank you

PR #1092, opened 2024-07-19
fix: #1086 make JWT jti consistent with what's in system

### LINKED ISSUE #1086: Possible to include token ID in the response to create
### Feature description 💡

The token API returns only token, which is fine if people uses the UI, but when we create token programatically, it's hard to index them without the unique ID, yet there doesn't seems another combination in the database that could uniquely identify a token except the ID, for example hard to write a Terraform resource to manage tokens.

https://github.com/AzBuilder/terrakube/blob/150824d031394c3f5ef1d552c7887ccca55525fe/api/src/main/java/org/terrakube/api/plugin/token/team/TeamTokenController.java#L36

### Anything else?

_No response_

stanleyz:
The previous code generates the JWS before saving the token records by then a new key ID is generated, hence causing the ID in the JWT different from what's saved in Terrakube database.

This change addresses this by moving the save operation before the JWS generation and reusing the ID generated by JPA/Spring in the JWT, in an exception, the record in the Terrakube database will be removed.

fixes #1086

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/token/pat/PatService.java
+++ b/api/src/main/java/org/terrakube/api/plugin/token/pat/PatService.java
@@ -29,34 +29,35 @@ public class PatService {
     private PatRepository patRepository;
 
     public String createToken(int days, String description, Object name, Object email, Object groups) {
-
+        String jws = "";
         SecretKey key = Keys.hmacShaKeyFor(Decoders.BASE64URL.decode(this.base64Key));
-        UUID keyId = UUID.randomUUID();
-
-        log.info("Generated Pat {}", keyId);
-
-        String jws = Jwts.builder()
-                .setIssuer(ISSUER)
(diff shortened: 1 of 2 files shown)


alfespa17:
This is related to #781, before that change the classes didn't include the following: 

```
@GeneratedValue(strategy = GenerationType.UUID)
```

That is why the UUID was saved correctly. I think that bug will affect if anyone want to revoke a token from the UI if the token was generated using version 2.21.X because the id in the database will be wrong.

For now reverting removing `@GeneratedValue(strategy = GenerationType.UUID)` or changing the logic for the ID will have the same effect

stanleyz:
Thanks @alfespa17. The UI is fine since it doesn't decrypt the token value to get the ID but simply re-loads the page which reads from the database. 

With this change however, we can write Terraform provider to consistently handle tokens.

Issue #703, opened 2024-02-13
Set VCS Trigger based on path


alfespa17:
In Terraform Cloud and OTF, which I'm migrating from. You could specify the paths that would cause the run to trigger.

We too use a mono repo setup for our different environments. But we also have a modules folder in the root directory, which should also trigger the build, but the working directory isn't in the root.

For example here are the options in OTF and it was also the same in terraform cloud when we used that. This would be amazing if it could be added. It's great it already works for the workspace path!

![image](https://github.com/AzBuilder/terrakube/assets/809944/91f06ee8-2bf8-42ed-8531-92bc6512e4a3)

_Originally posted by @adamlc in https://github.com/AzBuilder/terrakube/issues/654#issuecomment-1938215535_

stanleyz:
@alfespa17 I am also interested in this feature. 

Considering we've already set the branch field to allow multiple branches separated by a comma, I am wondering whether we can do something similar to the folders, i.e. the first folder listed will be used for runs kicked off from UI, and the rest can be a regex or string to match against for runs triggered from VCS webhooks.

I understand it's a bit ugly, but with the combination of multiple branches and multiple folders, the set goal above can be archived.

In the long term, however, I agree with you in other posts that the code for webhook should be refactored, along with the UI. I feel it's better to separate the triggers on webhook from the general settings, because sometimes you want a different template for the triggered runs from webhook, for example only run plan on the triggered runs, but full workflow on the default branch. Basically you have two groups of settings:

* Default settings for the UI triggered runs with the below fields
1.  Branch, single value field
2. Folder, single value field
3. Template

* Settings for webhook-triggered runs
1. Branch, supports multiple values, can use regex
2. Folder, supports multiple values, can use regex
3. Template

adamlc:
Regex would certainly be a good starting point I think!

alfespa17:
I think folder changes could use something similar to this code

https://github.com/AzBuilder/terrakube/blob/a9ecd20f09256ad6adde90069e2a5c9f3802d7b8/api/src/main/java/org/terrakube/api/plugin/vcs/WebhookService.java#L147

I guess adding some validation in this part to check not just the workspace folder but also other trigger folder in the repository could work.

https://github.com/AzBuilder/terrakube/blob/a9ecd20f09256ad6adde90069e2a5c9f3802d7b8/api/src/main/java/org/terrakube/api/plugin/vcs/WebhookService.java#L99

PR #1129, opened 2024-07-27
feat: support trigger by regex patterns 

stanleyz:
This change extends the folder settings on workspace settings by allowing multiple fields separated by comma, the first path will be used as the default workspace path for runs triggered from UI, the rest of the paths are regex to match against the changed files in a push event sent from VCS providers.
    
Example folder path: /workspaces/test,modules/.*.tf

related to #703


alfespa17:
One thing to keep in mind is that the property `folder` in the  `workspace table` is limited to `512 characters`, I think it should work as long as you don't put like a lot of folders

https://github.com/AzBuilder/terrakube/blob/d3c4ceca98d316a99c2c40996116edb508fdd9ea/api/src/main/resources/db/changelog/local/changelog-2.19.0-workspace-folder-length.xml#L10

stanleyz:
Thanks @alfespa17 , I still would like the settings for the triggered run to be split out at some stage so that a different template/workflow can be used for triggered runs. I guess something for later.

PR #1207, opened 2024-08-21
Add private extension repository support

### LINKED ISSUE #928: Support private repository extensions
### Feature description 💡

I created a private terrakube-extensions, and now I need to use it on the executors, though it's showing this error to me:

```
ERROR org.terrakube.executor.service.scripts.ScriptEngineService - https://github.com/MyOrg/terrakube-extensions: Authentication is required but no CredentialsProvider has been registered
```

How do I pass the credentials? I thought the same GitHub App used to make Dex work was also being used here, but it's not.

### Anything else?

Opened as a feature request because idk if it's a bug 😬

alfespa17:
Adding support to use a private extension repository, to use it we just need to add the following environment variable at the organization level.

PRIVATE_EXTENSION_VCS_ID_AUTH

> The above will require the vcs id to use when authentication to the private extension repository

image

For example when using the helm chart we can set the private extension repository like the following:

```yaml
executor:
  properties:
    toolsRepository: "https://myprivate.com/secret/my-private-repository.git"
    toolsBranch: "main"
```

Fix #928

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
+++ b/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
@@ -208,10 +208,10 @@ public class ExecutorService {
             workspaceEnvVariables = dynamicCredentialsService.generateDynamicCredentialsGcp(job, workspaceEnvVariables);
         }
 
-        if (workspaceEnvVariables.containsKey("TERRAKUBE_PRIVATE_EXTENSION_VCS_ID_AUTH")){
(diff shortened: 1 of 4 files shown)


xczdev:
Hi,

Does it use only one repository at a time?
If we set a private extension repository, does it disable the default Terrakube extension, or can both be used simultaneously?
(For some private purposes, I would prefer to use only the private one.)

Thanks !

alfespa17:
> Hi,
> 
> Does it use only one repository at a time? If we set a private extension repository, does it disable the default Terrakube extension, or can both be used simultaneously? (For some private purposes, I would prefer to use only the private one.)
> 
> Thanks !

It only uses one repo at the time

PR #1213, opened 2024-08-26
feat: separate webhook configuration out

stanleyz:
This change updates the webhook functionality and adds three fields in the UI: branch, path and template so that the webhook can be triggered on a set path, branch and execute the configured template.

Additional changes:

1. Replace the deprecated functions/methods in `DexWebSecurityAdapter.java` while `Accept` is added to the allowed header for Elide JSON atomic operations.
2. Replace the deprecated `Where` annotation in `Workspace.java` with the new `SQLRestriction`

This is enhancement to #703


alfespa17:
Interesting I didn't know you can use [atomic operations](https://jsonapi.org/ext/atomic/) with elide

alfespa17:
I think this looks good just one small issue.

When I was testing I select the "Default Template" for VCS push as you can see here:

image

After the creation when I go the the workspace setting the value is empty

image

stanleyz:
> From what I understand these two are different

Yes, they are different. The webhook template is used when a push event is sent from the VCS provider, and the `Default Template` is used for any runs kicked off from the UI.

Thanks for checking, yes, missed the `defaultTemplate` field in the `Create.jsx`, have updated.

Issue #1268, opened 2024-09-04
Wrong terraformWorkingDir when user not type path with leading "/" in workspace configuration


jiin995:
### Bug description 🐞

If user does not write the path with the initial "/" the executor working directory path is wrong and can be not exist.

For example if you set Terraform Working Directory as `environment/test`, the executor run Terraform or OpenTofu in directory: `/home/cnb/.terraform-spring-boot/executor/816a1372-1c53-42b8-a0bd-a1fbd8a191dc/94408cba-17df-4381-a604-b019563bf34aenvironment/test`

### Steps to reproduce

1. Create new workspace
2. Set working directory without initial "/"
3. Run plan
4. Wait plan without changes because working directory it's wrong and check executor log for find bad path.

### Expected behavior

Path correct even if the initial "/" is missing

jiin995:
I create pull request for fix this bug https://github.com/AzBuilder/terrakube/pull/1269

PR #1269, opened 2024-09-04
fix: Wrong terraformWorkingDir when user not type path with leading "/" in workspace configuration (#1268)

jiin995:
Fix issue [#1268](https://github.com/AzBuilder/terrakube/issues/1268)


alfespa17:
Thank you for your help @jiin995

PR #1287, opened 2024-09-12
feat: Correct queue and run prompt for CLI

### LINKED ISSUE #1285: Waiting for -36 queued run(s) to finish before starting...
### Bug description 🐞

The message printed from the local terraform run about queues is wrong.

### Steps to reproduce

Have multiple workspaces, multiple runs at the same time, and 10 jobs waiting for approval in the history

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

stanleyz:
This change takes a parameter of the number of executor replicas and uses that as the capacity for the whole Terrakube instance. This could be improved later on if more different types of agents are supported.

fix #1285

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java
+++ b/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java
@@ -1,15 +1,24 @@
 package org.terrakube.api.plugin.state;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import lombok.extern.slf4j.Slf4j;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.text.ParseException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import org.apache.commons.text.TextStringBuilder;
 import org.quartz.SchedulerException;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.data.domain.Range;
-import org.springframework.data.domain.Range.Bound;
-import org.springframework.data.redis.connection.Limit;
-import org.springframework.data.redis.connection.stream.*;
+import org.springframework.data.redis.connection.stream.MapRecord;
+import org.springframework.data.redis.connection.stream.StreamOffset;
 import org.springframework.data.redis.core.RedisTemplate;
 import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken;
 import org.springframework.stereotype.Service;
@@ -30,7 +39,12 @@ import org.terrakube.api.plugin.state.model.outputs.OutputData;
 import org.terrakube.api.plugin.state.model.outputs.StateOutputs;
 import org.terrakube.api.plugin.state.model.plan.PlanRunData;
 import org.terrakube.api.plugin.state.model.plan.PlanRunModel;
-import org.terrakube.api.plugin.state.model.runs.*;
+import org.terrakube.api.plugin.state.model.runs.ApplyModel;
+import org.terrakube.api.plugin.state.model.runs.Relationships;
+import org.terrakube.api.plugin.state.model.runs.RunEventsModel;
+import org.terrakube.api.plugin.state.model.runs.RunsData;
+import org.terrakube.api.plugin.state.model.runs.RunsDataList;
+import org.terrakube.api.plugin.state.model.runs.RunsModel;
 import org.terrakube.api.plugin.state.model.state.StateData;
 import org.terrakube.api.plugin.state.model.state.StateModel;
 import org.terrakube.api.plugin.state.model.terraform.TerraformState;
@@ -42,7 +56,16 @@ import org.terrakube.api.plugin.state.model.workspace.state.consumers.StateConsu
 import org.terrakube.api.plugin.state.model.workspace.tags.TagDataList;
 import org.terrakube.api.plugin.storage.StorageTypeService;
 import org.terrakube.api.plugin.token.team.TeamTokenService;
-import org.terrakube.api.repository.*;
+import org.terrakube.api.repository.ArchiveRepository;
+import org.terrakube.api.repository.ContentRepository;
+import org.terrakube.api.repository.HistoryRepository;
+import org.terrakube.api.repository.JobRepository;
+import org.terrakube.api.repository.OrganizationRepository;
+import org.terrakube.api.repository.StepRepository;
+import org.terrakube.api.repository.TagRepository;
+import org.terrakube.api.repository.TemplateRepository;
+import org.terrakube.api.repository.WorkspaceRepository;
+import org.terrakube.api.repository.WorkspaceTagRepository;
 import org.terrakube.api.rs.Organization;
 import org.terrakube.api.rs.job.Job;
 import org.terrakube.api.rs.job.JobStatus;
@@ -56,11 +79,10 @@ import org.terrakube.api.rs.workspace.history.archive.Archive;
 import org.terrakube.api.rs.workspace.history.archive.ArchiveType;
 import org.terrakube.api.rs.workspace.tag.WorkspaceTag;
 
-import java.io.InputStream;
-import java.nio.charset.StandardCharsets;
-import java.text.ParseException;
-import java.util.*;
-import java.util.concurrent.atomic.AtomicBoolean;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 @Service
@@ -76,7 +98,9 @@ public class RemoteTfeService {
     private String hostname;
     private StorageTypeService storageTypeService;
     private StepRepository stepRepository;
+    @SuppressWarnings("rawtypes")
     private RedisTemplate redisTemplate;
+    private int executorCount;
 
     private TagRepository tagRepository;
 
@@ -96,7 +120,8 @@ public class RemoteTfeService {
             @Value("${org.terrakube.hostname}") String hostname,
             StorageTypeService storageTypeService,
             StepRepository stepRepository,
-            RedisTemplate redisTemplate,
+            @SuppressWarnings("rawtypes") RedisTemplate redisTemplate,
+            @Value("${org.terrakube.executor.replicas}") int executorCount,
             TagRepository tagRepository,
             WorkspaceTagRepository workspaceTagRepository,
             TeamTokenService teamTokenService,
@@ -112,6 +137,7 @@ public class RemoteTfeService {
         this.storageTypeService = storageTypeService;
         this.stepRepository = stepRepository;
         this.redisTemplate = redisTemplate;
+        this.executorCount = executorCount;
         this.tagRepository = tagRepository;
         this.workspaceTagRepository = workspaceTagRepository;
         this.teamTokenService = teamTokenService;
@@ -194,16 +220,8 @@ public class RemoteTfeService {
             orgCapacityData.getData().setAttributes(new OrgCapacityAttributes());
             orgCapacityData.getData().getAttributes().setPending(0);
 
-            List<Job> jobList = jobRepository.findAllByOrganizationAndStatusNotInOrderByIdAsc(
-                    organization,
-                    Arrays.asList(
-                            JobStatus.failed,
-                            JobStatus.completed,
-                            JobStatus.rejected,
-                            JobStatus.cancelled,
-                            JobStatus.approved));
             orgCapacityData.getData().setType("organization-capacity");
-            orgCapacityData.getData().getAttributes().setRunning(jobList.size());
+            orgCapacityData.getData().getAttributes().setRunning(executorCount);
             log.info("orgCapacityData: {}", orgCapacityData);
             return orgCapacityData;
         } else {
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -929,41 +949,16 @@ RunsData getRun(int runId, String include) {
...
                 Arrays.asList(
-                        JobStatus.failed,
-                        JobStatus.completed,
-                        JobStatus.rejected,
-                        JobStatus.cancelled,
-                        JobStatus.approved));
-        int runQueue = 0;
+                        JobStatus.pending,
+                        JobStatus.running,
+                        JobStatus.queue));
```
alfespa17, line 959: I think you need to add JobStatus.waitingForApproval here

api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -981,34 +976,20 @@ RunsDataList getRunsQueue(String organizationName) {
...
-                }
-            }
+        if (!workspaceList.isPresent()) {
+            return runsDataList;
         }
+        Workspace workspace = workspaceList.get();
+        jobRepository.findAllByWorkspaceAndStatusInOrderByIdDesc(workspace, Arrays.asList(
+                JobStatus.pending,
+                JobStatus.running,
+                JobStatus.queue)).forEach(job -> {
```
alfespa17, line 988: you also need to add waiting for aproval here too

alfespa17:
I tested this and I found a little issue for example if I run a run two terraform commands the second command should be in waiting state until the first is completed example:

## First terraform apply 

This operation is running fine but once terraform asked for confirmation all the other jobs should be waiting in the queue.

```
user@pop-os:~/git/simple-terraform$ terraform apply
Running apply in HCP Terraform. Output will stream here. Pressing Ctrl-C
will cancel the remote apply if it's still pending. If the apply started it
will stop streaming the logs, but will not stop the apply running remotely.

Preparing the remote apply...

To view this run in a browser, visit:
https://terrakube-api.minikube.net/app/simple/locked-remote/runs/run-139

Waiting for the plan to start...

***************************************
Running Terraform PLAN
***************************************

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # null_resource.next will be created
  + resource "null_resource" "next" {
      + id = (known after apply)
    }

  # null_resource.next2 will be created
  + resource "null_resource" "next2" {
      + id = (known after apply)
    }

  # null_resource.next3 will be created
  + resource "null_resource" "next3" {
      + id = (known after apply)
    }

  # null_resource.next4 will be created
  + resource "null_resource" "next4" {
      + id = (known after apply)
    }

  # null_resource.previous will be created
  + resource "null_resource" "previous" {
      + id = (known after apply)
    }

  # time_sleep.wait_30_seconds will be created
  + resource "time_sleep" "wait_30_seconds" {
      + create_duration  = "2m"
      + destroy_duration = "45s"
      + id               = (known after apply)
    }

  # module.time_module.random_integer.time will be created
  + resource "random_integer" "time" {
      + id     = (known after apply)
      + max    = 5
      + min    = 1
      + result = (known after apply)
    }

Plan: 7 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + creation_time = "2m"
  + fake_data     = {
      + data     = "Hello World"
      + resource = {
          + resource1 = "fake"
        }
    }

Do you want to perform these actions in workspace "locked-remote"?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: 
```

## Second Terraform command

For example I ran a second operation at the same time, the terraform command was locked and waiting  as you can see below, but once the first job asked for confirmation this plan was executed and it should be waiting for the first one to be completed, failed or cancelled.

```
user@pop-os:~/git/simple-terraform$ terraform apply
Running apply in HCP Terraform. Output will stream here. Pressing Ctrl-C
will cancel the remote apply if it's still pending. If the apply started it
will stop streaming the logs, but will not stop the apply running remotely.

Preparing the remote apply...

To view this run in a browser, visit:
https://terrakube-api.minikube.net/app/simple/locked-remote/runs/run-140

Waiting for -7 queued run(s) to finish before starting...
Waiting for -7 queued run(s) to finish before starting... (30s elapsed)
Waiting for -7 queued run(s) to finish before starting... (1m0s elapsed)

***************************************
Running Terraform PLAN
***************************************

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # null_resource.next will be created
  + resource "null_resource" "next" {
      + id = (known after apply)
    }

  # null_resource.next2 will be created
  + resource "null_resource" "next2" {
      + id = (known after apply)
    }

  # null_resource.next3 will be created
  + resource "null_resource" "next3" {
      + id = (known after apply)
    }

  # null_resource.next4 will be created
  + resource "null_resource" "next4" {
      + id = (known after apply)
    }

  # null_resource.previous will be created
  + resource "null_resource" "previous" {
      + id = (known after apply)
    }

  # time_sleep.wait_30_seconds will be created
  + resource "time_sleep" "wait_30_seconds" {
      + create_duration  = "2m"
      + destroy_duration = "45s"
      + id               = (known after apply)
    }

  # module.time_module.random_integer.time will be created
  + resource "random_integer" "time" {
      + id     = (known after apply)
      + max    = 5
      + min    = 1
      + result = (known after apply)
    }

Plan: 7 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + creation_time = "2m"
  + fake_data     = {
      + data     = "Hello World"
      + resource = {
          + resource1 = "fake"
        }
    }
```

I think that if you add the "waitingForApproval" status to the logic that you have updated that should fix it.

alfespa17:
This is an example using TFC.

## First terraform apply

This is waiting for confirmation after the plan is completed
```
user@pop-os:~/git/simple-terraform$ terraform apply
Running apply in HCP Terraform. Output will stream here. Pressing Ctrl-C
will cancel the remote apply if it's still pending. If the apply started it
will stop streaming the logs, but will not stop the apply running remotely.

Preparing the remote apply...

To view this run in a browser, visit:
https://app.terraform.io/app/alfespa17/locked-remote/runs/run-TDhGFxb4oEPJEvha

Waiting for the plan to start...

Terraform v1.9.5
on linux_amd64
Initializing plugins and modules...

Terraform used the selected providers to generate the following execution plan.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # null_resource.next will be created
  + resource "null_resource" "next" {
      + id = (known after apply)
    }

  # null_resource.next2 will be created
  + resource "null_resource" "next2" {
      + id = (known after apply)
    }

  # null_resource.next3 will be created
  + resource "null_resource" "next3" {
      + id = (known after apply)
    }

  # null_resource.next4 will be created
  + resource "null_resource" "next4" {
      + id = (known after apply)
    }

  # null_resource.previous will be created
  + resource "null_resource" "previous" {
      + id = (known after apply)
    }

  # time_sleep.wait_30_seconds will be created
  + resource "time_sleep" "wait_30_seconds" {
      + create_duration  = "2m"
      + destroy_duration = "45s"
      + id               = (known after apply)
    }

  # module.time_module.random_integer.time will be created
  + resource "random_integer" "time" {
      + id     = (known after apply)
      + max    = 5
      + min    = 1
      + result = (known after apply)
    }

Plan: 7 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + creation_time = "2m"
  + fake_data     = {
      + data     = "Hello World"
      + resource = {
          + resource1 = "fake"
        }
    }

Do you want to perform these actions in workspace "locked-remote"?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value:
```

## Second Terraform apply 

it is being queue and waiting for the first to be completed:

```
user@pop-os:~/git/simple-terraform$ terraform apply
Running apply in HCP Terraform. Output will stream here. Pressing Ctrl-C
will cancel the remote apply if it's still pending. If the apply started it
will stop streaming the logs, but will not stop the apply running remotely.

Preparing the remote apply...

To view this run in a browser, visit:
https://app.terraform.io/app/alfespa17/locked-remote/runs/run-AantdC2opyRBsbqV

Waiting for 1 run(s) to finish before being queued...
Waiting for 1 run(s) to finish before being queued... (30s elapsed)
Waiting for 1 run(s) to finish before being queued... (1m0s elapsed)
Waiting for 1 run(s) to finish before being queued... (1m30s elapsed)

```

api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java
stanleyz: Don't think you need it here because this function is used to calculate the capacity of the runners to see when the Job can be queued.

api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java
stanleyz: This worth a bit discussion. Technically yes the next run should queue if the previous run(s) are pending on approval for apply.

TFC works differently than Terrakube. TFC doesn't have the idea of workflow as Terrakube, so it only really has two types of operations: Plan and Apply, hence it's easy to get the operation type and deduce whether the current run should be queued upon the previous job based on the operation type, basically if previous job type is apply, then waiting to be queued, if previous job is plan, then let it go.

https://github.com/hashicorp/terraform/blob/0ae6bc34c40f5a96b36c43c18f098d30df40ff0c/internal/cloud/backend_common.go#L136

Terrakube uses workflows which could potentially leads to a large number of jobs left in "WaitingForApproval" state, especially when used with webhook previously that doesn't a dedicated workflow/template, meaning the workspace and webhook could share the same "Plan->Approval->Apply" template. Yes, this situation improves once the webhook has its own dedicated template, but how do we deal with historical jobs that are in "WaitingForApproval" state? If we add "WaitingForApproval" state to the above `getWorkspaceRuns` function, all new jobs will be sitting to be queued indefinitely so long as there is a "WaitingForApproval" job in the history, no matter old or new. 

As far as I can see, there seems two ways to deal with this:

1. Force the customer to cancel all previous jobs that's in "WaitingForApproval" state, not too bad actually especially with webhook now using dedicated template.
2. Write a Liquid script to update all previous job in a upgrade
3. Do something smart in code, potentially
     1. A quartz job to cancel all "WaitingForApproval" jobs that are old than say 24 hours, (you might have that already in code somewhere)
     2. A function to check and cancel all previous "WaitingForApproval" jobs so long as a new "Apply" task is scheduled. 
     
I feel that item 2 is good for short-term, and option i in item 3 could potentially be adopted for long-term use. What do you think? @alfespa17

alfespa17: I had some logic to cancel all jobs that have been running for 6 hours in the code in this part of the code

https://github.com/AzBuilder/terrakube/blob/9b45945f3e9a2b509a91b440052831af330519ba/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java#L74

That code is no longer working now because in the past a quartz job was executed every 1 minute to validate the job status and continue the execution (terrakube was really slow during execution becuase of that), but I changed that logic some months ago so a quartz job is created dynamically when the Job entity is updated using the Elide hooks here:

https://github.com/AzBuilder/terrakube/blob/9b45945f3e9a2b509a91b440052831af330519ba/api/src/main/java/org/terrakube/api/rs/hooks/job/JobManageHook.java#L37

This is the reason why jobs in "waitingForApproval" state are no longer cancel after 6 hours...... maybe we will have to create some separete quartz job that fix it

So in theory once we fix that validation for jobs shouldn't be running for more that 6 hours

alfespa17: By the way calculating the capacity based on the executor replicas could partially work because there is an option to have ephemeral executors like explained here but it is a good starting point.

https://docs.terrakube.io/getting-started/deployment/ephemeral-agents

alfespa17: I added the logic to cancel long running jobs in a separate PR [here](https://github.com/AzBuilder/terrakube/pull/1289/files)

stanleyz: Cool, I've added "WaitingForApproval" to the status query as suggested.

> That could be an option but I am not sure how the terraform CLI will handle that because if I remember correctly the CLI is only waiting for the user to answer yes/no to run the apply and it is not actively checking the job status

No, it's not and the Job will be printed to "Waiting to be queued", there should be documentation/knowledge some where to let customer know that you need to have all "WaitingForApproval" cancelled/finished. 

> I added the logic to cancel long running jobs in a separate PR [here](https://github.com/AzBuilder/terrakube/pull/1289/files)

That's great, thanks. I feel running the scheduled task every 5 mins is too frequent for jobs that will only be updated every 6 hours, even running the schedule task every 1 or 2 hours is helpful. As you have mentioned, it could make Terrakube slow, actually my team has complained many times about this, wondering whether some profiling should be done.

Additionally, correct me if I am wrong, the jobs sent over executor/runner hold the access token which only is valid for 1 or 2 hours,  out of the four job status, "pending" and "queued" jobs won't run successfully anyway?

alfespa17: You are correct the job queue logic can be improved, I will try to think how to improved it in the following days, I already did some progress in the latest release triggering the job directly once the job status is updated instead of waiting for 1 minute, maybe I can improve it some more for 2.23.0.

Regarding the schedule to cancel long running jobs maybe it is better to add some parameter so anyone can customize the value.

For your last question the access token could be valid for 1 or 2 hours but API sent all the information to the executor each time it executes a template "step" for example in the following 

```yaml
flow:
  - name: "Plan"
    type: "terraformPlan"
    step: 100
  - name: "Apply"
    type: "terraformApply"
    step: 200
```

Terrakube will send all the information to the executor so it can execute the "plan" and mark the "plan" as completed, later it will send all of the information for the "apply" and the executor will mark the "apply" as completed.

In a high level the job status will be moving like this:

```
Pending (plan) -> Queue (plan) -> Running (plan) -> Pending (apply) -> Queue (apply)  -> Running (apply)  -> Pending (other step if need it) -> Completed (No other steps to execute)
```

api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -314,8 +332,8 @@ WorkspaceData getWorkspace(String organizationName, String workspaceName, Map<St
             workspaceModel.setAttributes(attributes);
             workspaceData.setData(workspaceModel);
 
-            Optional<Job> currentJob = jobRepository.findFirstByWorkspaceAndAndStatusInOrderByIdAsc(workspace.get(),
-                    Arrays.asList(JobStatus.pending, JobStatus.running, JobStatus.waitingApproval));
+            Optional<Job> currentJob = jobRepository.findFirstByWorkspaceAndStatusInOrderByIdAsc(workspace.get(),
+                    Arrays.asList(JobStatus.pending, JobStatus.running, JobStatus.queue));
```
alfespa17, line 336: This line is missing waitingForApproval state

alfespa17: I will add it manually I think the rest looks good to me

PR #1313, opened 2024-09-22
feat: Add manageState permission

### LINKED ISSUE #1245: Additional role in org (to protect the state)
### Feedback

Currently, a user in an organization needs to have `Manage Workspace` permission to create new jobs, aka. run Terraform, but once the user has this permission, who can change all the settings of the workspace, is it possible to have a separate role to allow users to "use" the workspace but not manage it? Operations that are potentially in the use category could include:

1. Run Terraform from CLI
2. Issue a Run from UI

but not include:

1. Update any settings of the workspace
2. Read the state of the workspace
3. Pull (dump out) the workspace state to local

The thinking behind this is that the state file could include sensitive data that we don't want all users to read and that users of the workspace should not have the ability to update settings such as IaC version, template etc.

Potential work included in the above suggestion:

1. A new role (Use workspace maybe)
2. Enhancement to UI to disable state file output and settings of the workspace.

stanleyz:
This change adds a separate role for managing Terraform state and also disables the corresponding UI items for users who can't mange workspace or state.

Additional changes:

1. Users who doesn't have the permission to manage state can run Terraform plan and apply but can't manipulate state with command such as `terraform state rm|pull|push`

fix #1245

Screenshot 2024-09-22 at 7 18 05 PM
Screenshot 2024-09-22 at 7 28 25 PM

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/security/state/StateService.java
+++ b/api/src/main/java/org/terrakube/api/plugin/security/state/StateService.java
@@ -1,0 +1,33 @@
+package org.terrakube.api.plugin.security.state;
+
+import java.util.List;
+import java.util.UUID;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken;
+import org.springframework.stereotype.Service;
+import org.terrakube.api.repository.TeamRepository;
+import org.terrakube.api.rs.team.Team;
+
+@Service
+public class StateService {
+   @Autowired
+   private TeamRepository teamRepository;
+
+   public boolean hasManageStatePermission(Authentication authentication, String orgnizationId) {
+      Object groupNames = ((JwtAuthenticationToken) authentication).getTokenAttributes().get("groups");
+      if (groupNames == null) {
+         return false;
+      }
+      @SuppressWarnings("unchecked")
+      List<Team> teams = teamRepository.findAllByOrganizationIdAndNameIn(UUID.fromString(orgnizationId), (List<String>) groupNames);
+      for (Team team : teams) {
+         if (team.isManageState()) {
+            return true;
+         }
+      }
+      
+      return false;
+   } 
+}
--- a/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeController.java
+++ b/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeController.java
@@ -67,7 +67,7 @@ public class RemoteTfeController {
         HttpHeaders responseHeaders = new HttpHeaders();
         responseHeaders.set("TFP-API-Version", "2.5");
         responseHeaders.set("TFP-AppName", "Terrakube");
-        ResponseEntity response = new ResponseEntity<>(responseHeaders, HttpStatus.NOT_FOUND);
+        ResponseEntity<String> response = new ResponseEntity<>(responseHeaders, HttpStatus.NOT_FOUND);
         return response;
     }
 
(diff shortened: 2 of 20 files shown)


alfespa17:
Hello @stanleyz thank you for your help, I add some comments about the code so you can check

stanleyz:
> Hello @stanleyz thank you for your help, I add some comments about the code so you can check

hmm, I don't see any comments in the code, did you click on finish review?

Screenshot 2024-09-23 at 4 41 33 PM

stanleyz:
Hi @alfespa17 can you please check your page? I don't see any comment.

alfespa17:
hello @stanleyz these are my comments:

- I think it could be better to leave it with "isManageWorkspace" instead of harding the values here

https://github.com/stanleyz/terrakube/blob/9fa415c3a6071c8daad8b76085e20fbe80bb59ae/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java#L315

https://github.com/stanleyz/terrakube/blob/9fa415c3a6071c8daad8b76085e20fbe80bb59ae/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java#L317

https://github.com/stanleyz/terrakube/blob/9fa415c3a6071c8daad8b76085e20fbe80bb59ae/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java#L318

https://github.com/stanleyz/terrakube/blob/9fa415c3a6071c8daad8b76085e20fbe80bb59ae/api/src/main/java/org/terrakube/api/plugin/state/RemoteTfeService.java#L320

image

- I think you could remove this part 

https://github.com/stanleyz/terrakube/blob/9fa415c3a6071c8daad8b76085e20fbe80bb59ae/api/src/main/java/org/terrakube/api/plugin/token/team/TeamTokenController.java#L67
```
permissions.manageState ||
permissions.manageWorkspace ||
permissions.manageModule ||
permissions.manageProvider ||
permissions.manageTemplate  || 
permissions.manageVcs ||
``` 

image


those values will always be `false` because you define `PermissionSet permissions` above, the default value for a boolean in false and you are not changing those values before

stanleyz:
> I think it could be better to leave it with "isManageWorkspace" instead of harding the values here

With it set to `isManageWorkspace`,  to use workspace (run Terraform plan/apply) the person has to have `Manage Workspace` permission. However `Manage workspace` permission gives more power than needed to use workspace, for example, users with such permission could change the template for the workspace to bypass necessary checks, making it not ideal for situations where for example contractors are only allowed to code but not manage any checks/processes.

By settings these a few values to `true`, any users who are in a group added to an organisation but doesn't have `manage` permissions can still use the workspaces (run terraform plan/apply) of that organization following checks/processes set in these workspaces.

> I think you could remove this part

I don't agree. A user could in multiple groups which have different permissions, for example:

* Group A has manageWorkspace but not manageTemplate
* Group B has manageTemplate but not manageWorkspace

The users in both Group A and B should have both `manageWorkspace` and `manageTemplate` permission.

alfespa17:
> > I think it could be better to leave it with "isManageWorkspace" instead of harding the values here
> 
> With it set to `isManageWorkspace`, to use workspace (run Terraform plan/apply) the person has to have `Manage Workspace` permission. However `Manage workspace` permission gives more power than needed to use workspace, for example, users with such permission could change the template for the workspace to bypass necessary checks, making it not ideal for situations where for example contractors are only allowed to code but not manage any checks/processes.
> 
> By settings these values to `true`, any users who are in a group added to an organisation but doesn't have `manage` permissions can still use the workspaces (run terraform plan/apply) of that organization following checks/processes set in these workspaces.
> 
> > I think you could remove this part
> 
> I don't agree. A user could in multiple groups which have different permissions, for example:
> 
>     * Group A has manageWorkspace but not manageTemplate
> 
>     * Group B has manageTemplate but not manageWorkspace
> 
> 
> The users in both Group A and B should have both `manageWorkspace` and `manageTemplate` permission.

Thank you @stanleyz , now I understand

PR #1314, opened 2024-09-23
feat: Add verbose mode

stanleyz:
Only print the content of the script when verbose is on.

Additional changes:

1. import and format added by the editor when saving


alfespa17:
Hello @stanleyz you will have to add the value here too

https://github.com/AzBuilder/terrakube/blob/main/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/model/Command.java

PR #1363, opened 2024-10-05
Filter text appropriate to the filter applied to the workspace 

### LINKED ISSUE #1359: Workspace view remains filtered after changing screens but no filter text is shown
### Bug description 🐞

If you filter the workspaces in the UI and change the screen to another page any page the filtering stays active. However the filter text is disappeared but active. Pressing enter again will reset the filter and shows all workspaces. It is mostly confusing that the filter is active but now showing. 

image
In this example you can see I have no filter text but only 2 workspaces showing. However this organization has a lot more workspaces present.

### Steps to reproduce

* Filter the workspace view
* Go to another page in Terrakube 
* Go back to the workspace view
* Filter remains active

### Expected behavior

Filter to be either gone after changing pages in Terrakube or the filter text to be shown. I would prefer myself that the filter is reset. 

### Example repository

_No response_

### Anything else?

_No response_

Avinashs7:
Fixes #1359 Workspace view remains filtered after changing screens but no filter text is shown

## Description
- Now if workspace is filtered and moved to next page and returning will keep filter active with the proper filter applied, can be text in the search bar or filter based on tags.
- Earlier these tags or search input wasn't persistent in the view. 

Commit: 004dff9642032f9e8282c0e4a1883ca005ee6dab
Author: Avinashs7
Date: 05/10/2024

### DIFF
--- a/ui/src/domain/Organizations/Details.jsx
+++ b/ui/src/domain/Organizations/Details.jsx
@@ -161,7 +161,6 @@ export const OrganizationDetails = ({
     console.log(searchValue || "serach empty");
     console.log(filterValue || "filter empty");
     console.log(selectedTags || "tags empty");
-
     var filteredWorkspaces = filterWorkspaces(
       workspaces,
       searchValue,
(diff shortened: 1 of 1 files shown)


alfespa17:
This is not working @Avinashs7, I was testing it in [GITPOD](https://github.com/AzBuilder/terrakube/blob/main/development.md)

Now I am not able to type anything in the search bar.

image

alfespa17:
> Sorry! I overlooked the search input filter and it wasn't working because forgot to mention the helper function for onChange event. I have raised new PR solving this issue since I didn't know how to re raise the same request modifying. Please can you check the updated one and let me know if it has any issue.

You only need to send a push to the same branch that you are using in your fork and I will see the additional changes in the PR here

PR #1365, opened 2024-10-07
feat: Add Maven cache step to GitHub Actions workflow for disk optimization

ilkerispir:


ilkerispir:
Hello @alfespa17, @jcanizalez,

Related Issue: https://github.com/AzBuilder/terrakube/issues/1364

Add Maven Cache Step to GitHub Actions Workflow

This PR introduces a caching mechanism for Maven dependencies in the GitHub Actions workflow. By caching the ~/.m2 directory, we can significantly reduce build times and disk usage, especially for repeated builds where the dependencies haven't changed.

Key Changes:
Added a step to cache Maven dependencies using actions/cache@v3.
The cache key is based on the operating system and the pom.xml file hash, ensuring that cache is invalidated only when dependencies change.
Configured restore-keys to allow fallback to older caches if an exact match is not found.
Benefits:
Faster Builds: Dependencies won't be downloaded repeatedly on each build, which will speed up the build process.
Optimized Disk Usage: By reusing the cached dependencies, the amount of disk space required during the build process will be reduced.
Reduced Network Overhead: Fewer external network calls to Maven repositories, as dependencies are retrieved from cache when possible.
This change should help streamline the build process and improve overall performance, especially when running multiple builds.

alfespa17:
Thank you @ilkerispir

ilkerispir:
It's a pleasure @alfespa17, could you please trigger the release pipeline one more time when you're available? It won't break anything but I'm not sure if it will fix it, I mean there might still be no space on the disk for the GitHub action agent, I hope it works.

alfespa17:
> It's a pleasure @alfespa17, could you please trigger the release pipeline one more time when you're available? It won't break anything but I'm not sure if it will fix it, I mean there might still be no space on the disk for the GitHub action agent, I hope it works.

@ilkerispir we will have to wait for @jcanizalez , to build the ARM image we are using a github self hosted runner and he is the one with access to that machine and it appear to be offline

image

jcanizalez:
Thanks @ilkerispir 
The issue it's fixed and the multi arch images are available now for all the versions that failed

PR #1389, opened 2024-10-07
filter tag and search value remain until the filter is removed.

### LINKED ISSUE #1359: Workspace view remains filtered after changing screens but no filter text is shown

Avinashs7:
Fixes #1359 Workspace view remains filtered after changing screens but no filter text is shown

## Description
- Now if workspace is filtered and moved to next page and returning will keep filter active with the proper filter applied, can be text in the search bar or filter based on tags.
- Earlier these tags or search input wasn't persistent in the view. 

Commit: 6fa23e15cdb5a3e9f7d4e073bda2effc0a7a3cc3
Author: Avinashs7
Date: 07/10/2024

### DIFF
--- a/ui/src/domain/Organizations/Details.jsx
+++ b/ui/src/domain/Organizations/Details.jsx
@@ -137,10 +137,11 @@ export const OrganizationDetails = ({
   };
 
   const onSearch = (value) => {
-    setSearchValue(value);
     applyFilters(value, filterValue, filterTags);
   };
-
+  const updateSearchValue=(e)=>{
+    setSearchValue(e.target.value);
+  }
   const getTagName = (tagId) => {
     return tags.data.find((tag) => tag.id === tagId)?.attributes?.name;
   };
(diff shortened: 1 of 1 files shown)


alfespa17:
Now it is working correctly, thank you for your help @Avinashs7 

image

PR #1418, opened 2024-10-16
fix: problems encountered while building the effective model

dawidmalina:
fix: https://github.com/AzBuilder/terrakube/issues/1417

### DIFF
--- a/registry/pom.xml
+++ b/registry/pom.xml
@@ -74,10 +74,6 @@
 			<artifactId>spring-boot-starter-actuator</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-test</artifactId>
-		</dependency>
-		<dependency>
 			<groupId>org.projectlombok</groupId>
 			<artifactId>lombok</artifactId>
 			<version>${lombok.version}</version>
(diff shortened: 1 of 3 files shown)


registry/pom.xml
```diff
@@ -93,8 +89,8 @@
 			<version>${terrakube-client-starter.version}</version>
 			<exclusions>
 				<exclusion>
-					<groupId>com.squareup.okio:okio-jvm</groupId>
-					<artifactId>3.0.0 </artifactId>
+					<groupId>com.squareup.okio</groupId>
```
dawidmalina, line 92: Looks like it was not working initially and when I fixed that tests start to fail. Removing that exclusion will fix the build. @alfespa17 what will be your suggestion here?

alfespa17: I am not sure, this is why I never fixed the warning messages

dawidmalina: I just double checked and it was not working. You can test it by doing that:
- check this branch (including proper exclusion):
```
cd registry
mvn help:effective-pom
```
- check main branch:
```
cd registry
mvn help:effective-pom
```

You will see that in the main branch effective pom does not contains that exclusion. In other words we should remove it completely.

dawidmalina: Re-based with latest changes related to sonar build.

PR #1420, opened 2024-10-16
feat: AWS assume role support

### LINKED ISSUE #848: Ability to use S3 without access/secret key
### Feature description 💡

If running the containers (API/Registry/Executor) on native AWS hardware - there is no need to provide Access Key and Secret Key.

The containers can assume the native IAM role of the EC2/ECS/EKS. Therefore, the S3Client does not have to be initiated with access/secret key.

A toggle for this would be nice.

### Anything else?

_No response_

alfespa17:
I have no idea if this PR will work, I don't have a way to test it. I will leave it as draft for now if anyone would like to test it.

This will require to add the following env variables.

AwsEnableRoleAuth=true
AwsRoleArn=(some aws value)
AwsRoleSessionName=(some aws value)

Fix #848


alfespa17:
@ehuizar1028 if you can create a custom image based on this branch to test, maybe this work or not I have no way to test it, I just updated the old branch "awsrole" with all the recent changes and I added the changes inside all the components (this was a big copy paste XD)

alfespa17:
Hello @ehuizar1028  @BenjaminDecreusefond I created two images that you can test.

## 2.24.0-role.1 

This version can be used with the following environment variables.

AwsEnableRoleAuth=true
AwsRoleArn=(some aws value)
AwsRoleSessionName=(some aws value)

## 2.24.0-role.2

This version can be used with the following environment variables.

AwsEnableRoleAuth=true

> Maybe the second one works because it is using this [suggestion](https://github.com/AzBuilder/terrakube/issues/1422#issuecomment-2419588404) 

```
                        log.info("Using Role Authentication");
                        s3client = AmazonS3ClientBuilder.standard()
                                .withCredentials(new DefaultAWSCredentialsProviderChain())
                                .withRegion(awsStorageTypeProperties.getRegion())
                                .build();
```

for both If you deployed it  with the helm chart you will need to use `"12345" or some other value` for the accessKey and secretKey

alfespa17:
Closing PR I migrate the code to sdk v2

Issue #1431, opened 2024-10-18
Issues with private registry github app


BenjaminDecreusefond:
### Feedback

Hi!

We are trying to set up the private registry with a github app and we are facing issue regarding the app.
We followed this [https://docs.terrakube.io/user-guide/vcs-providers/github-app](doc) to create the VCS provider with the following parameter
- We named the app Terrakube (OrganizationName)
- Set the home page url to our api url endpoint
- set the correct privilege for the app
- generated the private key and put it into terrakube with the correct pkcs8 format
- and allowed it on all our repos

we then followed this [https://docs.terrakube.io/user-guide/private-registry/publishing-private-modules](documentation) to set up the module. We set the following parameter
- github url to the repo that stores the modules
- set a prefix tag for the repo
- base path for the module is /modules/mymodule/
When we validate the config it creates the module with the version already present, however when we create a new tag version it is never taken into account by the module and never appear in the UI and we keep having this log in the API logs
```
2024-10-18T18:18:00.242Z  INFO 1 --- [ryBean_Worker-2] o.terrakube.api.rs.module.GitTagsCache   : vcs using GITHUB
2024-10-18T18:18:00.242Z ERROR 1 --- [ryBean_Worker-2] o.t.a.plugin.scheduler.module.CacheJob   : Updating module index for Weezevent/database/aws
2024-10-18T18:18:00.242Z ERROR 1 --- [ryBean_Worker-2] o.t.a.plugin.scheduler.module.CacheJob   : Cannot invoke "org.terrakube.api.rs.vcs.GitHubAppToken.getToken()" because "gitHubAppToken" is null
``` 

It is very strange and we have trouble identifying where it could come from ?
Do you have any clue for us please ? 

Regards ! :)

alfespa17:
Did you install the app in your github organization?

BenjaminDecreusefond:
Yep we did ! We give it privileges to all repositories

alfespa17:
Did you use command for the private key?

`openssl pkcs8 -topk8 -inform PEM -outform PEM -nocrypt -in my-terrakube-app.private-key.pem -out pkcs8.key`

BenjaminDecreusefond:
yep we did ! and put the output key from pkcs8.key into the terrakube VCS provider

alfespa17:
Can you create a workspace using the github app and run a plan just to check if the connection is working correctly?

BenjaminDecreusefond:
[...]

PR #1432, opened 2024-10-19
fix: GitHub App Token not saved for modules

stanleyz:
When the module is saved, the ID of GitHubAppToken is not saved, this change fixes it by using the right JPA annotation.

Additional changes;

1. Organize imports


stanleyz:
fixes #1431

PR #1458, opened 2024-10-26
feat: Improve CLI driven workflow execution time

stanleyz:
People are quite sensitive to the amount of time they have to wait for Terraform runs kicked off from CLI, this change triggers such runs straight after it's created.

Additional changes:

1. Format
2. Remove excessive `startNow` call from `createJobContextNow`

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJobService.java
+++ b/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJobService.java
@@ -1,19 +1,32 @@
 package org.terrakube.api.plugin.scheduler;
 
-import lombok.AllArgsConstructor;
-import lombok.extern.slf4j.Slf4j;
+import java.text.ParseException;
+import java.time.Instant;
+import java.util.Date;
+import java.util.UUID;
+
+import org.quartz.CronExpression;
+import org.quartz.CronScheduleBuilder;
+import org.quartz.JobBuilder;
+import org.quartz.JobDataMap;
+import org.quartz.JobDetail;
+import org.quartz.JobKey;
+import org.quartz.Scheduler;
+import org.quartz.SchedulerException;
+import org.quartz.SimpleScheduleBuilder;
+import org.quartz.Trigger;
+import org.quartz.TriggerBuilder;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
 import org.terrakube.api.repository.StepRepository;
 import org.terrakube.api.repository.WorkspaceRepository;
 import org.terrakube.api.rs.job.Job;
 import org.terrakube.api.rs.job.JobStatus;
 import org.terrakube.api.rs.job.step.Step;
-import org.quartz.*;
-import org.springframework.stereotype.Service;
-import org.springframework.transaction.annotation.Transactional;
 import org.terrakube.api.rs.workspace.Workspace;
 
-import java.text.ParseException;
-import java.util.UUID;
+import lombok.AllArgsConstructor;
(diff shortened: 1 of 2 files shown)


stanleyz:
On this topic @ alfespa17, any reasons we should not do the same for all the runs?

alfespa17:
Hello @stanleyz there is a reason behind it.

When you create a job it validate if there are jobs running to continue the execution, if there are jobs running it will check in 1 minute

For example here, it validates if there are no previous jobs and does nothing until it checks again in 1 minute
https://github.com/AzBuilder/terrakube/blob/e30c9863d034fb8b7908ce566dab2a08a276215d/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java#L109

If you trigger the job intermediately the quartz job will only run once and terraform cli will be waiting forever to continue because the job wont change from the initial pending status never here

https://github.com/AzBuilder/terrakube/blob/e30c9863d034fb8b7908ce566dab2a08a276215d/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java#L112

And if you are running the CLI driven workflow the CLI is expecting previous jobs execute in order and shows something like this:
```
https://terrakube-api.minikube.net/app/simple/locked-remote/runs/run-876

Waiting for 29 queued run(s) to finish before starting...
Waiting for 28 queued run(s) to finish before starting... (30s elapsed)
Waiting for 27 queued run(s) to finish before starting... (1m0s elapsed)
```

This is why to start the job execution Terrakube always create the quartz job using using `scheduleJobService.createJobContext(job);` instead of `scheduleJobService.createJobContextNow(job);`

Once the job starts Terrakube trigger a job intermediately each time the Job entity is update here:

https://github.com/AzBuilder/terrakube/blob/e30c9863d034fb8b7908ce566dab2a08a276215d/api/src/main/java/org/terrakube/api/rs/hooks/job/JobManageHook.java#L37

let me know if you have any suggestion to improve that logic

stanleyz:
Right, thanks @alfespa17 for the explanation, I didn't take a close a look on the CRON expressions. 

Can we modify the `createJobContext` to kick off the job straight away with 30 seconds frequency (to be consistent with TF client prompting internal, or even make it customisable)?

alfespa17:
> Right, thanks @alfespa17 for the explanation, I didn't take a close a look on the CRON expressions.
> 
> Can we modify the `createJobContext` to kick off the job straight away with 30 seconds frequency (to be consistent with TF client prompting internal, or even make it customisable)?

Yeah sounds good, you can change that

alfespa17:
The value is here by the way

https://github.com/AzBuilder/terrakube/blob/e30c9863d034fb8b7908ce566dab2a08a276215d/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJobService.java#L25

stanleyz:
Can you please take a look @alfespa17 ? Thanks. I "feel" it's faster, if there are available executors, new job will be kicked off straight after it's created, otherwise the quartz will take about 20 seconds to pick it up.

alfespa17:
hello @stanleyz I added some comments in the method `scheduleJobService.createJobContextNow`

stanleyz:
> hello @stanleyz I added some comments in the method `scheduleJobService.createJobContextNow`

Hi mate, I do not see your comment, did you submit?

alfespa17:
> > hello @stanleyz I added some comments in the method `scheduleJobService.createJobContextNow`
> 
> Hi mate, I do not see your comment, did you submit?

Hello @stanleyz it was related to a `.startNow()` that was removed from method `createJobContextNow` but I just noticed it was duplicated, so forget about my comment I will merge the change

PR #1505, opened 2024-11-08
feat: add possibility to mount config map on ephemeral job agent

BenjaminDecreusefond:
Hey @alfespa17 !

We are currently working on ephemeral agents and we would like to add the ability to mount volumes on the jobs to allow for custom configuration like passing configuration files to interact with cloud providers. Do you think that could be a cool feature ?

Also, following this [PR](https://github.com/AzBuilder/terrakube/pull/1457) I update the message in the Terrakube UI to tell user that they can use regex for branch matching as well !

Regards !

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
+++ b/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
@@ -86,7 +86,7 @@ public class EphemeralExecutorService {
         Optional<String> configMapMountPathOpt = Optional.ofNullable(executorContext.getEnvironmentVariables().get("CONFIG_MAP_MOUNT_PATH"));
         if (configMapNameOpt.isPresent()) {
             String configMapName = configMapNameOpt.get();
-            String mountPath = configMapMountPathOpt.orElse("/etc/config");  // Default mount path if not specified
+            String mountPath = configMapMountPathOpt.orElse("/home/");  // Default mount path if not specified
 
             // Define ConfigMap volume
             Volume configMapVolume = new Volume();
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
```diff
@@ -80,6 +78,31 @@ public boolean sendToEphemeralExecutor(Job job, ExecutorContext executorContext)
...
 
+        // Volume and VolumeMount for ConfigMap if specified
+        List<Volume> volumes = new ArrayList<>();
+        List<VolumeMount> volumeMounts = new ArrayList<>();
+
+        Optional<String> configMapNameOpt = Optional.ofNullable(executorContext.getEnvironmentVariables().get("CONFIG_MAP_NAME"));
+        Optional<String> configMapMountPathOpt = Optional.ofNullable(executorContext.getEnvironmentVariables().get("CONFIG_MAP_MOUNT_PATH"));
+        if (configMapNameOpt.isPresent()) {
+            String configMapName = configMapNameOpt.get();
+            String mountPath = configMapMountPathOpt.orElse("/home/");  // Default mount path if not specified
```
juan-vg, line 89: I would move the default mount path to `/tmp` or `/mnt`. From my POV `/home` is not safe to be used as a mount point.

BenjaminDecreusefond: I set it as default because it is used by AWS for instance to fetch its .aws/config folder but you're right as a safe point i'll move to tmp !

alfespa17 (approved):

BenjaminDecreusefond:
Hi @alfespa17 !

Do you have an ETA for when this will be available in executor image ? :)

alfespa17:
Maybe I create the image later today I would like to see if I have some free time to add support for variable sets

BenjaminDecreusefond:
Ok sure thanks !

BenjaminDecreusefond:
Hey @alfespa17 ! 

Sorry to bother you, but do you have any updates about this ? :)

Issue #1548, opened 2024-11-19
Drift detection customisation


BenjaminDecreusefond:
### Feedback

Hi ! 

We would like to customise the default drift_detection provided on the Terrakube [documentation](https://docs.terrakube.io/user-guide/drift-detection). In its current state the template exec a plan and returns a slack message depending on the changes. However, if the plan fails for some reason the drift template returns nothing and we have no information about that. I see that the plan is triggered by the lines
```
flow:
  - type: "terraformPlan"
    step: 100
    name: "Running Terraform Plan with Drift Detection and Slack Notification"
```

But I wonder if this is a kind of shortcut for Terrakube itself ? Also, if the plan if failing during those lines the rest of the code will never execute and we have no options catch the error. Do you know if there's a way to bypass that behavior ?

Regards ! 
Benjamin

alfespa17:
Hello @BenjaminDecreusefond I dont think that will be possible because when Terrakube is running the terraform plan here it takes the exit code of the operation.

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/executor/src/main/java/org/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java#L118

Based on the exit code it updates the following flag

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/executor/src/main/java/org/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java#L125

That is later used to run the script after the terraform plan.

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/executor/src/main/java/org/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java#L130

In this validation

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/executor/src/main/java/org/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java#L290

BenjaminDecreusefond:
Hi @alfespa17 !

Ahh I see ! Is there a way to avoid this behavior ? Because not having error displayed from the drift check makes it kind of hard to manage many workspaces ? In this casing allowing the rest of the script to run is not an issue as we mainly want to return slack messages. Do you see a case where it would not be preferable to keep it running ?

alfespa17:
> Hi @alfespa17 !
> 
[...]

PR #1549, opened 2024-11-19
fix: always exec plan after error

BenjaminDecreusefond:
Hey @alfespa17 !

So I'm pushing as a first idea. The logic is that, for read-only operations like plan command we might want to run what's next to for instance display a slack message in case of error as mentionned in #1548. I don't think it is breaking any current logic and allows to achieve more complex patterns with templates ?

Idk what do you think ?
Lemme know if I missed something !


alfespa17:
I am not going to merge this you only hardcoded the value XD, it wont break any logic but at least try to change the code to validate if the execution needs to continue in some way not just hardode the value to true always XD

PR #1550, opened 2024-11-20
feat: Allow to continue template execution using "ignoreError" property

BenjaminDecreusefond:
Hi @alfespa17 !

I'd like to propose this small change and add a new field to the Terrakube template. I change the Flow template in the API and added it to the Executor job. 

However, I'd like your inputs about it, I think I'm missing the part where the data is transmitted from the API that parse the yaml to the Executor that use the content of the Yaml to run the Job. I tried to find it but I couldn't find the connection between the 2 components unfortunately. My guess is that the API set the job in the queue and the executor will pick'em for execute. I'm really not sure that the field `ingore_error` will be available in the Executor context. Could you explain me please ? :) 

Regards

### DIFF
--- a/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
+++ b/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
@@ -27,6 +27,7 @@ public class ExecutorContext {
     private boolean refresh;
     private boolean refreshOnly;
     private boolean showHeader;
+    private boolean ignore_error;
     private String accessToken;
     private String moduleSshKey;
     private String commitId;
--- a/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
+++ b/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
@@ -27,7 +27,7 @@ public class ExecutorContext {
     private boolean refresh;
     private boolean refreshOnly;
     private boolean showHeader;
-    private boolean ignore_error;
+    private boolean ignoreError;
     private String accessToken;
     private String moduleSshKey;
     private String commitId;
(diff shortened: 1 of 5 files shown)


alfespa17:
This looks better, let me check the logic that you have changed, and I will let you know if it requires any additional change

alfespa17:
This is the part of the code that is parsing the YAML file 

The job table has a field called "tcl" that has the base64 encoded yaml that will be executed when the job is running.

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/api/src/main/java/org/terrakube/api/rs/job/Job.java#L71

This is the service that parse the yaml into a java class:

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/TclService.java#L82

> By the way TCL means "terrakube configuration language"

When the job is running it is checking the pending steps here

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java#L159

Parsing the pending steps from here

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/api/src/main/java/org/terrakube/api/plugin/scheduler/job/tcl/TclService.java#L111

And it validates the pending job parts here

https://github.com/AzBuilder/terrakube/blob/a06994657607155604e83f419e54c75991372c59/api/src/main/java/org/terrakube/api/plugin/scheduler/ScheduleJob.java#L159

alfespa17:
> Okayy thank you @alfespa17 that's a Java trick ! I didn't know it !
> 
> I added the value to the ExecutorContext which should then allow the Executor code to retrieve the value, am I right ?

I think your change should work just one small thing can you rename `ignore_error` to `ignoreError` so we use the same variable name standard?

BenjaminDecreusefond:
Sure! Done ! :)

alfespa17:
I tested with this and it is working, it is ignoring the terraform plan error and it is running the script

```yaml
flow:
- type: "terraformPlan"
  name: "Plan"
  step: 100
  ignoreError: true
  commands:
  - runtime: "BASH"
    priority: 100
    after: true
    script: |
      echo "Running after error"
```

image

BenjaminDecreusefond:
Hi @alfespa17 ! 

Would it be possible to have a release please ? :) 

thanks ! :)

alfespa17:
You can check [2.24.0-beta.6](https://github.com/AzBuilder/terrakube/releases/tag/2.24.0-beta.6) @BenjaminDecreusefond

PR #1592, opened 2024-12-05
feat: support value in the tolerations config

klinux:
This PR support value in the toleration config.

In GCP we can't create tolerations using only key, and in some configurations, we need to support value in the toleration.

Now, terrakube supports only: **key:operator:effect**, the PR insert the possibilily to set: **key=value:operator:effect**.

Thank you


alfespa17:
I will check th pr tomorrow

alfespa17:
Do you need this one in 2.24.1? @klinux

Issue #1729, opened 2025-01-27
Include github-markdown-css locally to support offline environments


Jaguar-Kwok:
I am using Terrakube in an air-gapped/offline environment, and I noticed that the current implementation relies on an external CDN to load the github-markdown-css stylesheet ([via cdnjs.cloudflare.com](https://cdnjs.cloudflare.com/ajax/libs/github-markdown-css/4.0.0/github-markdown.min.css)). Since offline environments cannot access this CDN, the styling for markdown content is broken, affecting the user experience.

**Proposed Solution:**
To make Terrakube fully functional in offline setups, we recommend embedding the github-markdown-css CSS file directly in the project instead of using the CDN. This involves:

Adding the [MIT-licensed github-markdown-css](https://github.com/sindresorhus/github-markdown-css) file (v4.0.0) to the repository (e.g., in the public/ directory).

Updating the HTML template to reference the local CSS file instead of the CDN URL:
ui/public/index.html
` <link href="https://cdnjs.cloudflare.com/ajax/libs/github-markdown-css/4.0.0/github-markdown.min.css" rel="stylesheet">
`
Proposed change
`<link href="%PUBLIC_URL%/github-markdown.min.css" rel="stylesheet">`

**Why This Matters:**
This change ensures Terrakube works out-of-the-box in restricted networks without compromising functionality. Since the CSS is MIT-licensed, we can safely include it by preserving the license header in the file.

**Additional Notes:**
The specific version (4.0.0) can be downloaded directly from [the GitHub release](https://github.com/sindresorhus/github-markdown-css/releases/tag/v4.0.0).

Would the team be open to a pull request implementing this change? We’re happy to contribute if this aligns with the project’s goals!

PR #1751, opened 2025-01-28
feat: add NOTICE file for third-party components and update CSS link

Jaguar-Kwok:
#1729


alfespa17:
I kind of confused, this https://github.com/AzBuilder/terrakube/pull/1752 could also fix https://github.com/AzBuilder/terrakube/issues/1729 right?

Jaguar-Kwok:
Yes.

PR #1841, opened 2025-02-27
fix: UI directs to login page when the access token is expired

### LINKED ISSUE #1620: UI doesn't redirect to login when session expires
### Bug description 🐞

When the user session expires (UI+DEX GitHub), the UI stops loading resources, and the loading spinner is there forever.  

### Steps to reproduce

Wait until your login expires (a few days, I believe) and try to load a workspace (as an example). You'll notice no resources are loaded. Even after refreshing the page as many times as you want.
Then logout and login again, and everything works again.

### Expected behavior

Instead of displaying the loading spinner forever, I expect the UI to detect the expired session (API error, I believe) and redirect me to the login page. On top, renewing the session each time is used could be useful too.

### Example repository

_No response_

### Anything else?

_No response_

Avinashs7:
## Description

I have fixed the issue by using the expiry timestamp existing in the user object in localstorage. But I assume it needs to be fixed when it is been rejected in authorization process of each protected route.

Let me know if authorization middleware is done then I can work that as well.

## Results 

There is an expiry timestamp in the localstorage, change it to nearby time and wait until time crosses and try to request anything in the UI then it will route to login page.
I have tested manually by modifying the expiry time in the localstorage to mimic the real time expiration.
Kindly let me know if any error exists.

Fix #1620

### DIFF
--- a/ui/src/domain/Home/App.js
+++ b/ui/src/domain/Home/App.js
@@ -24,6 +24,12 @@ const { Header, Footer } = Layout;
 const App = () => {
   const auth = useAuth();
   const [organizationName, setOrganizationName] = useState("...");
+  const expiry = auth?.user?.expires_at;
+  
+  // Checking with the expiry time in the localstorage and when it has crossed the access has been revoked so It will clear the local storage and by default with no localstorage object it will route to login page.
+  if(auth.isAuthenticated && auth?.user && Math.floor(Date.now() / 1000) > expiry){
+    localStorage.clear();
+  }
 
   if (!auth.isAuthenticated) {
     return <Login />;


alfespa17:
Thank you @Avinashs7.

I think it is working correctly, I tested in GITPOD changing DEX configuration and set the token expiration in 2 minutes and the UI is redirected to the login once the token expired.

Example changing the default DEX yaml in `/scripts/setup/dex` to something like this:

```yaml
issuer: https://5556-avinashs7-terrakube-plfesnm3bvg.ws-us118.gitpod.io/dex

storage:
  type: memory
web:
  http: 0.0.0.0:5556
  allowedOrigins: ['*']
expiry:
  deviceRequests: "2m"
  signingKeys: "2m"
  idTokens: "2m"
  refreshTokens:
    reuseInterval: "5s"
    validIfNotUsedFor: "5m" 
    absoluteLifetime: "5m" 
```

Avinashs7:
@alfespa17 It works, but storing the whole user object as it is done now is not safe. So I believe only the tokens and trivial data should be stored. 
It is for security purposes and reduces vulnerabilities. Let me know your thoughts regarding this.

alfespa17:
> @alfespa17 It works, but storing the whole user object as it is done now is not safe. So I believe only the tokens and trivial data should be stored. It is for security purposes and reduces vulnerabilities. Let me know your thoughts regarding this.

If you have any suggestion feel free to create an issue describing the problem and a possible solution

Avinashs7:
@alfespa17 Sure, Once I will work on it in some time.

PR #1918, opened 2025-03-21
fix: ui workflow not triggering on pull requests

joachimdalen:
Seems like the workflow had an error defining `branches: [ opened, synchronize, reopened ]` instead of `types: [opened, synchronize, reopened]`

Also added lint and format check to steps for the new module directory.

1. Do you want this to trigger on merges as well? Linting and format checking does not really need to be run after a merge, but filtering these out seems to a bit tricky when the workflow runs on both PRs and merges. 

2. Would it be better to have one workflow for merges and one for PRs? 

3. Does API/backend validation/builds need to run if only the UI has changed?


alfespa17:
> Seems like the workflow had an error defining `branches: [ opened, synchronize, reopened ]` instead of `types: [opened, synchronize, reopened]`
> 
> Also added lint and format check to steps for the new module directory.
> 
> 1. Do you want this to trigger on merges as well? Linting and format checking does not really need to be run after a merge, but filtering these out seems to a bit tricky when the workflow runs on both PRs and merges.
> 2. Would it be better to have one workflow for merges and one for PRs?
> 3. Does API/backend validation/builds need to run if only the UI has changed?

1 - I was already working for merges
2 - It is a good idea to keep just one, to be honest I dont remember why we have several
3 - Backend build should not run for UI changes

Issue #1859, opened 2025-03-06
Make UI element right clickable to open in a new tab


BenjaminDecreusefond:
### Feature description 💡

Hi ! 

As of today in the 2.25.0 version of the terrakube UI we have no way to open workspace page in a new tab. 

For instance, let's say I want to see several workspaces configuration page. I must open manually myself tab to compare it, it is not possible to right click on the workspace and open it in new tab ! 

Lemme know if I was unclear ! 

Regards :)

alfespa17:
Fixed by #1919

PR #1919, opened 2025-03-22
feat: refactor workspaces list page

joachimdalen:
Refactor the home page of an organization and breaks it down into smaller components. 

I am a bit unsure about the colors of the workspace tags. They are deterministic based on id, so should be the same across workspaces. Any input here? Keep the new colors or just use the old blue one for all?

image

## Summary

- Refactor list of workspaces and update design
- Add page wrapper

## Other issues fixes
- Fixes workspace repo label for Azure DevOps
- Fixes workspace repo link for Azure DevOps

**Solves these issues**
- #1859


alfespa17:
I am not a UX designer but I think those new colors look good

alfespa17:
Thank you @joachimdalen

PR #1982, opened 2025-04-01
feat: mount pvc on jobs

BenjaminDecreusefond:
Hi @alfespa17 !

I'd like to propose this small change for ephemeral agents. The idea is that we use a Persistent volume claim (PVC) that we mount on our jobs at each run. Then since `TF_PLUGIN_CACHE_DIR` is set, tofu "automatically" comes to grab the content of this folder in case the provider was previously downloaded. 

NB: PVC must be defined in the Terrakube Helm charts. Terrakube app should not be responsible for PVC provisionning.

Lemme know your thoughts
PS: not tested


alfespa17:
Hello @BenjaminDecreusefond 

I think you will need to do some additional changes to this PR, maybe adding some code to validate if PVC_CLAIM_NAME is present if not the container will break because the is no pvc to add.

BenjaminDecreusefond:
Hi @alfespa17 !

Made rectification on the code ! Lemme know :)

PR #1985, opened 2025-04-02
feat: allow remote apply for vcs workspace

BenjaminDecreusefond:
Hi @alfespa17 !

I made this change as workaround for remote apply. This allows us to set a boolean to allow remote apply with VCS workspace. This can be useful to be able to use all features of VCS (like github connection) and avoid being blocked by VCS remote apply. I'd like your input as it changes the philosophy of VCS ! 

PS: UI part was not added due to lack of knowledge on UI components + I believe this parameter should only be used if really necessary.

Regards !


alfespa17:
You are adding a new column to the table you need to create the script to add the new column to the database, you could check several examples here

https://github.com/AzBuilder/terrakube/tree/main/api/src/main/resources/db/changelog/local

Can you rename this private `boolean allow_remote_apply = false;` to `private boolean allowRemoteApply = false;` to follow the same naming convention

BenjaminDecreusefond:
Hi @alfespa17 ! 

I made some changes, lemme know what you think ! :)

alfespa17:
Look like test cases are failing you can run `mvn test` in your local machine to do some debugging

BenjaminDecreusefond:
Hi @alfespa17 ! 

Was able to find a fix for the CI ! Should be good now ! :)

PR #2013, opened 2025-04-10
feat: change variables size

Vx-V:
Would like variables to hold larger values, private keys etc.
Current "input" UI element doesn't seem to work with multiline inputs


alfespa17:
thank you @Vx-V

alfespa17:
This PR is breaking when using with postgresql for example here: 

https://github.com/AzBuilder/terrakube/issues/2047

alfespa17:
I updated the code to only run for mssql

https://github.com/AzBuilder/terrakube/pull/2049

PR #2056, opened 2025-04-18
fix: Add listen for IPv6 for terrakube UI

guuse:
Currently, the UI only listens to IPv4. I'm running an IPv6 only cluster and was unable to run the UI unless I override the conf with a volume mount with proposed changes.


alfespa17:
Thank you @guuse

PR #2057, opened 2025-04-18
fix: Fix column name in workspace table

alfespa17:
Related to the issue describe here

https://github.com/AzBuilder/terrakube/pull/2052#issue-2999392560

### DIFF
--- a/api/src/main/java/org/terrakube/api/rs/workspace/Workspace.java
+++ b/api/src/main/java/org/terrakube/api/rs/workspace/Workspace.java
@@ -84,7 +84,7 @@ public class Workspace extends GenericAuditFields {
     @Column(name = "deleted")
     private boolean deleted;
 
-    @Column(name = "allowRemoteApply")
+    @Column(name = "allow_remote_apply")
     private boolean allowRemoteApply = false;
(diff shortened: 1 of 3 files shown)


BenjaminDecreusefond:
Hi @alfespa17 !

Perhaps the terraform provider also needs to be updated in consequence then since we are changing the column name ? 
Btw sorry but I don't understand what was the issue with the `AllowRemoteApply` parameter ? Was the syntax incorrect ?

Regards

alfespa17:
> Hi @alfespa17 !
> 
> Perhaps the terraform provider also needs to be updated in consequence then since we are changing the column name ? Btw sorry but I don't understand what was the issue with the `AllowRemoteApply` parameter ? Was the syntax incorrect ?
> 
> Regards

I only changed the colum name inside the database, but the property in the rest API will be the same

image

The problem was related to postgresql when doing some queries it was changing `allowRemoteApply ` to lowercase `allowremoteapply ` and some queries were failing because it was not able to find the column, that is why I renamed the column

BenjaminDecreusefond:
Ok cool thank !

PR #2161, opened 2025-05-30
fix: change volume ownership

BenjaminDecreusefond:


alfespa17:
There is a special scenario for the arm image the user is heroku, not sure if hardcoding the value is a good idea and I am not sure if it is using the same id or that will make it break for the heroku user. 

https://github.com/AzBuilder/terrakube/blob/67172ba6db63788d13d50002d33c2d525a8fcb10/.github/workflows/release.yml#L77

Maybe you can leave it as a default value but allow to change it in some way

BenjaminDecreusefond:
Hm, what you show me seems to be related to the executor image. Is it the same one used to run remote agents jobs ?

alfespa17:
> hi @alfespa17 In which case is this user used ? Perhaps I can try to test to see which group ID it is ?

that user id is used when you are running the arm image

alfespa17:
> Hm, what you show me seems to be related to the executor image. Is it the same one used to run remote agents jobs ?

That is correct, the executor image is the one that will run the job

fishfacemcgee:
Below is the output of the same command on an ephemeral container running on Arm

```
heroku@job-6335-2lkgk:/workspace$ cut -d: -f1 /etc/passwd
root
daemon
bin
sys
sync
games
man
lp
mail
news
uucp
proxy
www-data
backup
list
irc
_apt
nobody
sshd
heroku
```

```
heroku@job-6335-2lkgk:/workspace$ id -u
1000
heroku@job-6336-jd297:/workspace$ id -g
1000
```

So it looks like 1000 should work consistently in this case.

alfespa17:
Thank you for the confirmation @fishfacemcgee , in that case I think we can merge the PR

alfespa17:
> @alfespa17 do you think it would be possible to have a release to be able to test it out ? :)

I will do it later today

BenjaminDecreusefond:
Thanks!

BenjaminDecreusefond:
Hi @alfespa17 !

Would it also be possible to make a release for the terraform provider as it supports the allow_remote_apply option for VCS workspace ?

Regards !

alfespa17:
> Hi @alfespa17 !
> 
> Would it also be possible to make a release for the terraform provider as it supports the allow_remote_apply option for VCS workspace ?
> 
> Regards !

The option is already there

https://github.com/AzBuilder/terraform-provider-terrakube/blob/5baa949c8a4f871d78a5e08f5b7c9e71295e6db4/internal/provider/workspace_vcs_resource.go#L135

BenjaminDecreusefond:
I can't see it [here](https://registry.terraform.io/providers/AzBuilder/terrakube/latest/docs/resources/workspace_vcs) ?
And i can't see my change in the [release](https://github.com/AzBuilder/terraform-provider-terrakube/releases) list

alfespa17:
> I can't see it [here](https://registry.terraform.io/providers/AzBuilder/terrakube/latest/docs/resources/workspace_vcs) ? And i can't see my change in the [release](https://github.com/AzBuilder/terraform-provider-terrakube/releases) list

You are correct I created a new release

alfespa17:
The new release is failing I have no idea why, maybe you can check if you have some time

BenjaminDecreusefond:
Hi @alfespa17 !

You're talking about the Terrakube release ? not the provider one? 
Do you have any traces or anything ?

alfespa17:
This release has your changes https://github.com/AzBuilder/terraform-provider-terrakube/releases/tag/v0.14.0

BenjaminDecreusefond:
Hi @alfespa17 !

Do you think that would be possible to make a release containing this PR please ? :) 

Regards !

PR #2324, opened 2025-07-19
feat: Store module versions in database

stanleyz:
This change adds a table to store versions of modules and a Quartz job to fetch module versions periodically.
    
The benefit of such change is that:
    
1. When there is a network failure between Terrakube and upstream VCS, the  can still function as normal
2. The change also replaces the synchronous process to fetch versions in a module with an asynchronous process that conforms with the same  used by other components when fetching necessary information
3. Simplifies the module entity

Additional changes:

1. Replace deprecated functions / components as required
2. Fix a bug in compareVersion function within `Workspace.tsx` where parseInt takes one of version numbers for the radix parameter.


alfespa17:
Hello @stanleyz you can use this version, it should be ready in some minutes so you can use it for testing 

https://github.com/terrakube-io/terrakube-spring-boot-starter/releases/tag/1.1.0-beta.1

stanleyz:
> Hello @stanleyz you can use this version, it should be ready in some minutes so you can use it for testing
> 
> https://github.com/terrakube-io/terrakube-spring-boot-starter/releases/tag/1.1.0-beta.1

Thanks @alfespa17 , please take a look. ta.

alfespa17:
Hello @stanleyz I was testing this and I noticed the first time someone check the module there is an error like this.

image

In the background there is a job that will get the versions from the repository.

what do you think about running some small function when the API is starting to get those version instead of waiting for an update

Something like this using the `ApplicationListener<ContextRefreshedEvent>` to get all the moduleversion in parallel when this relationship is empty

```java
    @OneToMany(cascade = { CascadeType.REMOVE }, mappedBy = "module")
    private List<ModuleVersion> version;
```

Running when the API is starting:

https://github.com/terrakube-io/terrakube/blob/72c1265e9966c2e46c77a3020725371b27340d21/executor/src/main/java/io/terrakube/executor/service/mode/batch/BatchModeServiceImpl.java#L19

In theory that should only run once because all relationships will be empty

stanleyz:
Thanks @alfespa17 for the review. 

Yes, that's basically the below code, people will get an error the first time when they hit their existing modules. 

https://github.com/terrakube-io/terrakube/blob/3a2b397bbb949804537bc4f74bb5df95d80b736c/registry/src/main/java/io/terrakube/registry/service/module/ModuleServiceImpl.java#L42

Are you suggesting to have the function and also the above logic or just the function? And yes agree we can remove such logic in the next version once all modules are onboarded to the new logic.

alfespa17:
Hello @stanleyz I was thinking something like this

If you see the API logs when the API is starting you will see some internal process that is executed at startup time:

```
2025-07-23T16:57:16.763Z  INFO 3150 --- [           main] i.t.a.p.s.inactive.InactiveJobsService   : Setup job to cancelled inactive jobs
2025-07-23T16:57:16.767Z  INFO 3150 --- [           main] i.t.a.p.s.inactive.InactiveJobsService   : Create Schedule for inactive jobs: DEFAULT.TerrakubeV2_InactiveJobs_991b8346-dff6-4c21-b8a1-9da6f5fef75a
```

This is executed [here](https://github.com/terrakube-io/terrakube/blob/72c1265e9966c2e46c77a3020725371b27340d21/api/src/main/java/io/terrakube/api/plugin/scheduler/inactive/InactiveJobsService.java#L24)

```java
    @Transactional
    @PostConstruct
    public void initInactiveJobs() {
        try {
            log.info("Setup job to cancelled inactive jobs");
            runInactiveJobs();
            JobDetail jobDetail = scheduler.getJobDetail(new JobKey(PREFIX_INACTIVE_JOBS));
            log.info("jobDetail is null {}", jobDetail == null);
            if(jobDetail == null){
                setupInactiveJob("0 */5 * ? * *");
            } else {
                log.info("Delete Old Quartz Job for inactive jobs");
                scheduler.deleteJob(new JobKey(PREFIX_INACTIVE_JOBS));
                setupInactiveJob("0 */5 * ? * *");
            }
        } catch (Exception ex) {
            log.error(ex.getMessage());
        }

    }
```

You could use something similar to start getting all the module versions at startup time instead of waiting to failed the first time.

something like:

```java
    @Transactional
    @PostConstruct
    public void getInitialModuleVersion() {
        try {
            List<Organization> organizationList = organizationRepository.findAll();
            organizationList.forEach(organization -> {
                List<Module> moduleList = organization.getModule();
                moduleList.forEach(module -> {
                    if (module.getModuleVersion() == null)
                        trigger quartz job to get version now
                })
            })
        }
    } catch (Exception ex) {
        log.error(ex.getMessage());
    }

}
```

stanleyz:
Thanks, @alfespa17. I think that was a brilliant idea. I was just unsure whether we should have both this init function and the fail-then-fetch logic, but all is good; it's probably best to keep both. They can be deleted in future versions once everybody is on board with the new logic anyway.

Please take another look. Thanks.

alfespa17:
Thank you @stanleyz  it looks good now

PR #2389, opened 2025-08-06
feat: pre init step within runs

### LINKED ISSUE #2349: Run `before: true` before init step
### Feature description 💡

Hello,

I propose running `commands` defined using `before: true` prior to the `terraform init` step instead of after, basically (IIUC) moving the executePreOperationScripts calls up above the init step, e.g.
https://github.com/AzBuilder/terrakube/blob/704663ac6904e0e97de160b77e4407eaaf5c2fc2/executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java#L116

Alternatively introduce a new `beforeInit` option, if that makes more sense.

This would make the executors (especially the ephemeral ones) more flexible without having to build custom images, since we can e.g. use credentials generated from a script to pull Terraform modules. I don't see any downsides from the change, but might be missing something.

Happy to take a stab at it if you think it's a good idea.

-----

Some background and reasoning:

We're currently using the normal/stateful executor, but looking to move to the ephemeral ones.

We have a small blocker where we run a customScript to 1) grab a token from a Github App and 2) override the module sources to use this token instead of SSH:
```
  - type: "customScripts"
    name: "Set up Github auth"
    step: 100
    commands:
      - runtime: "BASH"
        priority: 100
        script: |
          echo "Setting up Github auth"

          # This script gets a short lived GH token
          source /opt/executor-helpers/git-helper.sh

          for remote in git@github.com: ssh://git@github.com/; do
            echo ":: Configuring https override for $remote"
            git config --global url."https://oauth2:${GITHUB_TOKEN}@github.com/".insteadOf "$remote"
          done

          echo "Done setting up Github auth"
  - type: "terraformPlan"
    name: "Plan"
    commands:
      - runtime: "BASH"
        priority: 100
        script: |
          echo "Cleaning up $HOME/.gitconfig"
          rm $HOME/.gitconfig
```

This relies on settings saved in `$HOME/.gitignore`, and as such won't persist across ephemeral executors.

I realize this might be a bit of an unorthodox setup, but we can't use SSH keys due to a myriad of reasons so this is what we ended up with.

### Anything else?

_No response_

CraigSamuelson:
fix #2349

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -111,8 +111,7 @@ x-ui: &ui_env
   JAVA_TOOL_OPTIONS: -Dcom.sun.security.enableAIAcaIssuers=true
 
 x-traefik_env: &traefik_env
-  TRAEFIK_API_DASHBOARD: "true"
-  TRAEFIK_API_INSECURE: "true"
+  TRAEFIK_API_DASHBOARD: "false"
   TRAEFIK_ENTRYPOINTS_WEB: "true"
   TRAEFIK_ENTRYPOINTS_WEB_ADDRESS: ":${TRAEFIK_HTTP_PORT}"
   TRAEFIK_ENTRYPOINTS_WEB_HTTP_ENCODEQUERYSEMICOLONS: "true"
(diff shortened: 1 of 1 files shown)


alfespa17:
I did a quick test in gitpod using the following template

```
flow:
  - type: "terraformPlan"
    step: 100
    commands:
      - runtime: "BASH"
        priority: 100
        beforeInit: true
        script: |
          echo "Hello before-init"
```

Template definition:

image

Checking the executor logs I can see the before init message

image

Message in the UI

image

alfespa17:
Finally if I remove it I can see there are no commands to run before init like this

image

CraigSamuelson:
@alfespa17 I think its ready, any idea if we can merge or run tests on this ?

alfespa17:
> @alfespa17 I think its ready, any idea if we can merge or run tests on this ?

I merged it

PR #2425, opened 2025-08-12
feat: allow optional redis password

BenjaminDecreusefond:
Hi @alfespa17 !

I would like to suggest this change for the redis configuration. The idea is as follow:
When using a redis cache on cloud provider like AWS, a password is not strictly necessary in order to connect to the redis. If we enable password, it requires turning on SSL and managing certificate which can quickly become complex set up.

I'd like to make password optional to facilitate redis set up when using cloud providers

Regards !


alfespa17:
Did you try the following https://github.com/terrakube-io/terrakube-helm-chart/issues/115#issuecomment-2402741845 to use redis with ssl ?

BenjaminDecreusefond:
Hi @alfespa17 !

Not sure to understand what should be stored in values `TerrakubeRedisTruststorePassword` and `TerrakubeRedisTruststorePath`, should i set the CA cert ?

Any case, I don't think we need SSL for our cache in our use case at my company as everything is running locally between our EKS nodes and our AWS managed cache, all in private networks. I'm thinking this PR only adds more flexibility for those who wants it, while keeping the same level of security if required.

What do you think ?

alfespa17:
> Hi @alfespa17 !
> 
> Not sure to understand what should be stored in values `TerrakubeRedisTruststorePassword` and `TerrakubeRedisTruststorePath`, should i set the CA cert ?
> 
> Any case, I don't think we need SSL for our cache in our use case at my company as everything is running locally between our EKS nodes and our AWS managed cache, all in private networks. I'm thinking this PR only adds more flexibility for those who wants it, while keeping the same level of security if required.
> 
> What do you think ?

By default the java truststore inside the container already includes 147 Ca certs, in my case the Azure CA cert was already there when I was testing, I guess the one need it by AWS should be there too.

Example:

image


`TerrakubeRedisTruststorePath=/layers/paketo-buildpacks_bellsoft-liberica/jre/lib/security/cacerts ` and `TerrakubeRedisTruststorePassword=changeit` are always the same because it is using the java keystore that is already included in the container to use SSL with redis

alfespa17 (approved):

BenjaminDecreusefond:
Ah great thank you ! I will try !

BenjaminDecreusefond:
Hi @alfespa17 !

I was able to make it work thank you ! We just had a small issue because parameters does not replicate inside the ephemeral job executor environment variable so it was trying to connect without SSL.
Could be nice to add some documentation about all this I think ahah !
thanks!

alfespa17:
> Hi @alfespa17 !
> 
> I was able to make it work thank you ! We just had a small issue because parameters does not replicate inside the ephemeral job executor environment variable so it was trying to connect without SSL. Could be nice to add some documentation about all this I think ahah ! thanks!

I always forget to update the docs XD

alfespa17:
I have updated the docs, now you can see the example to enable an external redis

https://docs.terrakube.io/getting-started/deployment/external-redis

alfespa17:
> Hi @alfespa17 !
> 
> I was able to make it work thank you ! We just had a small issue because parameters does not replicate inside the ephemeral job executor environment variable so it was trying to connect without SSL. Could be nice to add some documentation about all this I think ahah ! thanks!

Hey @BenjaminDecreusefond  quick question how did you fix the missing parameters for the ephemeral executor?

alfespa17:
I think this could be a workaround but not sure if this if the way you solved it @BenjaminDecreusefond 

There is a secret called `terrakube-executor-secrets` inside the terrakube namespace, the ephemeral executor is taking all the env variables from there when it is creating the kubernetes job
So if you duplicate the `terrakube-executor-secrets `and create a new one for example `terrakube-executor-secrets-ephemeral ` and  add the values:

- SERVICE_BINDING_ROOT=/mnt/platform/bindings
- TerrakubeRedisSSL=true
- TerrakubeRedisTruststorePath=/layers/paketo-buildpacks_bellsoft-liberica/jre/lib/security/cacerts 
- TerrakubeRedisTruststorePassword=changeit

Now you could override the default secret used for the ephemeral executor like this

```yaml
api:
  defaultRedis: false
  env:
  - name: TerrakubeRedisSSL
    value: "true"
  - name: TerrakubeRedisTruststorePath
    value: /layers/paketo-buildpacks_bellsoft-liberica/jre/lib/security/cacerts 
  - name: TerrakubeRedisTruststorePassword
    value: changeit
  - name: ExecutorEphemeralSecret
    value: terrakube-executor-secrets-ephemeral
  properties:
    redisHostname: "MY-REDIS-IN-AZURE.redis.cache.windows.net"
    redisPassword: "MY-REDIS-ACCESS-KEY"
    redisPort: "6380"
```

BenjaminDecreusefond:
> I think this could be a workaround but not sure if this if the way you solved it @BenjaminDecreusefond
> 
> There is a secret called `terrakube-executor-secrets` inside the terrakube namespace, the ephemeral executor is taking all the env variables from there when it is creating the kubernetes job So if you duplicate the `terrakube-executor-secrets `and create a new one for example `terrakube-executor-secrets-ephemeral ` and add the values:
> 
> * SERVICE_BINDING_ROOT=/mnt/platform/bindings
> * TerrakubeRedisSSL=true
> * TerrakubeRedisTruststorePath=/layers/paketo-buildpacks_bellsoft-liberica/jre/lib/security/cacerts
> * TerrakubeRedisTruststorePassword=changeit
> 
> Now you could override the default secret used for the ephemeral executor like this
> 
> ```yaml
> api:
>   defaultRedis: false
>   env:
>   - name: TerrakubeRedisSSL
>     value: "true"
>   - name: TerrakubeRedisTruststorePath
>     value: /layers/paketo-buildpacks_bellsoft-liberica/jre/lib/security/cacerts 
>   - name: TerrakubeRedisTruststorePassword
>     value: changeit
>   - name: ExecutorEphemeralSecret
>     value: terrakube-executor-secrets-ephemeral
>   properties:
>     redisHostname: "MY-REDIS-IN-AZURE.redis.cache.windows.net"
>     redisPassword: "MY-REDIS-ACCESS-KEY"
>     redisPort: "6380"
> ```

Sorry didn't see the mesage! 

Yes we fixed it like this ! Added key to the secret and executor was taking it into account !

PR #2428, opened 2025-08-14
feat: Add Consul backend support

crs07:


alfespa17:
Hello @crs07 

I few comments:

1 - You need to update the application.properties to include the new configuration options like [this](https://github.com/terrakube-io/terrakube/blob/089fca75747ac3410eeec68658b9fa89e4854d47/executor/src/main/resources/application.properties#L28) so you can add the properties from environment variables.

2 - I think you are not using this library so it can be removed. (correct my if I am wrong)

```
		<dependency>
			<groupId>com.bettercloud</groupId>
			<artifactId>vault-java-driver</artifactId>
			<version>5.1.0</version>
		</dependency>
```

3 - The dependency that you are using is not maintained and it has a lot of vulnerabilities you can check [here](https://mvnrepository.com/artifact/com.orbitz.consul/consul-client/1.5.3), doing some search I found [this comment](https://github.com/apache/pekko-management/issues/243#issuecomment-2314760312), looks like this is a fork that has active development [repo](https://github.com/kiwiproject/consul-client?tab=readme-ov-file)

alfespa17:
Another thing that you need to consider is that you will need to implement something similar for the API and Registry to manage the storage with consul

https://github.com/terrakube-io/terrakube/tree/main/api/src/main/java/io/terrakube/api/plugin/storage

https://github.com/terrakube-io/terrakube/tree/main/registry/src/main/java/io/terrakube/registry/plugin/storage

alfespa17:
I am closing this PR, it hasn't been updated in a long time and it has some conflict with the current code. If anyone would like to help with this changes send a new PR

Issue #2276, opened 2025-07-07
proxy http with authentification


chris93111:
### Bug description 🐞

Hello 

i try ton configure terrakube with a proxy http and need authentification

i send JAVA_TOOL_OPTIONS env with -Dhttp.proxyHost -Dhttp.proxyPort -Dhttp.proxyUser -Dhttp.proxyPassword , same with https 

but this not work , we can see error 407 authentification is not passed to proxy

api pod

```
2025-07-07T14:25:14.857Z  INFO 1 --- [nio-8080-exec-4] o.t.a.p.json.TerraformJsonController     : Using terraform releases URL https://releases.hashicorp.com/terraform/index.json
2025-07-07T14:25:14.982Z ERROR 1 --- [nio-8080-exec-4] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception

java.io.IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 authenticationrequired"
	at java.base/sun.net.www.protocol.http.HttpURLConnection.doTunneling0(Unknown Source) ~[na:na]
	at java.base/sun.net.www.protocol.http.HttpURLConnection.doTunneling(Unknown Source) ~[na:na]
	at java.base/sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(Unknown Source) ~[na:na]
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream0(Unknown Source) ~[na:na]
	at java.base/sun.net.www.protocol.http.HttpURLConnection.getInputStream(Unknown Source) ~[na:na]
	at java.base/sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(Unknown Source) ~[na:na]
	at java.base/java.net.URL.openStream(Unknown Source) ~[na:na]
	at org.apache.commons.io.IOUtils.toString(IOUtils.java:3204) ~[commons-io-2.19.0.jar:2.19.0]
	at org.apache.commons.io.IOUtils.toString(IOUtils.java:3179) ~[commons-io-2.19.0.jar:2.19.0]
	at org.apache.commons.io.IOUtils.toString(IOUtils.java:3293) ~[commons-io-2.19.0.jar:2.19.0]
	at org.apache.commons.io.IOUtils.toString(IOUtils.java:3252) ~[commons-io-2.19.0.jar:2.19.0]
[...]

PR #2451, opened 2025-08-25
Added support for proxy

capita-user:
This commit adds support for using a proxy with the WebClient in the Terrakube API. Key changes include:

•	Replaces usage of RestTemplate with WebClient in multiple services and controllers for HTTP requests.
•	Introduces a new configuration class (WebClientConfig) and properties (WebClientConfigProperties) to enable, configure, and use a proxy (including TLS, host, port, username, and password) for all WebClient instances.
•	Updates application properties files to include new proxy-related configuration options.
•	Refactors affected classes to use Spring-managed WebClient.Builder via dependency injection, ensuring consistent proxy settings.
•	Improves error handling and logging for HTTP requests made via WebClient.
•	Overall, this update centralizes and standardizes HTTP client configuration, enabling proxy support across the application.

### DIFF
--- a/api/src/main/resources/application-demo.properties
+++ b/api/src/main/resources/application-demo.properties
@@ -198,7 +198,7 @@ io.terrakube.dynamic.credentials.private-key-path=${DynamicCredentialPrivateKeyP
 #######################
 # Webclient Proxy     #
 #######################
-io.terrakube.api.plugin.webclient.proxy-enabled=${WebClientProxyEnabled:true}
+io.terrakube.api.plugin.webclient.proxy-enabled=${WebClientProxyEnabled:false}
 io.terrakube.api.plugin.webclient.proxy-use-tls=${WebClientProxyUseTls:false}
 io.terrakube.api.plugin.webclient.proxy-host=${WebClientProxyHost:}
 io.terrakube.api.plugin.webclient.proxy-port=${WebClientProxyPort:}


alfespa17:
Hello @P50040741 thank you for your help.

Some weeks ago I did some small changes when I was testing how to fix the proxy auth issue in #2276 and I found that there is a standard way to enable the proxy when using webclient just by using it like the following and the standard JAVA_TOOL_OPTIONS:

```java
        WebClient webClient = webClientBuilder
                .exchangeStrategies(ExchangeStrategies.builder()
                        .codecs(configurer -> configurer.defaultCodecs().maxInMemorySize(10 * 1024 * 1024))
                        .build())
                .baseUrl(terraformIndex)
                .clientConnector(
                        new ReactorClientHttpConnector(
                                HttpClient.create().proxyWithSystemProperties())
                ).build();
```

The important part is 

```java
                .clientConnector(
                        new ReactorClientHttpConnector(
                                HttpClient.create().proxyWithSystemProperties())
                ).
```

If you use the above you would not have to maintain these properties:

```properties
#######################
# Webclient Proxy     #
#######################
io.terrakube.api.plugin.webclient.proxy-enabled=${WebClientProxyEnabled:true}
io.terrakube.api.plugin.webclient.proxy-use-tls=${WebClientProxyUseTls:false}
io.terrakube.api.plugin.webclient.proxy-host=${WebClientProxyHost:}
io.terrakube.api.plugin.webclient.proxy-port=${WebClientProxyPort:}
io.terrakube.api.plugin.webclient.proxy-username=${WebClientProxyUsername:}
io.terrakube.api.plugin.webclient.proxy-password=${WebClientProxyPassword:}
```

Or these class

api/src/main/java/io/terrakube/api/plugin/webclient/WebClientConfigProperties.java

And you could simplify use `HttpClient.create().proxyWithSystemProperties())`

api/src/main/java/io/terrakube/api/plugin/webclient/WebClientConfig.java

For more detail information you could check [this comment](https://github.com/terrakube-io/terrakube/issues/2276#issuecomment-3115284274)

What do you think about that option?

capita-user:
Hi, @alfespa17, I'm picking up where the previous engineer left off - apologies for the lack of response.

I understand that you have identified an easier way to create the webClient without the need for defining the additional properties. This totally makes sense - it would be great if we can omit creating them. I will try this out.

I've got 2 follow up questions:

1. If the above is true and we can simply call:
```java
    WebClient webClient = webClientBuilder
            .exchangeStrategies(ExchangeStrategies.builder()
                    .codecs(configurer -> configurer.defaultCodecs().maxInMemorySize(10 * 1024 * 1024))
                    .build())
            .baseUrl(terraformIndex)
            .clientConnector(
                    new ReactorClientHttpConnector(
                            HttpClient.create().proxyWithSystemProperties())
            ).build();
```
whenever we need to create a webClient, it would elimiate the need to configure WebClientBuilder centrally and AutoWire it in other classes. Is this what you had in mind?

2. In terms of the other files changed in this PR, are you in agreement with the proposal to replace the usage of RestTemplate with WebClient and generally to use WebClient when routing via a proxy is needed? I know TerraformJsonController has already changed - other files would change in a similar way. 

I will review each of the changes and test that things are working as expected, but I want to check if the proposed changes here make sense to you.

alfespa17:
Correct I think the best way will be to use webclient and the proxy settings from the above example, it will simplify the changes proposed in this pr

PR #2565, opened 2025-10-02
fix: Registry performance issues

alfespa17:


BenjaminDecreusefond:
Hi @alfespa17 ! 

We tried this version out (2.27.3) on our Terrakube instance and it turns out that it was not able to detect new module versions anymore. After switching back to 2.27.2 our Terrakube was able to see new module versions again ! 

Regards !

PR #2580, opened 2025-10-06
Fix postgres AWS IAM token expiry issue

capita-user:
We have found that the token issued by the AWS RDS service for connections to an AWS Postgres intance often times expires earlier than the documented 10 minutes, which causes requests to the database to fail.

Incressing the refresh frequency to 5 minutes has resolved 99% of the occurances of these errors, however they still occur every now and then.

The try - catch block ensures that when such a failure does occur due to a timing issue, the token is refreshed before another attempt to establish a connection to the database is made.

General formatatting changes included.


alfespa17:
Thank you @capita-user

PR #2582, opened 2025-10-08
Wait for terraform state pull to complete

capita-user:
We have identified a strange issue where the output of the terraform state pull command does not get fully produced before it is set on the terraformJob, therefore truncating the rawStateJSON.

This happens especially often when the state of the workspace is very large.

One issue from a while ago discusses a similar problem: https://github.com/hashicorp/terraform/issues/293, but it is unclear why it is occuring here.

Adding a 5 second wait has resolved the issue and the state doesn't get truncated.


capita-user:
I can see the state pull command has been added to the terrakube fork of the terraform client and I have reviewed the code.

Nothing stands out as a potential issue - the state pull command executes like any other terraform command, returns a Future, which then gets resolved with get().

capita-user:
@alfespa17 I was not able to identify the root cause of this. It seems the terraform command itself it not returning the whole output in time. I don't see any issue with the Java code.

This solution is a workaround, but should resolve the issue.

Would it be okay to merge this?

alfespa17:
I don't have a way tor produce the issue with a big state file but I don't see any issue merging it

PR #2645, opened 2025-10-28
feat: Allow executor to send logs via the API

capita-user:
This pull request adds an optional functionality (enabled via an application property) to allow the executors to store the terraform logs in Redis by sending them to the API first. The API receives the logs from the executors and writes them to Redis.

Currently, if many executors exist, they write the terraform logs directly to Redis over the internet. With this optional feature, the API will serve as a form of gateway and all terraform logs will flow through it to Redis.

This increases the network security as Redis does not need to be directly exposed via inbound networking rules. This is particularly beneficial in solutions which deploy many executors to different environments.

This is accompanied by a pull request in the terrakube client repo (https://github.com/terrakube-io/terrakube-spring-boot-starter/pull/49), which allows the executor to send requests containing logs to the API.

In summary:
- The API exposes two new endpoints to set up redis consumer groups and receive terraform logs
- The executor can send terraform logs to the API instead of to Redis directly
- The terrakube client library is modified to include the new requests.

### DIFF
--- a/executor/pom.xml
+++ b/executor/pom.xml
@@ -17,7 +17,7 @@
 		<java.version>17</java.version>
 		<commons-text.version>1.14.0</commons-text.version>
 		<terraform-spring-boot-starter.version>1.2.2</terraform-spring-boot-starter.version>
-		<terrakube-client-starter.version>1.1.0</terrakube-client-starter.version>
+		<terrakube-client-starter.version>1.2.0-beta.1</terrakube-client-starter.version>
 		<commons-io.version>2.20.0</commons-io.version>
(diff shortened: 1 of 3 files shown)


alfespa17:
@capita-user can you help me with the sonar cloud issues, those are related to the new code from this PR 

https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&types=BUG&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&inNewCodePeriod=true&sinceLeakPeriod=true&id=AzBuilder_azb-server

PR #2657, opened 2025-11-03
fix: handle thread interruption in LogsServiceApi methods

alfespa17:
- Added `Thread.currentThread().interrupt()` to preserve interrupt status after catching `InterruptedException`.


alfespa17:
Related to #2645

Issue #2686, opened 2025-11-11
Unable to publish module using AZURE_SP_MI VCS type


Bbett:
### Bug description 🐞

I've setup an AZURE_SP_MI VCS provider and when I try to publish a module in the Registry using the VCS the module never publishes. 

I get the following error message in the api container. 
```
2025-11-11T18:25:21.585Z ERROR 1 --- [ryBean_Worker-6] i.t.a.p.s.module.ModuleRefreshJob        : Failed to refresh module acs on organization/user azure, error https://Repo_link: Authentication is required but no CredentialsProvider has been registered.
2025-11-11T18:25:21.585Z ERROR 1 --- [ryBean_Worker-6] i.t.a.p.s.module.ModuleRefreshJob        : There are no tags available for module Modulename on organization/user azure, error No versions found.
```

I reviewed the api code and believe the ModuleRefreshJob.java is the code that is trying to determine the versions/tags. 

I noticed that the case statement doesn't include AZURE_SP_MI like the GitServiceImpl.java code does. So the credentialsProvider is set to null. 

```
                case AZURE_DEVOPS:
                    credentialsProvider = new UsernamePasswordCredentialsProvider("dummy", vcs.getAccessToken());
                    break;
```
I think updating the case statement to include AZURE_SP_MI will fix the issue. There maybe other spots in the code that need to be updated as well. 

### Steps to reproduce

Setup a AZURE_SP_MI VCS provider and try to publish a module using the provider. 
The module will try to publish but fail. 
Then review the API logs. 

### Expected behavior

The AZURE_SP_MI VCS is able to be used to publish a module.

alfespa17:
Hello @Bbett 

Would you like to help with the fix?

I think you only need it here:

https://github.com/terrakube-io/terrakube/blob/ee114b9ab00c0835a04c8fe196e0dab17a9d18a0/api/src/main/java/io/terrakube/api/plugin/scheduler/module/ModuleRefreshJob.java#L161

https://github.com/terrakube-io/terrakube/blob/ee114b9ab00c0835a04c8fe196e0dab17a9d18a0/executor/src/main/java/io/terrakube/executor/service/scripts/ScriptEngineService.java#L132

I did this search to find those places

https://github.com/search?q=repo%3Aterrakube-io%2Fterrakube%20credentialsProvider%20%3D%20new%20UsernamePasswordCredentialsProvider(%22dummy%22&type=code

Bbett:
[...]

PR #2687, opened 2025-11-11
fix:add_azure_sp_mi_to_case

Bbett:
Added AZURE_SP_MI to the case statements for ModuleRefeshJob.java and ScriptEngineService.java so that the AZURE_SP_MI VCS type can be used with them.


Bbett:
Fix #2686

alfespa17:
Thank you @Bbett

PR #2717, opened 2025-11-23
feat: Ask Kubernetes to delete ephemeral jobs on completion

bittrance:
Kubernetes Jobs have an option to set a retention after which they are automatically deleted, called `ttlSecondsAfterFinished`. Using on this feature makes the code simpler, but it also removes speculative deletes from code paths in `ScheduleJob` where we did not actually know whether the job was an ephemeral job and so could fail because the service account had no perms on Jobs:
```
Failure executing: GET at: [https://10.96.0.1:443/apis/batch/v1/namespaces/terrakube/jobs/job-2.](https://10.96.0.1/apis/batch/v1/namespaces/terrakube/jobs/job-2.) Message: jobs.batch "job-2" is forbidden: User "system:serviceaccount:terrakube:default" cannot get resource "jobs" in API group "batch" in the namespace "terrakube". Received status: Status(apiVersion=v1, code=403, details=StatusDetails(causes=[], group=batch, kind=jobs, name=job-2, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=jobs.batch "job-2" is forbidden: User "system:serviceaccount:terrakube:default" cannot get resource "jobs" in API group "batch" in the namespace "terrakube", metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=Forbidden, status=Failure, additionalProperties={}).
```
At present, this was merely an annoying log entry, but I want next to propagate errors properly through `ScheduleJob/ExecutorService/...` so that they can be communicated to the user. In this case, the job may equally have been ephemeral and the role misconfigured, which should be communicated to the user.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
@@ -38,7 +38,6 @@ public class EphemeralExecutorService {
 
     public ExecutorContext sendToEphemeralExecutor(Job job, ExecutorContext executorContext) {
         final String jobName = "job-" + job.getId();
-        deleteEphemeralJob(job);
         log.info("Ephemeral Executor Image {}, Job: {}, Namespace: {}, NodeSelector: {}", ephemeralConfiguration.getImage(), jobName, ephemeralConfiguration.getNamespace(), ephemeralConfiguration.getNodeSelector());
         SecretEnvSource secretEnvSource = new SecretEnvSource();
         secretEnvSource.setName(ephemeralConfiguration.getSecret());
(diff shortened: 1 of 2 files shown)


api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
```diff
@@ -274,42 +272,20 @@ public ExecutorContext sendToEphemeralExecutor(Job job, ExecutorContext executor
...
                 .endSpec()
                 .endTemplate()
+                .withTtlSecondsAfterFinished(30)
                 .endSpec()
                 .build();
 
         log.info("Running ephemeral job");
         try {
-            kubernetesClient.batch().v1().jobs().inNamespace(ephemeralConfiguration.getNamespace()).createOrReplace(k8sJob);
+            kubernetesClient.batch().v1().jobs().inNamespace(ephemeralConfiguration.getNamespace()).resource(k8sJob).serverSideApply();
```
bittrance: The `createOrReplace()` method is deprecated. The new API follows the kubectl style.

PR #2736, opened 2025-11-30
feat: Preparing for executors throwing exceptions

bittrance:
This PR introduces a reasonably comprehensive test suite for the `ScheduleJob` service. The PR is mostly to review and discuss that the test suites assert on the important bits of the behavior. The PR also takes the opportunity to catch exceptions from `ExecutionService` since e.g. permission errors from the Kubernetes API can arise outside the Terrakube's control. More on propagate such errors in future PRs.

The tests are not pretty; each code path performs a large number of state interactions. It is my hope that further PRs can reduce the number of interaction significantly. Some examples of promising improvements to production code that would also simplify tests:

1. a minor change (see TODO) could make maintaining the cache synonymous with descheduling, meaning it could be removed from the test suite
2. stepid is looked up in all code paths, but only used in some
3. several code paths make multiple/unnecessary calls to repositories and vcs (see various `times(2)` below)

Also, if steps used UUID v7, they would not have to have both an int id and a ref id since UUID v7 is both unique and guaranteed to lexicographically sort in chronological order.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
@@ -1,8 +1,8 @@
 package io.terrakube.api.plugin.scheduler;
 
 import io.terrakube.api.plugin.scheduler.job.tcl.TclService;
+import io.terrakube.api.plugin.scheduler.job.tcl.executor.ExecutionException;
 import io.terrakube.api.plugin.scheduler.job.tcl.executor.ExecutorService;
-import io.terrakube.api.plugin.scheduler.job.tcl.executor.ephemeral.EphemeralExecutorService;
 import io.terrakube.api.plugin.scheduler.job.tcl.model.Flow;
 import io.terrakube.api.plugin.scheduler.job.tcl.model.FlowType;
 import io.terrakube.api.plugin.scheduler.job.tcl.model.ScheduleTemplate;
@@ -45,7 +45,6 @@ public class ScheduleJob implements org.quartz.Job {
     private final TemplateRepository templateRepository;
 
     public static final String JOB_ID = "jobId";
-    private final EphemeralExecutorService ephemeralExecutorService;
     private final GitLabWebhookService gitLabWebhookService;
 
     JobRepository jobRepository;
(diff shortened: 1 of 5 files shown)


alfespa17:
Hello, I will check the PR later today.

alfespa17:
> Also, if steps used UUID v7, they would not have to have both an int id and a ref id since UUID v7 is both unique and guaranteed to lexicographically sort in chronological order.

The above is because when we define a template for example this:

```yaml
flow:
  - name: "Plan"
    type: "terraformPlan"
    step: 100
  - name: "Apply"
    type: "terraformApply"
    step: 200
```

I had to put the step number because I did not find a way to use [snakeyaml](https://github.com/terrakube-io/terrakube/blob/a74318b73f742dff56d687bfa37c0a498e788575/api/pom.xml#L272) and load the yaml values in order, that is why you see the id with UUID and the int value for the step inside the database

alfespa17:
> 1. a minor change (see TODO) could make maintaining the cache synonymous with descheduling, meaning it could be removed from the test suite

Related to the above, I think this could be a good idea, in the end it is working in a similar to  way `descheduling`, we just need to remove the REDIS logs when using some specific status like pending, completed and rejected.

alfespa17:
> Seems odd that we do not remove the job from the scheduler?

To answer the above.

Terrakube creates 2 types of jobs:

1 - When the job is triggered for the first time, the job is schedule to run every every 30 seconds because you can trigger several jobs and all jobs need to be executed in order

https://github.com/terrakube-io/terrakube/blob/a74318b73f742dff56d687bfa37c0a498e788575/api/src/main/java/io/terrakube/api/rs/hooks/job/JobManageHook.java#L33

https://github.com/terrakube-io/terrakube/blob/a74318b73f742dff56d687bfa37c0a498e788575/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJobService.java#L86

Something like:

- job 1 is  running, we don't need to run the job every 30 seconds anymore, we need to run it only for updates so we remove it from quartz
- job 2 is waiting for job 1 to be completed, check in 30 seconds
- job 3 is waiting for job 2 to be completed, check in 30 seconds
- job X .......

2 - When the jos is updated, it will create a quartz job that `will be triggerd "now"` and will only run once

https://github.com/terrakube-io/terrakube/blob/a74318b73f742dff56d687bfa37c0a498e788575/api/src/main/java/io/terrakube/api/rs/hooks/job/JobManageHook.java#L42

https://github.com/terrakube-io/terrakube/blob/a74318b73f742dff56d687bfa37c0a498e788575/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJobService.java#L117

PR #2737, opened 2025-11-30
feat: Support GitHub App Tokens For Private Extensions

AviBackToBlack:
## Problem

1. PR #1207 introduced `PRIVATE_EXTENSION_VCS_ID_AUTH`, which lets the executor pull tokens for private extensions.  
   Unfortunately the implementation simply reuses `vcs.getAccessToken()`, so it only works for **GitHub OAuth** connections.  
   Workspaces configured with **GitHub App** connections never receive an installation token and cloning extensions breaks.

2. To request a **GitHub App** installation token we must know which repo the executor is cloning (the tools repo).  
   [ExecutorService](api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java) runs inside the API process and has no access to executor-side config, so it cannot discover the tools repo URL on its own.  
   Without propagating that URL/branch into the API service we can't call `TokenService.getAccessToken(...)`.

## Solution

- Expose `io.terrakube.tools.repository/branch` as environment variables for the API service (docker compose, dev scripts, telemetry compose, etc.) so Spring can inject them into [ExecutorService](api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java).
- Update [loadOtherEnvironmentVariables](api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java) to branch on `VcsConnectionType`:  
  - **GitHub OAuth** → reuse the stored access token (same behavior as before).  
  - **GitHub App** → call `tokenService.getAccessToken(toolsRepository, vcs)` and push the installation token into `TERRAKUBE_PRIVATE_EXTENSION_REPO_TOKEN`.
- Leave all other VCS types untouched; they still rely on their existing tokens/credentials.
- Verified locally with telemetry-compose: a workspace pointing to a private extensions repo via GitHub App now clones successfully.

## Testing

- [x] `mvn clean install`
- [x] `docker compose -f telemetry-compose/docker-compose.yml up` → run job with `PRIVATE_EXTENSION_VCS_ID_AUTH` (GitHub App) and confirm extensions repository is cloned.

## Related PRs

- Companion Helm changes: terrakube-io/terrakube-helm-chart#239

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
@@ -1,8 +1,8 @@
 package io.terrakube.api.plugin.scheduler;
 
 import io.terrakube.api.plugin.scheduler.job.tcl.TclService;
+import io.terrakube.api.plugin.scheduler.job.tcl.executor.ExecutionException;
 import io.terrakube.api.plugin.scheduler.job.tcl.executor.ExecutorService;
-import io.terrakube.api.plugin.scheduler.job.tcl.executor.ephemeral.EphemeralExecutorService;
 import io.terrakube.api.plugin.scheduler.job.tcl.model.Flow;
 import io.terrakube.api.plugin.scheduler.job.tcl.model.FlowType;
 import io.terrakube.api.plugin.scheduler.job.tcl.model.ScheduleTemplate;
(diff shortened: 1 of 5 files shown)


alfespa17:
@AviBackToBlack can you help me fixing the build please so I can merge it 

I think you need to add the new properties for the tests

https://github.com/terrakube-io/terrakube/blob/main/api/src/test/resources/application-test.properties

AviBackToBlack:
> @AviBackToBlack can you help me fixing the build please so I can merge it
> 
> I think you need to add the new properties for the tests
> 
> https://github.com/terrakube-io/terrakube/blob/main/api/src/test/resources/application-test.properties

This is done.

AviBackToBlack:
@alfespa17, thanks for merging! Just in case, there is related PR in the Helm Chart terrakube-io/terrakube-helm-chart#239 to expose tools repository URL to the API service so it would generate an installation token.

PR #2755, opened 2025-12-06
feat: Executors now throw errors for propagation to users

bittrance:
This PR is an exercise in writing tests and fixing what comes up:
- extract talking to local executors (named "persistent" but there is probably a better name?) into its own service for symmetry with ephemeral execution. Harmonize contract.
- both persistent and ephemeral executors set job status to `queue`; this now happens in the executor service so that all state transitions are collected in one place
- persistent job executor no longer mutates the `ExecutorContext` when scrubbing the log entry
- normalize the URL to executors, mostly to make life easier for tests, but it also means errors will be raised if it is not a valid URL

Clean up creation of Kubernetes Job:
- repair passing env vars to jobs
- repair and update job labels
- ensure all `K=V` vars are parsed the same way and propagate parse errors
- move default config map from /tmp to /data (/tmp should be for writable)

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
@@ -39,7 +39,7 @@ public class EphemeralExecutorService {
     EphemeralConfiguration ephemeralConfiguration;
 
     public void send(Job job, ExecutorContext executorContext) throws ExecutionException {
-        final String jobName = "job-" + job.getId();
+        final String jobName = String.format("job-%s-%s", job.getId(), System.currentTimeMillis());
         log.info("Ephemeral Executor Image {}, Job: {}, Namespace: {}, NodeSelector: {}", ephemeralConfiguration.getImage(), jobName, ephemeralConfiguration.getNamespace(), ephemeralConfiguration.getNodeSelector());
         SecretEnvSource secretEnvSource = new SecretEnvSource();
         secretEnvSource.setName(ephemeralConfiguration.getSecret());
--- a/api/src/test/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorServiceTest.java
+++ b/api/src/test/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorServiceTest.java
@@ -136,7 +136,7 @@ public class EphemeralExecutorServiceTest {
         subject().send(job(), context());
 
         verify(namespaced, times(1)).resource(job.capture());
-        assertEquals("job-4711", job.getValue().getMetadata().getName());
+        assertEquals("job-4711-", job.getValue().getMetadata().getName().substring(0, 9));
     }
 
     @Test


bittrance:
Going back to draft. I think I broke how beans are scheduled somehow; investigating.

api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java
```diff
@@ -39,7 +39,7 @@ public class EphemeralExecutorService {
     EphemeralConfiguration ephemeralConfiguration;
 
     public void send(Job job, ExecutorContext executorContext) throws ExecutionException {
-        final String jobName = "job-" + job.getId();
+        final String jobName = String.format("job-%s-%s", job.getId(), System.currentTimeMillis());
```
bittrance: I initially thought to use the step number for disambiguation, but it seems we don't have that here. The executor context step is the UUID which is not easy to find out, so not much info value.

alfespa17:
Thank you, I will check tomorrow morning when I have some free time from work

alfespa17:
Hello @bittrance I was checking and I think it looks good I only have one comment related to this line 

https://github.com/bittrance/terrakube/blob/7b8ef14d70e90902b441a2ed6492ae5cce8f87cd/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java#L356

The `setOutput` is expecting a URL where the logs are saved ( check the output property at the end of the payload)

Example:

GET https://terrakube-api.platform.local/api/v1/organization/d9b58bd3-f3fc-4056-a026-1163297e80a8/job?include=step

```json
{
    "data": [
        {
            "type": "job",
            "id": "1",
            "attributes": {
                "approvalTeam": null,
                "comments": null,
                "commitId": null,
                "createdBy": "admin@example.com",
                "createdDate": "2025-12-11T13:38Z",
                "output": null,
                "overrideBranch": "main",
                "planChanges": true,
                "refresh": true,
                "refreshOnly": false,
                "status": "failed",
                "tcl": null,
                "templateReference": "2db36f7c-f549-4341-a789-315d47eb061d",
                "terraformPlan": null,
                "updatedBy": "Internal",
                "updatedDate": "2025-12-11T13:39Z",
                "via": "UI"
            },
            "relationships": {
                "address": {
                    "data": []
                },
                "organization": {
                    "data": {
                        "type": "organization",
                        "id": "d9b58bd3-f3fc-4056-a026-1163297e80a8"
                    }
                },
                "step": {
                    "data": [
                        {
                            "type": "step",
                            "id": "c52d86b9-af56-48d2-a3fc-d805a88f8665"
                        }
                    ]
                },
                "workspace": {
                    "data": {
                        "type": "workspace",
                        "id": "5ed411ca-7ab8-4d2f-b591-02d0d5788afc"
                    }
                }
            }
        },
        {
            "type": "job",
            "id": "2",
            "attributes": {
                "approvalTeam": null,
                "comments": null,
                "commitId": "937b249da41c50a9e944f7205e95f694842c5394",
                "createdBy": "admin@example.com",
                "createdDate": "2025-12-11T13:42Z",
                "output": "",
                "overrideBranch": "main",
                "planChanges": false,
                "refresh": true,
                "refreshOnly": false,
                "status": "completed",
                "tcl": "ZmxvdzoKICAtIHR5cGU6ICJ0ZXJyYWZvcm1QbGFuIgogICAgbmFtZTogIlBsYW4iCiAgICBzdGVwOiAxMDAK",
                "templateReference": "42201234-a5e2-4c62-b2fc-9729ca6b4515",
                "terraformPlan": "/home/vscode/.terraform-spring-boot/local/state/d9b58bd3-f3fc-4056-a026-1163297e80a8/5ed411ca-7ab8-4d2f-b591-02d0d5788afc/2/f057cd80-1926-4543-a21d-c1db5ce37d81/terraformLibrary.tfPlan",
                "updatedBy": "Internal",
                "updatedDate": "2025-12-11T13:42Z",
                "via": "UI"
            },
            "relationships": {
                "address": {
                    "data": []
                },
                "organization": {
                    "data": {
                        "type": "organization",
                        "id": "d9b58bd3-f3fc-4056-a026-1163297e80a8"
                    }
                },
                "step": {
                    "data": [
                        {
                            "type": "step",
                            "id": "f057cd80-1926-4543-a21d-c1db5ce37d81"
                        }
                    ]
                },
                "workspace": {
                    "data": {
                        "type": "workspace",
                        "id": "5ed411ca-7ab8-4d2f-b591-02d0d5788afc"
                    }
                }
            }
        }
    ],
    "included": [
        {
            "type": "step",
            "id": "c52d86b9-af56-48d2-a3fc-d805a88f8665",
            "attributes": {
                "name": "Plan",
                "output": null,
                "status": "failed",
                "stepNumber": 100
            },
            "relationships": {
                "job": {
                    "data": {
                        "type": "job",
                        "id": "1"
                    }
                }
            }
        },
        {
            "type": "step",
            "id": "f057cd80-1926-4543-a21d-c1db5ce37d81",
            "attributes": {
                "name": "Plan",
                "output": "https://terrakube-api.platform.local/tfoutput/v1/organization/d9b58bd3-f3fc-4056-a026-1163297e80a8/job/2/step/f057cd80-1926-4543-a21d-c1db5ce37d81",
                "status": "completed",
                "stepNumber": 100
            },
            "relationships": {
                "job": {
                    "data": {
                        "type": "job",
                        "id": "2"
                    }
                }
            }
        }
    ]
}
```

That URL is built inside the executor and the logs are uploaded to the storage here:

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/executor/src/main/java/io/terrakube/executor/service/status/UpdateJobStatusImpl.java#L125

It has like different implementations:

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java#L245

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/executor/src/main/java/io/terrakube/executor/plugin/tfstate/azure/AzureTerraformStateImpl.java#L190

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/executor/src/main/java/io/terrakube/executor/plugin/tfstate/gcp/GcpTerraformStateImpl.java#L186

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/executor/src/main/java/io/terrakube/executor/plugin/tfstate/local/LocalTerraformStateImpl.java#L179

You could update the UI to show the message that you are saving there if it does not start with "https" and maybe use some like "internal error:" prefix

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/ui/src/domain/Jobs/Details.tsx#L44

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/ui/src/domain/Jobs/Details.tsx#L46

Also keep in mind that field only allows 1024 characters

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/api/src/main/resources/db/changelog/local/changelog-1.5.xml#L65

That will allow you to see the error here:

image

bittrance:
Good catch. There are currently two types of errors coming out of this flow: 1) communication errors resulting from a misconfigured setup or a broken executor, and 2) parse errors from `k=v` type variable values. The former will require action from people with access to logs, so there we just need some indication to users that the system is broken in a way that they are unlikely to be able to fix themselves. The problem is really with the second type. I suppose we could do something like introduce an `error:` scheme for inline messages in output, but is it worth the complexity? We could just add as much as can fit in the 128 chars of the step name?

alfespa17:
> Good catch. There are currently two types of errors coming out of this flow: 1) communication errors resulting from a misconfigured setup or a broken executor, and 2) parse errors from `k=v` type variable values. The former will require action from people with access to logs, so there we just need some indication to users that the system is broken in a way that they are unlikely to be able to fix themselves. The problem is really with the second type. I suppose we could do something like introduce an `error:` scheme for inline messages in output, but is it worth the complexity? We could just add as much as can fit in the 128 chars of the step name?

Adding error info inside the name property sound good to me, I did something similar with yaml errors in some point.

https://github.com/terrakube-io/terrakube/blob/a2abbe42b5c76201da5b272c9de9a911953261fd/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/TclService.java#L110

alfespa17:
I truncated the error using the name property [here](https://github.com/terrakube-io/terrakube/pull/2786/changes)

Issue #2768, opened 2025-12-08
MinIO STS Support


walnuts1018:
### Feature description 💡

I am using MinIO's Security Token Service (STS) feature.
I thought setting the `AwsEnableRoleAuth` environment variable to `true` would enable it, but when specifying MinIO's endpoint in `AwsEndpoint`, it seems `DefaultCredentialsProvider` cannot be used.

https://github.com/terrakube-io/terrakube/blob/92d3aca1b9b6bb2355f7ace723ab756abef6f736/api/src/main/java/io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.java#L65-L86

As a workaround, `AwsEndpoint` can be set to an empty string. However, doing so prevents settings like `skip_requesting_account_id` from being applied during Terraform backend construction. This results in an error like: `Error: validating provider credentials: retrieving caller identity from STS: operation error STS: GetCallerIdentity, https response error StatusCode: 400, RequestID: 187F4916A75BBAF3, api error InvalidParameterValue: Unsupported action GetCallerIdentity`.

https://github.com/terrakube-io/terrakube/blob/92d3aca1b9b6bb2355f7ace723ab756abef6f736/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java# L92-L109

Would it be possible to modify `StorageTypeAutoConfiguration.java` to evaluate `awsStorageTypeProperties.isEnableRoleAuthentication()` before the condition that checks `awsStorageTypeProperties.getEndpoint()`? I believe this change would allow enabling STS features for non-AWS providers like MinIO.

### Anything else?

Here are the Kubernetes manifests I am currently using:
https://github.com/walnuts1018/infra/tree/99c3bb11b85af7566f76a671c95e2b29ef359d3e/k8s/apps/terrakube

alfespa17:
Looks like a good idea @walnuts1018, feel free to create the PR with the improvement

walnuts1018:
@alfespa17 

I've created a PR #2769 for this fix. I'd appreciate your review.

PR #2769, opened 2025-12-08
feat: Prioritize DefaultCredentialsProvider to support STS on custom endpoints

walnuts1018:
Fixes: #2768

- Updates s3client initialization to prioritize DefaultCredentialsProvider regardless of the endpoint configuration.
  - This enables STS authentication for third-party S3 services like MinIO.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.java
+++ b/api/src/main/java/io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.java
@@ -62,7 +62,13 @@ public class StorageTypeAutoConfiguration {
                 break;
             case AWS:
                 S3Client s3client;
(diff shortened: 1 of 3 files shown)


alfespa17:
@walnuts1018 How will you make it work with minio STS without using the endpoint method?

Did you create a container to test? In a high level your are only changing the code order.

walnuts1018:
In my case, I specify the STS and S3 endpoints in `~/.aws/config`, so there is no need to explicitly set the endpoint in the code.

https://github.com/walnuts1018/infra/blob/99c3bb11b85af7566f76a671c95e2b29ef359d3e/k8s/apps/terrakube/configmap-aws.yaml#L21-L31

Also, since the S3 and STS endpoints are separate when using STS with the MinIO Operator, simply using `endpointOverride` seems difficult to implement correctly.

I believe this modification is sufficient to enable MinIO STS with minimal changes.

I haven't tested this locally yet. Since it involves only a simple reordering of conditions, I am confident it will work, but I am happy to perform local testing if necessary.

PR #2770, opened 2025-12-10
feat: Add sortable columns to variable tables in UI

denniswebb:
We have a few workspaces with 70+ variables. The unsorted variables made maintenance a pain. I employed Claude to add sorting for me.

Adds default sorting by variable name (key) and enables sorting on category columns for all variable tables:
- Workspace variables table
- Collection variables table (read-only view in workspace)
- Global variables table (read-only view in workspace)
- Organization global variables table (settings)

Tables are sorted by key in ascending order by default. Users can click column headers to change sort order.


alfespa17:
I did a quick test using my devcontinaer and I think it looks good

imagen

PR #2794, opened 2025-12-17
feat: Proper error propagation on execution failure

bittrance:
Some further testing of #2736 indicated that if the executor is unreachable (and probably several other classes of runtime errors) the error is actually suppressed because of `@Transactional` on `ExecutorService.execute`. However, I don't see much point in making that method transactional, since it is always called from `ScheduledJob` which is already transactional. Thus, this PR removes the inner transaction so that errors can flow freely. There is anyway very little mutation going on inside.

Also, that PR did not update commit statuses on error. Reasonably, if we fail to execute at all, that should be `unknown` commit status.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/persistent/PersistentExecutorService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/persistent/PersistentExecutorService.java
@@ -12,7 +12,6 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.http.client.reactive.ReactorClientHttpConnector;
 import org.springframework.stereotype.Service;
 import org.springframework.web.reactive.function.client.WebClient;
-import org.springframework.web.reactive.function.client.WebClientResponseException;
 
 import io.terrakube.api.plugin.scheduler.job.tcl.executor.ExecutionException;
 import io.terrakube.api.plugin.scheduler.job.tcl.executor.ExecutorContext;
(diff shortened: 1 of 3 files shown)


api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/persistent/PersistentExecutorService.java
```diff
@@ -61,7 +60,7 @@ public void send(Job job, ExecutorContext executorContext) throws ExecutionExcep
                     .retrieve()
                     .toEntity(ExecutorContext.class)
                     .block();
-        } catch (URISyntaxException | WebClientResponseException ex) {
+        } catch (Exception ex) {
             throw new ExecutionException(ex);
         }
```
bittrance: Communication errors would tend to be runtime errors which we also want to convert to checked exceptions.

alfespa17 (approved):

alfespa17:
> Also, that PR did not update commit statuses on error. Reasonably, if we fail to execute at all, that should be unknown commit status.

`unknow commit status` will be marked as error here:

### Github

https://github.com/terrakube-io/terrakube/blob/c581ea69d1459a479735775a3ca6e547e89e05df/api/src/main/java/io/terrakube/api/plugin/vcs/provider/github/GitHubWebhookService.java#L167

### Gitlab

https://github.com/terrakube-io/terrakube/blob/c581ea69d1459a479735775a3ca6e547e89e05df/api/src/main/java/io/terrakube/api/plugin/vcs/provider/gitlab/GitLabWebhookService.java#L510

Issue #805, opened 2024-04-12
Runs with no changes do not need confirmation


calvinbui:
### Bug description 🐞

When a CLI run has no changes, there shouldn't be any prompts and the plan should complete.

![image](https://github.com/AzBuilder/terrakube/assets/3604363/a50c38ad-2dc5-41fa-bcee-251b8e401903)

![image](https://github.com/AzBuilder/terrakube/assets/3604363/ab9be29a-29e9-4288-9e25-16c0f272de47)

### Steps to reproduce

1. Apply a run with no changes
2. View the CLI or UI

### Expected behavior

No prompt. This is how terraform cloud does it:

![image](https://github.com/AzBuilder/terrakube/assets/3604363/4fcadae8-ac12-4979-b510-7cb2e900285f)

### Anything else?

I am evaluating terrakube via the minikube helm deployment.

alfespa17:
This can be an improvement I will find out how to add this validation.

PR #2796, opened 2025-12-20
feat: Empty plan completes immediately

bittrance:
This PR removes one case from the central schedulejob state machine by saying that a job where tofu/terraform reports "no changes" is immediately completed. This allows us to also simplify redis interactions.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
@@ -63,7 +63,7 @@ public class ScheduleJob implements org.quartz.Job {
 
     ScheduleJobService scheduleJobService;
 
-    RedisTemplate redisTemplate;
+    RedisTemplate<String, Object> redisTemplate;
 
     GitHubWebhookService gitHubWebhookService;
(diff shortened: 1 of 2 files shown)


bittrance:
Eh? I reorganized Redis usage with essentially no functional difference and the test fails because there is no Redis?
```
Error:  Errors: 
Error:    CollectionTests.testCollectionPriorityAsOrgMember:508 » RedisConnectionFailure Cannot get Jedis connection
```
Mmm. Why mockbean is deprecated, I suppose. Fixed now.

bittrance:
It is worth noting that there is a defect in the logic both before and after this PR, in that a "no-change" plan will abort the whole job even if there are pertinent steps later in the flow. For example:
image

alfespa17:
> It is worth noting that there is a defect in the logic both before and after this PR, in that a "no-change" plan will abort the whole job even if there are pertinent steps later in the flow. 

To be honest I am not sure if we should allow to execute something else when terraform/tofu plan does not have any changes, some time ago someone complained about that [here](https://github.com/terrakube-io/terrakube/issues/805), that is why I created the "noChanges" status and mark everything as `notExecuted` when that happens

PR #2797, opened 2025-12-20
feat: Dark mode improvements

bittrance:
I noticed that the ID on the Workspace details page remained dark gray in dark mode, making it hard to read:
image
This led me to the `TeamPermissions.css`. I'm not a front-end maestro, but it seems odd that this CSS had effect on other pages? However, as I navigated to that control, it turns out to render poorly in darkmode (headers and tooltips almost invisible and using too-bright style):
image
This PR removes the custom table look-alike and replaces it with an actual table. This way, we can remove the whole CSS. Hopefully, it does not cause unexpected changes elsewhere, given that it seems to have effect on at least Workspace IDs:
image
The team permission UI now looks like this in dark mode:
image
Also, standardize delete buttons to all say "Delete" and use trash can Icons.


alfespa17:
I did a quick test and this looks good, thank you for your help @bittrance

PR #2798, opened 2025-12-21
bugfix: Team tokens are now displayed after creation

bittrance:
I noticed that when creating a team token, the token was never displayed. This PR fixes that by switching to reuse the dialog from PAT creation.

The modals differ in one important aspect: the team one support short-lived tokens (expiry < 1 day). I'm a bit skeptical of this feature in the UI (tho I can see how an API client might want to create short-lived tokens for delegating permissions). If we want to keep it, I think we should introduce short-lived tokens for PATs as well?

Also, this PR prefers the newer apiWrapper style of REST API interaction as it makes for somewhat simpler conde. The EditTeam seems to have been the only user of the (obsolete?) axiosClientAuth so I removed it. Tell me if this is in line with long-term plans?

PR #2815, opened 2025-12-22
feat: Unify token listing by reusing PAT listing for team tokens

bittrance:
Follow-up on #2798. Partly, this gives both tokens the same look and feel, but most to reduce the amount of code needed.


bittrance:
I can repeat the build error locally and it does not occur on main, but I struggle to see how it is related to this PR?

alfespa17:
Ignore the error, it is related to the API, you are only changing the UI in this PR.

alfespa17:
Checking the log it is related to the tofu index that fails from time to time because it is using a public endpoint, so it is not related to your code

PR #2817, opened 2025-12-23
feat: Avoid null execution_mode and clarify it is informational

bittrance:
The execution_mode concept was somewhat confusing and if you create an org, it will default to `null` execution_mode, which then propagates to workspaces, showing empty on the workspace details page.

This PR ensures orgs default to remote (which is how null is interpreted in practice) and that the UI propagates this to workspaces on creation (but CLI-driven workspaces are set as local, regardless of org setting). In order to reduce data model complexity, we also rewrite all nulls to `remote`. Also, clarify that this field is not used programmatically.

This is not the most important PR in the history of Terrakube. Since it is somewhat invasive and we are getting close to a 2.29 release, it might make sense to save this one till after the release?


alfespa17:
I think we can merge this , it is just a small change and I don't think it will create any major issue

PR #2818, opened 2025-12-24
fix(db): Use timestamp type for audit fields to ensure second-level precision (#2787)

### LINKED ISSUE #2787: Console calculates time from top of the minute, not by second
### Bug description 🐞

### Steps to reproduce

### Expected behavior

### Example repository

_No response_

### Anything else?

denniswebb:
## Summary
- Fix timestamp precision issue causing "X seconds ago" to display incorrectly in UI
- Change Liquibase schema from `datetime` to `timestamp` type for all audit columns
- Affects 21 tables across the application

## Problem
The UI shows "Triggered from UI 42 seconds ago" even when a job was just triggered. This happens because:
- Liquibase migrations use `datetime` type for `created_date` and `updated_date` columns
- On SQL Server/Azure SQL, `datetime` type has only ~3.33ms precision
- This causes timestamp rounding, losing sub-second precision
- The frontend luxon library (`DateTime.fromISO().toRelative()`) works correctly, but receives imprecise data from the backend

## Root Cause
Found in `api/src/main/resources/db/changelog/local/changelog-1.5.xml:12`:
```xml
<column name="created_date" type="datetime"/>
```

The `datetime` type maps differently across databases:
- **SQL Server/Azure SQL**: `datetime` with ~3.33ms precision ❌
- **PostgreSQL**: `TIMESTAMP` with microsecond precision ✓
- **MySQL <5.6.4**: `datetime` with 1-second precision ❌
- **H2**: `TIMESTAMP` with full precision ✓

## Solution
Created new Liquibase migration `changelog-2.30.0-fix-timestamp-precision.xml` that:
- Changes all `created_date` and `updated_date` columns from `datetime` to `timestamp` type
- Uses `modifyDataType` to safely convert existing columns without data loss
- Covers 21 tables: history, job, workspace, organization, team, template, vcs, module, provider, ssh, agent, global_var, tag, team_token, pat, action, webhook, collection, cli_address, webhook_event, project

The `timestamp` type maps to:
- **PostgreSQL**: `TIMESTAMP` (microsecond precision) ✓
- **SQL Server/Azure SQL**: `datetime2(7)` (100ns precision) ✓
- **MySQL**: `TIMESTAMP` (microsecond precision) ✓
- **H2**: `TIMESTAMP` (nanosecond precision) ✓

## Testing Plan
1. Deploy migration (automatic on application startup via Liquibase)
2. Create a new job immediately after the top of a minute
3. Verify UI shows "Triggered from UI 1 seconds ago" (or similar precise value)
4. Verify existing jobs continue to display correctly

## Impact
- **No data loss**: Existing timestamps are preserved during column type conversion
- **Backward compatible**: Frontend code requires no changes
- **Database agnostic**: Works correctly across PostgreSQL, SQL Server, MySQL, and H2
- **Performance**: No performance impact expected (same index behavior)

## Files Changed
- ✨ `api/src/main/resources/db/changelog/local/changelog-2.30.0-fix-timestamp-precision.xml` (new migration)
- 📝 `api/src/main/resources/db/changelog/changelog.xml` (added migration reference)
- 📋 `Auto Run Docs/issue-2787-1.md` (investigation and fix documentation)

Fixes #2787

### DIFF
--- a/api/src/main/resources/db/changelog/changelog.xml
+++ b/api/src/main/resources/db/changelog/changelog.xml
@@ -91,4 +91,5 @@
     <include file="/db/changelog/local/changelog-2.28.0-project-support.xml" />
     <include file="/db/changelog/local/changelog-2.29.0-branch-size.xml" />
     <include file="/db/changelog/local/changelog-2.29.0-execution-mode.xml" />
+    <include file="/db/changelog/local/changelog-2.30.0-fix-timestamp-precision.xml" />
 </databaseChangeLog>
--- a/api/src/main/resources/db/changelog/local/changelog-2.30.0-fix-timestamp-precision.xml
+++ b/api/src/main/resources/db/changelog/local/changelog-2.30.0-fix-timestamp-precision.xml
@@ -1,0 +1,147 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+
+    <!--
+        Fix for issue #2787: Console time calculation precision
+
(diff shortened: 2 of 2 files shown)


denniswebb:
Closing this PR - investigation was based on incorrect assumption. The issue occurs on PostgreSQL which already has microsecond precision, so the database schema change won't fix the root cause. Need to re-investigate the actual source of the timestamp precision problem.

PR #2820, opened 2025-12-24
bugfix: Strengthen protection for Job approval

bittrance:
The TeamApproveJob check had two shortcomings:
- it would allow the designated team to make any status change
- it would allow approving jobs before they were awaiting approval

Additionally, Elide checks on fields supercede checks at class level so there was no check to allow a team with job management perms to cancel a job. This is now fixed.

Finally, the check now ensures it opines only about changes to status, should it end up on some other field by mistake.

Apologies for the chaotic diff on JobTests.

### DIFF
--- a/api/src/test/java/io/terrakube/api/JobTests.java
+++ b/api/src/test/java/io/terrakube/api/JobTests.java
@@ -9,12 +9,17 @@ import org.junit.jupiter.api.Test;
 import org.mockito.MockitoAnnotations;
 import org.springframework.http.HttpStatus;
 
+import graphql.AssertException;
 import io.terrakube.api.rs.Organization;
+import io.terrakube.api.rs.job.Job;
 import io.terrakube.api.rs.job.JobStatus;
 import io.terrakube.api.rs.team.Team;
 import io.terrakube.api.rs.template.Template;
 import io.terrakube.api.rs.workspace.Workspace;
 
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Optional;
 import java.util.UUID;
 
 import static io.restassured.RestAssured.given;
@@ -121,6 +126,20 @@ class JobTests extends ServerApplicationTests {
         return templateRepository.save(template);
     }
 
+    private void awaitJobStatus(String jobId, JobStatus status, Duration timeout) {
+        Instant deadline = Instant.now().plusMillis(timeout.toMillis());
+        while(Instant.now().isBefore(deadline)) {
+            Optional<Job> maybeJob = jobRepository.findById(Integer.parseInt(jobId));
+            if(maybeJob.filter(job -> status.equals(job.getStatus())).isPresent()) {
+                return;
+            }
+            try {
+                Thread.sleep(100);
+            } catch (InterruptedException e) {}
+        }
+        throw new AssertException(String.format("Timeout waiting for job %s to reach status %s", jobId, status));
+    }
+
     @Test
     void createJobAsOrgMember() {
         devsManageJobs(true);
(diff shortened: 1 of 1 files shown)


api/src/main/java/io/terrakube/api/rs/job/Job.java
```diff
@@ -46,7 +46,7 @@ public class Job extends GenericAuditFields {
     @Column(name = "comments")
     private String comments;
 
-    @UpdatePermission(expression = "team approve job OR user is a super service")
+    @UpdatePermission(expression = "team approve job OR team limited manage job OR team manage job OR user is a super service")
```
bittrance: My understanding is that elide does not check class level perms if there are field-level perms, so we need to repeat perms here or users with job manage perms cannot cancel the job.

alfespa17: Elide will first check  field level permissions and later class level permission, for example if you have are using `"@ReadPermission"` inside a field that return `"false"` but you are using `"@ReadPermission"` at class level that return true elide will remove the field with `"false"`  from the response but will return all the others because the class level permission has `"true"`

api/src/test/java/io/terrakube/api/JobTests.java
```diff
@@ -1,151 +1,274 @@
...
+        template.setName("ze-template");
+        template.setTcl(Base64.encodeBytes(tcl.getBytes()));
+        template.setOrganization(organization);
+        return templateRepository.save(template);
+    }
+
+    private void awaitJobStatus(String jobId, JobStatus status, Duration timeout) {
+        Instant deadline = Instant.now().plusMillis(timeout.toMillis());
+        while(Instant.now().isBefore(deadline)) {
+            if(!jobRepository.findAllByWorkspaceAndStatusInOrderByIdDesc(workspace, List.of(status)).isEmpty()) {
```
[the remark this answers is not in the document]
alfespa17: @bittrance you could use "findById" that return Optional<Job>

```
if(!jobRepository.findById(Integer.parseInt(jobId)).get().getStatus().equals(status)) {
                return;
            }
```

Image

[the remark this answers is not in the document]
bittrance: Aha, nice! Fixed.

api/src/main/java/io/terrakube/api/rs/job/Job.java
bittrance: If you apply this diff:
```
iff --git a/api/src/main/java/io/terrakube/api/rs/job/Job.java b/api/src/main/java/io/terrakube/api/rs/job/Job.java
index 4b60f62f..7be5bc3b 100644
--- a/api/src/main/java/io/terrakube/api/rs/job/Job.java
+++ b/api/src/main/java/io/terrakube/api/rs/job/Job.java
@@ -46,7 +46,7 @@ public class Job extends GenericAuditFields {
     @Column(name = "comments")
     private String comments;

-    @UpdatePermission(expression = "team approve job OR team limited manage job OR team manage job OR user is a super service")
+    @UpdatePermission(expression = "team approve job OR user is a super service")
     @Enumerated(EnumType.STRING)
     private JobStatus status = JobStatus.pending;
```
and run the job tests you will see a failure despite matchin annotations on class. See "Permission Annotations" under https://elide.io/pages/guide/v4/03-security.html .

alfespa17: You are correct, this should be the correct expression `team approve job OR team limited manage job OR team manage job OR user is a super service` the other one creates errors

alfespa17 (approved):

PR #2825, opened 2025-12-29
Sketch of on-demand token refresh would look like

bittrance:
This sketch shows how to switch from a model where we continually hammer Vcs provider OAuth APIs for tokens to one in which tokens are refreshed as needed when tokens are requested. There are two motivations for this change:
- a large Terrakube installation would generate a lot of token requests, risking rate limiting
- a lot of those requests will be pointless, for the requested tokens will not have an IaC run during its lifetime (this is based on the assumption that many large organizations would have many Terrakube orgs, each with its own client/app/creds)
- requesting tokens on schedule makes it harder to communicate failures to the user. By contrast, an on-demand refresh failure can be communicated via the run UI where ordinary users can act on it

This is a draft and opened to facilitate discussing the merits of this change. There is lots of details left to fix and clean up.

Notes:
- In order to accommodate the registry, a new endpoint `/access-token/v1/vcs/{vcsId}` would be added. This avoids having to slow down the "ordinary" endpoint for reading Vcs:es (which anyway should not hand out secrets unless they are requested).
- There is some additional work to wire up refresh of tokens for pki-based GitHub apps and GitHub OAuth apps with expiring access tokens, but most of the code is there in `GitHubTokenService` so it should be straight-forward


registry/src/main/java/io/terrakube/registry/service/module/ModuleServiceImpl.java
```diff
@@ -138,33 +138,11 @@ public void updateModuleDownloadCount(String organizationName, String moduleName
...
     }
-    
-    private GitHubAppToken getGitHubAppTokenInformation(String vcsClientId, String repository_source) {
-        URI uri = URI.create(repository_source);
-        String owner = uri.getPath().split("/")[1];
-        List<GitHubAppToken> gitHubAppTokens = terrakubeClient.getGitHubAppTokenByVcsIdAndOwner(owner, vcsClientId).getData();
-        if (gitHubAppTokens.isEmpty())  return null;
-
-        return gitHubAppTokens.get(0);
-    }
```
bittrance: As far as I can understand, this code is just working around the fact that the ordinary token refresh is not implemented for pki-based apps.

registry/src/main/java/io/terrakube/registry/service/module/ModuleServiceImpl.java
```diff
@@ -105,7 +105,7 @@ public String getModuleVersionPath(String organizationName, String moduleName, S
                     module.getRelationships().getVcs().getData().getId());
             vcsType = vcsInformation.getAttributes().getVcsType();
             vcsConnectionType = vcsInformation.getAttributes().getConnectionType();
-            accessToken = getAccessToken(organizationId, vcsInformation.getId(), moduleSource);
+            accessToken = terrakubeClient.getAccessToken(vcsInformation.getId(), moduleSource);
```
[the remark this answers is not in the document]
alfespa17: Maybe you can encrypt value in the API response and the registry could use `io.terrakube.client.secretKey=${InternalSecret}` to decrypt it. 

The InternalSecret [should be different in each Terrakube installation ](https://docs.terrakube.io/getting-started/deployment/token-security)

api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleVcs.java
alfespa17: You will also need to delete the job information inside the quartz table, I did something similar in the past you could check 

https://github.com/terrakube-io/terrakube/blob/main/api/src/main/resources/db/changelog/local/changelog-2.27.2-delete-legacy-module-refresh.xml

You could see https://github.com/terrakube-io/terrakube/pull/2350 for more information

PR #2851, opened 2026-01-14
feat: implement job history cleanup based on `KEEP_JOB_HISTORY` variable for local execution

### LINKED ISSUE #2849: KEEP_JOB_HISTORY not used for local terraform execution
### Bug description 🐞

The local execution are not cleaning the workspace jobs list when using KEEP_JOB_HISTORY, it looks like it is only working with remote execution when using this change https://github.com/terrakube-io/terrakube/pull/2750

### Steps to reproduce

Use local terraform executions

### Expected behavior

Clean the job table automatically

### Example repository

_No response_

### Anything else?

_No response_

alfespa17:
- Introduced `getHistoryLimit` and `deleteOldJobs` methods to manage history retention using workspace variables.
- Added `VariableRepository` for retrieving and processing workspace parameters.
- Enhanced logging to trace cleanup operations.

fix #2849

### DIFF
--- a/api/src/test/java/io/terrakube/api/TfcApiTests.java
+++ b/api/src/test/java/io/terrakube/api/TfcApiTests.java
@@ -443,7 +443,7 @@ class TfcApiTests extends ServerApplicationTests {
 
         remainingJobs = jobRepository.findAllById(Arrays.asList(firstJob.get().getId(), secondJob.get().getId(), thirdJob.get().getId()));
 
-        // Verify that some jobs were deleted due to KEEP_JOB_HISTORY limit
+        // Verify that some jobs were deleted due to KEEP_JOB_HISTORY limit, firstJob should have been deleted only secondJOb and thirdJob should exists
         assertThat(remainingJobs.size()).isEqualTo(2);
     }
 }


api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -680,9 +687,52 @@ StateData createWorkspaceState(String workspaceId, StateData stateData) {
...
+    private int getHistoryLimit(Job job) {
+        AtomicInteger workspaceHistory = new AtomicInteger();
+        workspaceHistory.set(0);
+
+        var variables = variableRepository.findByWorkspace(job.getWorkspace());
+        variables.ifPresent(variableList -> variableList.forEach(variable -> {
+            if (variable.getKey().equals("KEEP_JOB_HISTORY") && variable.getCategory() == Category.ENV) {
+                workspaceHistory.set(Integer.parseInt(variable.getValue()));
+            }
+        }));
```
bittrance: You can `return variables.stream().filter(v -> ...).map(v -> Integer.parseInt(v)).findFirst();`.

alfespa17:
Thank you for the feedback @bittrance I will rewrite the code using the above suggestions

alfespa17:
I did a small refactor I think it looks better now @bittrance

PR #2879, opened 2026-01-22
feat: add bypassQueue flag for webhook-triggered plan-only jobs

### LINKED ISSUE #2850: Plan-only jobs should not queue behind other jobs
**Summary**: Jobs that only run terraform plan are read-only and don't modify state. These jobs should start immediately when a runner is available, rather than waiting in a per-workspace queue.

**Problem**: Currently all jobs in a workspace queue sequentially by creation order. This makes sense for apply operations (state must be modified in order), but plan-only jobs have no such constraint.

The worst case: a job is sitting in waitingApproval after completing its plan. Any new PR opened against that workspace has its plan blocked until someone approves or rejects the pending job - even though the new job only needs to run a read-only plan.

With ephemeral runners, this is especially wasteful - runners are available and ready, but plan jobs sit idle waiting for unrelated work to complete.

**Proposed Behavior**:
  - Plan-only jobs (templates containing only terraformPlan/terraformPlanDestroy) should dispatch immediately to available runners
  - No queuing behind other plan-only jobs or jobs waiting for approval
  - Apply/destroy jobs should continue to queue sequentially (state consistency)
  - Plan-only jobs should only wait if an apply/destroy is actively running

**Use Case**: PR checks. Multiple developers working on the same infrastructure repo should get immediate plan feedback on their PRs without waiting for each other.

I'm open to work on this if it is agreed that this is needed.

denniswebb:
Webhook-triggered PR checks using plan-only templates can now bypass the normal per-workspace queue. This enables immediate PR plan feedback without waiting for unrelated jobs (especially those stuck in waitingApproval).

Changes:
- Add TclService.isTemplatePlanOnly() to detect plan-only templates at runtime
- Add TclService.getFlowTypeForStep() to get flow type for a specific step
- Update ScheduleJob queue logic to check template type and allow bypass
- Add JobRepository query method for finding running jobs in workspace

Behavior:
- Plan-only webhook jobs proceed immediately past pending/waitingApproval jobs
- Plan-only webhook jobs can run in parallel with other plan-only jobs
- Plan-only webhook jobs still wait for active apply/destroy/customScripts operations
- CLI jobs and jobs with apply/destroy use normal queue behavior

The bypass is computed at runtime by parsing the template TCL rather than being persisted to the database. This avoids schema changes and keeps the bypass logic centralized in the scheduler.

Closes #2850

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
@@ -427,7 +427,7 @@ public class ScheduleJob implements org.quartz.Job {
 
     private boolean isActiveApplyOrDestroyRunning(Workspace workspace, int currentJobId) {
         Optional<List<Job>> runningJobs = jobRepository.findByWorkspaceAndStatusInAndIdLessThan(
-                workspace, Arrays.asList(JobStatus.running), currentJobId);
+                workspace, Arrays.asList(JobStatus.running, JobStatus.queue), currentJobId);
 
         if (!runningJobs.isPresent() || runningJobs.get().isEmpty()) {
             return false;


api/src/main/resources/db/changelog/local/changelog-2.30.0-bypass-queue-job.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <changeSet id="2-30-0-bypass-queue-1" author="terrakube">
```
alfespa17: Hello @denniswebb after creating the new field you will need to update all previous records inside the database to put the `false` value, if you dont do that the API will break because it is expecting a boolean inside the entity, for example [here](https://github.com/terrakube-io/terrakube/blob/669165aa5b7c5ac98b9a6ea1e2e99c45bd21ad14/api/src/main/java/io/terrakube/api/rs/job/Job.java#L94)

.github/workflows/release.yml
alfespa17: Maybe you can put this change in a different PR if you want to improve it

denniswebb:
Thanks @alfespa17\! Good catch. I've added the explicit UPDATE statement to set `bypass_queue=false` for existing records, following the pattern used in other changelogs like `changelog-2.21.0-auto-apply.xml`.

denniswebb:
@alfespa17 I changed the approach some because I really didn't like the DB update stuff. I mean there's no need to do a DB update IMO. Now it does not need to touch the DB.

Testing shows it works as expected.
image

alfespa17:
Thank you, I will check later.

api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
```diff
@@ -410,4 +424,33 @@ private void updateJobStatusOnVcs(Job job, JobStatus jobStatus) {
...
+        Optional<List<Job>> runningJobs = jobRepository.findByWorkspaceAndStatusInAndIdLessThan(
+                workspace, Arrays.asList(JobStatus.running), currentJobId);
+
+        if (!runningJobs.isPresent() || runningJobs.get().isEmpty()) {
+            return false;
+        }
+
+        for (Job runningJob : runningJobs.get()) {
+            Optional<Step> runningStep = stepRepository.findByJobId(runningJob.getId()).stream()
+                    .filter(step -> step.getStatus().equals(JobStatus.running))
```
alfespa17: It should consider the queue status here I think

alfespa17:
There is one status missing but I will add it in another PR, in general I think this looks good

PR #2920, opened 2026-02-03
add automatic creation of a workspace in terraform state

jplorier:
Adding post init functions that will select the existing workspace in the terraform state with the name of the terrakube workspace.
If the workspace doesn't exist in the state, checks if the default workspace has resources. If resources don't exist, will create the new workspace, if exist, will migrate the resources to the new workspace.

### DIFF
--- a/executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
+++ b/executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
@@ -19,6 +19,8 @@ import org.springframework.stereotype.Service;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
```diff
@@ -470,9 +472,183 @@ private String executeTerraformInit(TerraformJob terraformJob, File workingDirec
...
+
+    private void migrateDefaultStateToNewWorkspace(String workspaceName, File workingDirectory,
+                                                    TerraformJob terraformJob, Consumer<String> output,
+                                                    AnsiFormat colorMessage)
+            throws IOException, InterruptedException {
+        File tempStateFile = new File(workingDirectory, ".terrakube_temp_state.json");
+        
+        try {
+            // Step 1: Ensure we're on default workspace
+            executeCommand("terraform workspace select default", workingDirectory, terraformJob, output);
```
bittrance: I think there is tooling to handle selection of binary? For sure OpenTofu also supports workspaces.

executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
```diff
@@ -470,9 +472,183 @@ private String executeTerraformInit(TerraformJob terraformJob, File workingDirec
...
+            // Step 3: Create new workspace
+            output.accept(colorize("  → Creating new workspace '" + workspaceName + "'...", colorMessage));
+            executeCommand("terraform workspace new " + workspaceName, workingDirectory, terraformJob, output);
+            
+            // Step 4: Push state to new workspace with -force to override lineage check
+            // The -force flag is necessary because the new workspace has a different lineage UUID
+            // than the state we're pushing from the default workspace
+            output.accept(colorize("  → Migrating state to '" + workspaceName + "' (forcing lineage update)...", colorMessage));
+            String pushCommand = "terraform state push -force " + tempStateFile.getAbsolutePath();
+            executeCommand(pushCommand, workingDirectory, terraformJob, output);
```
bittrance: I'm not that experienced with workspaces, but could we not use `workspace new -state foo.tfstate` instead?

executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
```diff
@@ -470,9 +472,183 @@ private String executeTerraformInit(TerraformJob terraformJob, File workingDirec
...
+
+    private boolean checkWorkspaceExists(String workspaceName, File workingDirectory,
+                                         TerraformJob terraformJob, Consumer<String> output)
+            throws IOException, InterruptedException {
+        String command = "terraform workspace list";
+        String workspaceList = executeCommandWithOutput(command, workingDirectory, terraformJob);
+        
+        // Check if workspace name appears in the list
+        // Format is usually: "  default", "* current", "  workspace-name"
+        for (String line : workspaceList.split("\n")) {
```
bittrance: We build a string by concatenating lines only to immediately split them again here. In fact, you don't need the `executeCommandWithOutput()` method at all: you can just pass a closure to `executeCommand()` like so:
```
   private void generator(Consumer<String> output) {
        output.accept("foo");
    }

    private void consumer() {
        generator(s -> System.out.println(s));
    }
```

alfespa17:
I am changing this PR to draft, I will review it in more detail

alfespa17:
There is an open issue related to use "terraform workspace" command in the terraform library

https://github.com/terrakube-io/terraform-spring-boot/issues/30

alfespa17:
I was checking this PR and I wont merge it. 

The best way to implement this feature will be to add it firts to the terraform/tofu library [here](https://github.com/terrakube-io/terraform-spring-boot), once it is implemented you can create a new PR to add the feature here.

There are things in this PR that wont work for example this method `executeCommandWithOutput`, terraform or tofu is not installed directly in the container path, it is generated dynamically when using the library, so that method will simply failed because terraform or tofu wont be found.

```
ProcessBuilder processBuilder = new ProcessBuilder("/bin/sh", "-c", command);
```

This is an example how the library generate the correct PATH for tofu and terraform each time a command is executed

https://github.com/terrakube-io/terraform-spring-boot/blob/518bb18a729bcc3a324d93870902a68577198ee8/terraform-client/src/main/java/io/terrakube/terraform/TerraformClient.java#L290

Please create a pull request to add the `terraform workspace` command to the terraform/tofu library and later we can check the implementation for this repo

PR #2921, opened 2026-02-03
feat: edit collection variables using a modal

jplorier:
Allow to edit variables in a collection the same way it's done in the workspace variable.
Pease review the usage of a modal.


alfespa17:
I did a quick test and it looks like it is working.

image

alfespa17 (approved):

PR #2923, opened 2026-02-03
Split log view

jplorier:
In order to facilitate the view of the run output, I'm splitting the output in the different phases into separate frames. This will allow to expand only the section you are interested in viewing.
Now the init phase, the resource scanning for the graph and the output of either plan or apply are shown in separate frames.
Also introduces auto scroll to avoid manual scrolling to keep the process output in view.


alfespa17:
I was testing this branch but I dont see any difference in the UI @jplorier 

imagen

alfespa17:
I am closing this one, I merged https://github.com/terrakube-io/terrakube/pull/2937 with similar functionalities

Issue #2932, opened 2026-02-08
50% chance of 0x00 padding added to dynamic credentials JWKS


mattwillsher-rvu:
### Bug description 🐞

Hi,

I've been setting up Terrakube in a somewhat unusual way that involves the dynamic credential issue being seperate from Terrakube. I created an RSA key pair, and generate the public JWK part. I noticed when using the key pair with Terrakube the modulus was different. 

I tried various attempt to generate the modulus and always came up with the same result that the external tool was generating.

Upon further investigation, it looks like the issue is that Terrakube's will sometimes (50/50, leading byte if 0x80-0xFF) cause a padding byte to be added to the modulus to make the bigint positive, resulting in different encoding of the bytes. Using the OIDC token Terrakube generated in a JWK decode, the signing check fails with a modulus with padding whereas the  JWKS generated by another tool passes.

I have a fix for this has been tested locally which I'm happy raise as a PR.

### Steps to reproduce

50/50 chance of occuring depending on the public key generated for dynamic credentials

### Expected behavior

The modulus (and exponent, though that is usually the same string) is never padded with a prefix 0x00 and gives the same result as other implementations, conforming to RFC 7518

### Example repository

```
-----BEGIN PUBLIC KEY-----
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAvrtYBHbih+hGfs8GJRh9
YA7laLkl3/hUzZL8N5bCvzJlCi/tTLUOo2DurSghyRtN28Q0PvVDlXSs6kYe5HOy
n+BnjJuIj0f+FTn9lZnfX70B6W82SQRQMT6nxsCronOAcklOP578Et4ygLmQG/c+
UPmo7+gGjjx46TQ+Ga1AJM7/I6pp8H10k8JzaIoDG7tz2oPsH1DNr3oyxFvFj3nP
YjczSL34f7ZfGCifsBh8oShjJ4CtniXxxqrYuGPecsdWBlemSIPKGlmtI4mhraGe
+caWksFfuONkOWHY55kr7y1OAUpDLVn0aU2W/WOcwvFWIw6bcSkJt5DoQJViGbr+
zQIDAQAB
-----END PUBLIC KEY-----
```

Without fix:
```
2026-02-08T09:59:36.464Z  INFO 1 --- [nio-8080-exec-9] i.t.a.p.token.dynamic.JwksController     : RSA Exponent: AQAB
[...]

PR #2936, opened 2026-02-08
fix: strip leading sign byte from modulus, exponent for JWK comliance

mattwillsher-rvu:
Keys will sometimes (50/50), have a leading byte 0x80-0xFF causing a padding byte to be added to the modulus to make the bigint positive, resulting in different encoding of the bytes from RFC 7518. Using the OIDC token Terrakube generated in a JWK decode, the signing check fails with a modulus with padding whereas the  JWKS generated by another tool passes.

This code removes the first byte if zero resulting in a modulus that matches other implementations.

See #2932 for details of the validation done.


alfespa17 (approved):

alfespa17:
Thank you for your help @mattwillsher-rvu

mattwillsher-rvu:
Thank you for Terrakube!

PR #2954, opened 2026-02-10
fix: executor stream logs

themaroqa:
Removed exclusions for spring-boot-starter-logging and slf4j-simple dependency.

I noted the executor logs were always written to the stderr. 
This seems to be because slf4j-simple by default writes all logs to stderr.

Previously Executor implementation:
- Explicitly excludes spring-boot-starter-logging from spring-boot-starter
- Then manually adds slf4j-simple (which defaults to stderr)

What I did is correctly working on my environment since a couple of days. 
I was in doubt of opening this PR because I don't know if there were a reason for the explicit exclusion of spring-boot-starter-logging. 

Let me know your thoughts.

### DIFF
--- a/executor/pom.xml
+++ b/executor/pom.xml
@@ -46,12 +46,6 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter</artifactId>
-			<exclusions>
-				<exclusion>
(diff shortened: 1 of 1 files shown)


alfespa17:
To be honest, I have no idea......, I changed that 4 years ago when I was initially migrating the executor code to this repo

imagen

PR #2957, opened 2026-02-13
fix: handle self-hosted Git SSH URLs in formatSshUrl

### LINKED ISSUE #2925: formatSshUrl function doesn't handle self-hosted Git instances
### Bug description 🐞

We are currently testing terrakube, and we use a self-hosted gitlab instance. I have configured workspaces using git ssh as the source without using the gitlab integration.

The formatSshUrl function in the Terrakube UI fails to properly convert SSH URLs to HTTPS for self-hosted GitLab/GitHub/Bitbucket instances. It only handles hardcoded domains (github.com, gitlab.com, bitbucket.org), leaving the colon in SSH URLs unconverted for any other domain.

  Version Affected

  Terrakube 2.28.0 (UI component)

  Root Cause

  In the minified UI bundle, there are two similar functions:

  fixSshURL (correct implementation):
  i.replace(":", "/").replace("git@", "https://")
  This correctly handles any SSH URL by:
  1. First replacing : with /
  2. Then replacing git@ with https://

  formatSshUrl (buggy implementation):
  // Only handles specific domains:
  i.includes("github.com") && (i = i.replace(".com:", ".com/"))
  i.includes("gitlab.com") && (i = i.replace(".com:", ".com/"))
  i.includes("bitbucket.org") && (i = i.replace(".org:", ".org/"))
  i = i.replace(".git", "")
  i = i.replace("git@", "https://")

  The problem is that formatSshUrl has no fallback for domains not in the hardcoded list. Self-hosted instances like gitlab.example.com or github.mycompany.com are not handled.

  Workaround

  Users can apply a local patch by:
  1. Extracting the JS bundle from the container
  2. Adding support for their specific domain
  3. Mounting the patched file back into the container

  Example patch for gitlab.example.com:
  // Add after the bitbucket.org check:
  i.includes("gitlab.example.com") && (i = i.replace(".com:", ".com/"))

  Additional Context

  - The fixSshURL function elsewhere in the codebase handles this correctly with a generic approach
  - This affects any organization using self-hosted Git servers (GitLab CE/EE, GitHub Enterprise, Bitbucket Server)
  - The bug only affects URL display/linking in the UI; workspace execution may still work if the executor handles SSH URLs differently

### Steps to reproduce

  1. Create a workspace with an SSH source URL pointing to a self-hosted GitLab instance:
  git@gitlab.example.com:myorg/myrepo.git
  2. Navigate to the workspace in the Terrakube UI
  3. Observe that the repository link is broken/invalid

  Actual Behavior

  The URL is partially converted, leaving an invalid URL with a colon:
  https://gitlab.example.com:myorg/myrepo

  This happens because the colon between the domain and path is not replaced with a slash.

### Expected behavior

  The SSH URL should be converted to a valid HTTPS URL:
  https://gitlab.example.com/myorg/myrepo

### Example repository

_No response_

### Anything else?

_No response_

jcanizalez:
## Summary

- Replaces hardcoded domain-specific colon replacements (`.com:`, `.org:`) in `formatSshUrl` with a generic regex `^git@([^:]+):(.*)$` → `https://$1/$2`
- Fixes broken VCS links for self-hosted GitLab, GitHub Enterprise, Gitea, and any other Git instance
- Azure DevOps kept as special case due to its unique `:v3` path format

## Problem

The previous implementation only handled known domains (`github.com`, `gitlab.com`, `bitbucket.org`) by replacing domain-specific colons. Self-hosted instances like `git@gitlab.example.com:org/repo.git` were left with the SSH colon format, producing invalid URLs.

## Solution

Use a single generic regex that handles ANY SSH `git@host:path` URL, regardless of the hostname. This is the same approach already used by `fixSshURL()` in `Details.tsx`.

## Test plan

- [x] All 7 existing tests pass (GitHub, GitLab, Bitbucket, Azure DevOps)
- [x] Added 6 new tests for self-hosted instances:
  - Self-hosted GitLab SSH
  - Self-hosted GitHub Enterprise SSH
  - Self-hosted Gitea SSH
  - Self-hosted SSH with nested paths (group/subgroup/repo)
  - Self-hosted HTTPS with credentials
  - Plain HTTPS passthrough

Closes #2925

### DIFF
--- a/ui/src/modules/workspaces/utils/formatSshUrl.spec.ts
+++ b/ui/src/modules/workspaces/utils/formatSshUrl.spec.ts
@@ -35,4 +35,35 @@ describe("formatSshUrl", () => {
     const result = formatSshUrl("git@gitlab.com:org-name/project-name/repo-name.git");
     expect(result).toBe("https://gitlab.com/org-name/project-name/repo-name");
   });
+
+  // Self-hosted instances (issue #2925)
+  test("Self-hosted GitLab SSH url is formatted correctly", () => {
+    const result = formatSshUrl("git@gitlab.example.com:org-name/repo-name.git");
+    expect(result).toBe("https://gitlab.example.com/org-name/repo-name");
+  });
+
+  test("Self-hosted GitHub Enterprise SSH url is formatted correctly", () => {
+    const result = formatSshUrl("git@github.mycompany.com:org-name/repo-name.git");
+    expect(result).toBe("https://github.mycompany.com/org-name/repo-name");
+  });
+
+  test("Self-hosted Gitea SSH url is formatted correctly", () => {
+    const result = formatSshUrl("git@gitea.internal.dev:team/project.git");
+    expect(result).toBe("https://gitea.internal.dev/team/project");
+  });
+
+  test("Self-hosted SSH url with nested path is formatted correctly", () => {
+    const result = formatSshUrl("git@git.company.io:group/subgroup/repo-name.git");
+    expect(result).toBe("https://git.company.io/group/subgroup/repo-name");
+  });
+
+  test("Self-hosted HTTPS url with credentials is formatted correctly", () => {
+    const result = formatSshUrl("https://user@gitlab.example.com/org-name/repo-name.git");
+    expect(result).toBe("https://gitlab.example.com/org-name/repo-name");
+  });
+
+  test("Plain HTTPS url is returned unchanged", () => {
+    const result = formatSshUrl("https://github.com/org-name/repo-name");
+    expect(result).toBe("https://github.com/org-name/repo-name");
+  });
 });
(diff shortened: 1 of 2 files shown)


gangadhar-res:
Hi @jcanizalez, I was about to pick up #2925 and found this draft, the generic regex approach looks right. Do you plan to land it? If it helps I'm happy to finish it up: rebase it, add handling for the ssh:// form with custom ports (ssh://git@host:2222/group/repo.git) that self-hosted GitLab and Gitea commonly use, and fix the same hardcoded-domain assumption in getVcsNameFromUrl and getVcsTypeFromUrl so self-hosted instances detect correctly end to end. Or if you'd rather keep this PR as is, I can send the ssh:// and VCS-detection pieces as a separate PR on top.

alfespa17:
Fixed in #3169

PR #3017, opened 2026-03-02
fix(ui): Disable splitting of build artifacts for antd to avoid runtime errors

MatthewJohn:


alfespa17:
Thank you @MatthewJohn

PR #3038, opened 2026-03-09
feat(webhooks): shared repo-level VCS webhooks for GitHub

denniswebb:
## Summary

Terrakube creates **one GitHub webhook per workspace** (1:1 relationship). When a GitHub repository hosts 30-40 Terraform workspaces, it hits [GitHub's 20-webhook-per-repo limit](https://docs.github.com/en/rest/webhooks/repos). This PR introduces **shared repository-level webhooks** where a single GitHub hook per repository fans out to all workspaces using that repo.

## Problem

- Each workspace that enables webhooks creates its own GitHub webhook on the repository
- GitHub enforces a limit of 20 webhooks per repository
- Organizations with many Terraform workspaces backed by the same repo cannot enable webhooks for all of them
- This is a GitHub-specific limitation (GitLab/Bitbucket don't have this constraint)

### New `vcs_webhook` table

One row per unique `(vcs_id, repository_url)` pair — one shared GitHub webhook per repo per VCS connection. A single VCS connection can have many repos, each with its own VcsWebhook.

| Column | Purpose |
|--------|---------|
| `id` | PK (UUID) |
| `vcs_id` | FK to VCS connection |
| `repository_url` | Normalized repo URL (lowercase, no `.git` suffix) |
| `remote_hook_id` | GitHub's webhook ID |
| `secret` | Random UUID used as HMAC secret for payload verification |

### Two coexisting callback paths

| Path | Mode | Description |
|------|------|-------------|
| `POST /webhook/v1/{webhookId}` | Legacy | Existing per-workspace webhooks, **completely unchanged** |
| `POST /webhook/v1/vcs/{vcsWebhookId}` | Shared | New fan-out endpoint for shared webhooks |

The **URL path alone** determines which mode handles the request. There is no ambiguity or runtime detection needed.

### How fan-out works

1. GitHub sends a push/PR/release event to `/webhook/v1/vcs/{vcsWebhookId}`
2. Load the `VcsWebhook` record, verify HMAC signature using its stored secret
3. Parse the payload once (branch, changed files, event type, commit)
4. Query all workspaces where `vcs = this VCS AND normalized(source) = this repo AND webhook IS NOT NULL`
5. For each workspace, check its `WebhookEvent` rules (branch regex, path regex, event type, priority)
6. Create a Job for each matching workspace (exceptions are caught per-workspace so one failure doesn't block others)
7. Workspaces with legacy per-workspace hooks (those with `remoteHookId` set) are skipped during fan-out — they receive events via their own `/webhook/v1/{webhookId}` path

### Key design decisions

**No FK from `webhook` to `vcs_webhook`:** The fan-out query uses the workspace's existing `vcs` and `source` fields to find matching workspaces. This means **zero changes to the existing `webhook` table schema**.

**Random secret per VcsWebhook:** Each shared webhook gets a `UUID.randomUUID()` as its HMAC secret, stored in the `secret` column. This is different from the legacy approach which uses `Base64(workspaceId)` as the secret. A random secret is more secure — if one repo's secret leaks, it doesn't compromise others. The legacy approach continues to work unchanged for existing webhooks.

**URL normalization:** Repository URLs are normalized (lowercase, strip `.git`, strip trailing `/`) for matching. This is done via a static `WebhookServiceBase.normalizeRepositoryUrl()` method used consistently across creation, deletion, and fan-out.

**GitHub only for now:** Only GitHub gets shared webhook support. GitLab and Bitbucket continue with per-workspace webhooks (GitLab doesn't have this limit; Bitbucket can be added later).

**Event type union:** When multiple workspaces share a webhook, the GitHub hook is configured with the **union** of all event types across those workspaces. For example, if workspace A subscribes to PUSH and workspace B subscribes to PUSH + PULL_REQUEST, the GitHub hook gets both event types. When a workspace is removed, the union is recalculated and the GitHub hook is PATCHed.

**Migration endpoint requires auth:** The new `/api/v1/webhook/{webhookId}/migrate` endpoint is placed under `/api/v1/` (not `/webhook/v1/`) so it goes through the standard authentication filter. The `/webhook/v1/**` paths are `permitAll()` since they receive unauthenticated payloads from VCS providers.

### New files

| File | Description |
|------|-------------|
| `VcsWebhook.java` | JPA entity for `vcs_webhook` table |
| `VcsWebhookRepository.java` | Spring Data repository with `findByVcsAndRepositoryUrl` |
| `changelog-2.31.0-vcs-webhook.xml` | Liquibase migration (create table + unique constraint + FK) |

### Modified files

| File | Description |
|------|-------------|
| `WebhookService.java` | Core shared logic: creation, deletion cleanup, fan-out processing, migration |
| `WebhookServiceBase.java` | Added `normalizeRepositoryUrl()` and `handleWebhookWithSecret()` |
| `GitHubWebhookService.java` | Added `processSharedWebhook`, `createSharedWebhook`, `updateSharedWebhookEvents`, `deleteSharedWebhook` |
| `WebHookController.java` | Added shared inbound endpoint and authenticated migration endpoint |
| `WorkspaceRepository.java` | Added `findByVcsAndWebhookIsNotNull(Vcs)` query |
| `changelog.xml` | Registered new migration |
| `Webhook.tsx` | Alert banners for shared/legacy mode, "Migrate to shared" button |

### Unchanged

- `Webhook.java` — no entity changes
- `WebhookEvent.java` — no entity changes
- `webhook` table schema — no column additions or modifications
- Legacy `/webhook/v1/{webhookId}` processing — completely untouched

## Backward compatibility

- All existing per-workspace webhooks continue to work identically
- The legacy `processWebhook()` code path is not modified
- No database migration touches existing tables
- New webhooks on GitHub VCS connections automatically use shared mode
- A "Migrate to shared" button in the UI lets users opt-in for existing legacy webhooks

## Test plan

- [ ] Create webhook for workspace A on repo X → creates VcsWebhook + GitHub hook at `/v1/vcs/`
- [ ] Create webhook for workspace B on same repo → reuses VcsWebhook, no new GitHub API call
- [ ] Push to repo → `/webhook/v1/vcs/{id}` receives event, jobs created for both A and B based on their WebhookEvent rules
- [ ] Delete workspace A webhook → VcsWebhook persists, GitHub hook stays
- [ ] Delete workspace B webhook (last one) → VcsWebhook and GitHub hook both deleted
- [ ] Event union: workspace A has PUSH, workspace B has PUSH + PR → GitHub hook configured with both. Remove B → PATCHed to PUSH only
- [ ] Legacy webhooks: existing `/webhook/v1/{webhookId}` flow unchanged
- [ ] Migration: click "Migrate to shared" → old GitHub hook deleted, workspace uses shared VcsWebhook, WebhookEvent rules preserved
- [ ] Signature verification: shared endpoint uses VcsWebhook.secret, legacy uses Base64(workspace.id)

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/vcs/WebhookService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/vcs/WebhookService.java
@@ -3,10 +3,14 @@ package io.terrakube.api.plugin.vcs;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 import io.terrakube.api.plugin.scheduler.ScheduleJobService;
@@ -14,11 +18,14 @@ import io.terrakube.api.plugin.vcs.provider.bitbucket.BitBucketWebhookService;
 import io.terrakube.api.plugin.vcs.provider.github.GitHubWebhookService;
 import io.terrakube.api.plugin.vcs.provider.gitlab.GitLabWebhookService;
 import io.terrakube.api.repository.JobRepository;
+import io.terrakube.api.repository.VcsWebhookRepository;
 import io.terrakube.api.repository.WebhookEventRepository;
 import io.terrakube.api.repository.WebhookRepository;
+import io.terrakube.api.repository.WorkspaceRepository;
 import io.terrakube.api.rs.job.Job;
 import io.terrakube.api.rs.job.JobStatus;
 import io.terrakube.api.rs.vcs.Vcs;
+import io.terrakube.api.rs.webhook.VcsWebhook;
 import io.terrakube.api.rs.webhook.Webhook;
 import io.terrakube.api.rs.webhook.WebhookEvent;
 import io.terrakube.api.rs.webhook.WebhookEventType;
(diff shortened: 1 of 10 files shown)


alfespa17:
Hello @denniswebb 

With  `VcsWebhook` you will have a similar issue, imagine you have 20 organizations, you will be creating a webhook for each vcs, maybe one for each organization.

It will be a better idea to create a `new entity` that is not related to any `organization` or `vcs`, something like "RepoWebhook", that you can reuse for any workspace and organization.

Something like this could work and [Elide](https://elide.io/) will expose the rest API

```java
@Getter
@Setter
@Entity(name = "repo_webhook")
@Include(rootLevel = true)
public class RepoWebhook extends GenericAuditFields
```

You can create a new endpoint `v2` so previous webhooks are not impacted

`/webhook/v2/vcs/{vcsWebhookId}`


You could simply add a new property to the Webhook class to identify which are migrated successfully

```java
public class Webhook extends GenericAuditFields {

    @Id
    @JdbcTypeCode(Types.VARCHAR)
    @Convert(converter = IdConverter.class)
    private UUID id;

    @Column(name = "migrated_v2")
    private boolean migrated;
   .....
}
```

Each `webhook is created using the Id` from the above entity, for example [here](https://github.com/terrakube-io/terrakube/blob/67c8592a90e341f1ba78d26a90ea38706f968c4e/api/src/main/java/io/terrakube/api/plugin/vcs/provider/github/GitHubWebhookService.java#L286), so when you update the `migrated` property you could simply delete previous webhooks from the repo

Once you have disable the old webhooks you could simply add a new row to `RepoWebhook` (Only if the repo is not already there) with the repo information.

Finally when you received an event in the new webhook enpoint, you simply search in all workspaces where the migrated=true to trigger a job there using the logic that we already have in the webhooks service.

alfespa17:
Something like the idea in this commit https://github.com/terrakube-io/terrakube/commit/d8eadf595074ec6ab7b49ae75d963b6358a3435e (I didn't finish it, it is just a high level idea but you could check the file changes)

alfespa17:
I am closing this one, I saw the other PR

PR #3115, opened 2026-03-31
fix(executor): decouple S3 region from endpoint; add AwsForcePathStyle flag

SolomonHD:
## Problem

When `AwsEndpoint` is set to any non-empty value (including a real AWS regional endpoint like `https://s3.us-east-1.amazonaws.com`), `TerraformStateAutoConfiguration` builds the S3 client with a hard-coded `Region.of("auto")`:

```java
s3client = S3Client.builder()
        .region(Region.of("auto"))   // ← ignores AwsTerraformStateRegion entirely
        .endpointOverride(URI.create(awsTerraformStateProperties.getEndpoint()))
        ...
```

`Region.of("auto")` resolves to `aws-global` in the SigV4 signer. AWS S3 then rejects the request:

```
S3Exception: The authorization header is malformed; the region 'aws-global' is wrong; expecting 'us-east-1'
```

This means any user pointing `AwsEndpoint` at a real AWS S3 regional URL is silently broken — the configured `AwsTerraformStateRegion` is ignored.

The custom-endpoint code path also uncondtionally enables `pathStyleAccessEnabled(true)`, which is correct for MinIO/Wasabi/B2 but is deprecated and unsupported for new AWS S3 buckets (created after 2020).

## Root cause

The original design coupled two independent concerns into a single check:
1. **Custom endpoint** (needed for S3-compatible services)
2. **Path-style access** (required by S3-compatible services, forbidden for new AWS buckets)

## Fix

- **Always use `Region.of(awsTerraformStateProperties.getRegion())`** — the configured region is now used regardless of whether a custom endpoint is set.
- **Add `AwsForcePathStyle` boolean flag** (env var: `AwsForcePathStyle`, default `false`) — users explicitly opt in to path-style access when using S3-compatible storage.
- `AwsEndpoint` continues to work as before for setting a custom endpoint override; it no longer silently overrides the region.

## Changes

| File | Change |
|---|---|
| `AwsTerraformStateProperties.java` | Added `forcePathStyle` field |
| `TerraformStateAutoConfiguration.java` | Replaced 3-branch if/else with unified builder; region always from config; path-style conditional on `forcePathStyle` |
| `application.properties` | Added `AwsForcePathStyle:false` default; added `:` default to `AwsEndpoint` to prevent startup failure when unset |
| `docker-compose/docker-compose.yml` | Added `AwsForcePathStyle` to all three service anchors with explanatory comments |
| `docker-compose/.env` | Added `TK_OUTPUT_FORCE_PATH_STYLE=true` (correct default for bundled MinIO) with comments |
| `docker-compose/README.md` | Added Storage Backend Configuration section with variable table and examples for AWS S3 and MinIO |

## Backwards compatibility

- No breaking changes. `AwsForcePathStyle` defaults to `false`.
- Existing MinIO deployments that set `AwsEndpoint` must add `AwsForcePathStyle=true` to restore path-style access (previously it was implicit). The updated `.env` default handles this for the bundled compose setup.
- AWS S3 users with `AwsEndpoint` set to an AWS regional URL: remove or empty `AwsEndpoint` — the SDK now constructs the correct endpoint from the region automatically.

## Testing

- Real AWS S3: `AwsEndpoint=""`, `AwsTerraformStateRegion=us-east-1`, `AwsForcePathStyle=false` → SigV4 signing with `us-east-1` ✓
- MinIO: `AwsEndpoint=http://minio:9000`, `AwsTerraformStateRegion=us-east-1`, `AwsForcePathStyle=true` → path-style with custom endpoint ✓

### DIFF
--- a/docker-compose/.env
+++ b/docker-compose/.env
@@ -35,11 +35,17 @@ TK_LDAP_VERSION=2.6.9-debian-12-r10
 #######################
 TK_MINIO_VERSION=2025
 TK_OUTPUT_ACCESS_KEY=minioadmin
+# TK_OUTPUT_ENDPOINT: Full URL for S3-compatible storage (MinIO, Wasabi, Backblaze B2, etc.)
+# Leave empty for real AWS S3 — the SDK constructs the endpoint from TK_OUTPUT_STORAGE_REGION.
 TK_OUTPUT_ENDPOINT=http://terrakube-minio:9000
 TK_OUTPUT_SECRET_KEY=minioadmin
 TK_OUTPUT_STORAGE_REGION=us-east-1
 TK_OUTPUT_BUCKET_NAME=sample
 TK_OUTPUT_BUCKET_REGION=us-east-1
+# TK_OUTPUT_FORCE_PATH_STYLE: Set to true for S3-compatible storage (MinIO, Wasabi, B2, etc.)
+# that requires path-style access (https://endpoint/bucket/key).
+# Set to false for real AWS S3 (AWS deprecated path-style for new buckets).
+TK_OUTPUT_FORCE_PATH_STYLE=true
 
 ###################
 #DATABASE SETTINGS#
--- a/docker-compose/README.md
+++ b/docker-compose/README.md
@@ -58,4 +58,45 @@ Terrakube will be available in the following URL:
 
 * https://terrakube.platform.local
   * Username: admin@example.com
-  * Password: admin 
+  * Password: admin
+
+## Storage Backend Configuration
+
+Terrakube supports both real AWS S3 and S3-compatible storage (MinIO, Wasabi, Backblaze B2, Cloudflare R2, etc.).
+The default `.env` ships configured for the bundled MinIO container.
+
+### Key `.env` variables
+
+| Variable | Description | AWS S3 | MinIO / S3-compatible |
+|---|---|---|---|
+| `TK_OUTPUT_ENDPOINT` | Full endpoint URL | `""` (leave empty) | `http://terrakube-minio:9000` |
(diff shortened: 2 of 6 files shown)


alfespa17:
Hello @SolomonHD it has been a while since you developed the original version for the docker compose for this project a few years ago.

Your work was the inspiration to build the devcontainer support later.

A few days ago, I did a small change in the devcontainer that I never migrated to the docker compose, using the devcontainer you no longer need to create the following:

`docker network create terrakube-network -d bridge --subnet 10.25.25.0/24 --gateway 10.25.25.254`

Now you can simply expose the service using localhost:80 and localhost:443

You could check the details here if you are interested:

https://github.com/terrakube-io/terrakube/pull/3095

The solution was basically use network alises in the traefik container

```
### Containers
services:
  traefik:
    image: traefik:v2
    container_name: terrakube-traefik
    networks:
      terrakube-network:
        aliases:
        - terrakube-api.platform.local
        - terrakube-executor.platform.local
        - terrakube.platform.local
        - terrakube-registry.platform.local
        - terrakube-dex.platform.local
    environment: *traefik_env
    dns: *terrakube_dns
    ports:
      - $TRAEFIK_HTTP_PORT:80
      - $TRAEFIK_HTTPS_PORT:443
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - ./traefik_dynamic.yml:/etc/traefik_dynamic.yml
      - ./key.pem:/letsencrypt/privkey.pem:ro
      - ./cert.pem:/letsencrypt/fullchain.pem:ro
    restart: unless-stopped
```

alfespa17:
@SolomonHD there are some build issues, can you help me fixing those so I can merge your PR?

SolomonHD:
I'll look into it and glad to be back! My org has decided to give TK another look :D

alfespa17:
this was fixed in https://github.com/terrakube-io/terrakube/pull/3473

PR #3169, opened 2026-05-05
fix(ui): accept hyphenated SSH host and stop crashing on unknown SSH provider

### LINKED ISSUE #3168: UI bug: SSH Git URL with hyphen in domain name fails regex validation and causes blank organization UI
### Bug description 🐞

SSH git URLs containing a hyphen (`-`) in the domain name are rejected by the
workspace creation form's regex validation. Additionally, if such a URL is
manually inserted into the database, the entire organization UI breaks with a
blank page error.

Two distinct bugs are involved:
1. **Frontend validation regex** rejects valid SSH URLs with hyphens in the domain
2. **UI crash** when the URL is present in the database, caused by misinterpreting
   the SSH URL as an HTTP URL

I have the same problem on these versions:
- 2.30.2
- 2.31.0

Only the UI is broken, the executor works as expected.

I have created only 1 ticket as they are strongly related, but maybe I should have created 2.

### Bug 1 — Validation error on workspace creation

1. Go to an Organization → **New Workspace** → select **VCS (SSH)**
2. Enter an SSH URL where the domain contains a hyphen, e.g.: `git@github-test.com:terrakube-io/terrakube-docker-compose.git`

**Expected:** The URL is accepted (it is a valid SSH git URL).  
**Actual:** A validation error is displayed:
```
Git repo does not match the pattern /(empty)|(((git|ssh|http(s)?)|(git@[\w\.]+))(:(\/\/)?)([\w\.@\:/\-~]+)(\.git)?(\/)?)/
Git repo must be a valid git url using either https or ssh protocol.
```

You can continue but you cannot create the workspace.

---

### Bug 2 — Organization UI crash when a hyphenated SSH URL is in the database

1. Manually update a workspace's git URL in the database to use a hyphenated
   domain, e.g.: `git@github-test.com:terrakube-io/terrakube-docker-compose.git`
2. Navigate to the Organization page in the UI

**Expected:** The organization page loads normally.  
**Actual:** The page is blank and the browser console shows:
```
Unexpected Application Error!
URL constructor: https://github-test.com:terrakube-io/terrakube-docker-compose
is not a valid URL.
```

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

mvanhorn:
Closes #3168.

The repo URL form on the workspace and module create flows rejected valid SSH URLs like `git@github-test.com:org/repo.git` because the host part of the validation regex (`git@[\w\.]+`) excluded the hyphen. Once such a URL made it into the database, the organization page also crashed with `URL constructor: https://github-test.com:terrakube-io/... is not a valid URL` - `formatSshUrl` rewrote `git@` to `https://` but only replaced the host:path colon with a slash for the four hardcoded providers (`github.com`, `ghe.com`, `dev.azure.com`, `bitbucket.org`, `gitlab.com`). Self-hosted and hyphenated hosts left the colon intact, and the rest of the path was parsed as a port number.

## What changed

**1. Validation regexes (`ui/src/domain/Workspaces/Create.tsx`, `ui/src/domain/Modules/Create.tsx`)**

`git@[\w\.]+` -> `git@[\w\.\-]+` so the host can contain a hyphen.

**2. `ui/src/modules/workspaces/utils/formatSshUrl.ts`**

`source.replace("git@", "https://")` -> `source.replace(/^git@([^:]+):/, "https://$1/")`. This converts `git@host:path` to `https://host/path` in a single step so unknown hosts work the same as the four hardcoded ones (which previously each had their own host-specific `.com:`/`.org:` -> `/` replacement).

The DevOps branch already stripped `:v3`. The SSH form (`git@ssh.dev.azure.com:v3/...`) now lands as `.../v3/...` after the host rewrite, so I added a parallel `.replace("/v3", "")`. All existing test cases (DevOps SSH/HTTPS, GitHub SSH, BitBucket SSH/HTTPS, GitLab SSH/HTTPS) still produce the same expected output.

**3. New regression test** in `formatSshUrl.spec.ts` for the exact URL from the issue (`git@github-test.com:terrakube-io/terrakube-docker-compose.git` -> `https://github-test.com/terrakube-io/terrakube-docker-compose`).

## Test plan

- [x] All 7 existing `formatSshUrl` cases keep their original expected outputs.
- [x] New regression test for hyphenated host passes.
- [x] Both validation regexes match `git@github-test.com:org/repo.git` (and still match the existing GitHub/GitLab/BitBucket/DevOps shapes).

The unit tests run via `yarn test` (jest + ts-jest) cannot be executed cleanly in a fresh clone here because the project's `tsconfig.test.json` uses options (`baseUrl`, `moduleResolution=node10`) that the bundled TypeScript 6.0.3 flags as deprecated and treats as errors before any test code runs. That is independent of this change. I verified the new behavior by porting `formatSshUrl` to a standalone Node script and confirming all 7 existing assertions plus the new hyphenated-host case match the expected outputs - same logic, same TypeScript file, just the test runner in this clone is blocked on a pre-existing tsconfig compatibility gap.

### DIFF
--- a/ui/src/domain/Modules/Create.tsx
+++ b/ui/src/domain/Modules/Create.tsx
@@ -391,7 +391,7 @@ export const CreateModule = () => {
                   {
                     required: true,
                     pattern: new RegExp(
-                      "((git|ssh|http(s)?)|(git@[\\w\\.]+))(:(//)?)([\\w\\.@\\:/\\-~]+)(\\.git)?(/)?"
+                      "((git|ssh|http(s)?)|(git@[\\w\\.\\-]+))(:(//)?)([\\w\\.@\\:/\\-~]+)(\\.git)?(/)?"
                     ),
                   },
                 ]}
--- a/ui/src/domain/Workspaces/Create.tsx
+++ b/ui/src/domain/Workspaces/Create.tsx
@@ -643,7 +643,7 @@ export const CreateWorkspace = () => {
                   {
                     required: true,
                     pattern: new RegExp(
-                      "(empty)|(((git|ssh|http(s)?)|(git@[\\w\\.]+))(:(//)?)([\\w\\.@\\:/\\-~]+)(\\.git)?(/)?)"
+                      "(empty)|(((git|ssh|http(s)?)|(git@[\\w\\.\\-]+))(:(//)?)([\\w\\.@\\:/\\-~]+)(\\.git)?(/)?)"
                     ),
                   },
                 ]}
--- a/ui/src/modules/workspaces/utils/formatSshUrl.spec.ts
+++ b/ui/src/modules/workspaces/utils/formatSshUrl.spec.ts
@@ -35,4 +35,15 @@ describe("formatSshUrl", () => {
     const result = formatSshUrl("git@gitlab.com:org-name/project-name/repo-name.git");
     expect(result).toBe("https://gitlab.com/org-name/project-name/repo-name");
   });
+
+  test("self-hosted SSH url with hyphenated host is formatted correctly", () => {
+    // Regression test for issue #3168: a host containing a hyphen (here a
+    // self-hosted GitHub instance) used to round-trip through the function as
+    // "https://github-test.com:terrakube-io/terrakube-docker-compose", which
+    // crashed `new URL(...)` because "terrakube-io" parses as a port. The host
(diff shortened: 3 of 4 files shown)


alfespa17 (approved):

mvanhorn:
Thanks @alfespa17 for the review and merge!

PR #3167, opened 2026-05-03
feat(ui): add project management pages

hirokimatsueda:
## Summary

Closes #650

Adds **Project management** to the UI, allowing users to group and organize Workspaces under Projects within an Organization. Users can list, create, edit, and delete Projects, and assign Workspaces to Projects for filtering.

Also fixes a bug where `terraform init` with a `workspaces { project = "..." }` block failed to locate the configured project.

---

### New pages / components

| File | Description |
|---|---|
| `ProjectsPage.tsx` | List and create Projects within an Organization (includes 403 error handling) |
| `ProjectDetailPage.tsx` | Edit Project settings (name, description), delete Projects, and manage assigned Workspaces (includes 403 error handling) |
| `ProjectWorkspaces.tsx` | List, add, and remove Workspaces assigned to a Project (includes 403 error handling) |
| `projectService.ts` | Project CRUD operations (GraphQL reads + JSON:API REST writes) |

### Modified files

| File | Change |
|---|---|
| `App.tsx` | Add routes for `/organizations/:id/projects` and `/:orgid/projects/:id` |
| `MainMenu.tsx` | Add "Projects" menu item with `ProjectOutlined` icon to the sidebar |
| `Create.tsx` | Add optional Project selector when creating a Workspace |
| `Details.tsx` | Show the assigned Project name (with link) in the Workspace detail sidebar |
| `Workspaces/Settings/General.tsx` | Add "Project" section — a dropdown to assign/unassign a Project |
| `WorkspaceCard.tsx` | Display the assigned Project name as a badge on each Workspace card |
| `WorkspaceFilter.tsx` | Add Project filter (includes an "Unassigned" option); persist selection to sessionStorage |
| `OrganizationDetailsPage.tsx` | Derive unique Project list from Workspaces and pass it to `WorkspaceFilter` |
| `workspaceService.ts` | Add `assignWorkspaceToProject` / `removeWorkspaceFromProject` (`PATCH /workspace/:id/relationships/project`); include `project` in GraphQL query |
| `types.ts` (workspaces) | Add `projectId` / `projectName` to `WorkspaceListItem` |
| `types.ts` (domain) | Define `ProjectModel` type |
| `PageWrapper.tsx` | Ant Design v5 compatibility: replace Breadcrumb `label` with `title` |
| `RemoteTfeService.java` | Fix: JSON:API type corrected from `"project"` → `"projects"` in project listing response (see below) |

---

## Error handling

[...]

PR #3178, opened 2026-05-09
feat: add project-level team permissions

hirokimatsueda:
Related #3167

## Overview
This PR introduces project-level permissions, allowing teams to be granted roles at the project level that apply to all workspaces within that project.
This addresses the need to simplify permission management for large organizations with many workspaces grouped under projects.

The API endpoints and the UI are both implemented and have been verified to work as expected.

### What's Included
- `ProjectAccess` entity: Stores team permissions at the project level
- Security checks: New permission rules that propagate project-level permissions
  to workspaces and jobs within the project
- API endpoints: Support for creating, reading, and managing project access rules
- UI: A "Team Access" tab on the project detail page for managing team assignments

### How It Works
1. Grant a team a role on a project (e.g., "Write")
2. All workspaces within that project inherit the team's permissions
3. Organization-level permissions take precedence if they are more permissive

### Permission Hierarchy
Organization-level (most permissive wins)
↓
Project-level
↓
Workspace-level

### Supported Project-level Roles

| Role  | Manage Project | Manage Workspace | Create Workspace | Plan Job | Approve Job |
|-------|:---:|:---:|:---:|:---:|:---:|
| Admin | ✅ | ✅ | ✅ | ✅ | ✅ |
| Write | ❌ | ✅ | ✅ | ✅ | ✅ |
| Plan  | ❌ | ❌ | ❌ | ✅ | ❌ |
| Read  | ❌ | ❌ | ❌ | ❌ | ❌ |

"Manage Project" covers editing project settings and managing team access entries for the project.

Note: Reassigning a workspace to a different project requires organization-level Manage Workspace permission,
regardless of project-level roles, to prevent unauthorized workspace moves across projects.

## Screenshots

image

image

image

image

### Q1: Should Project-level permissions include "Manage State"?

Currently, Manage State is only enforced at the organization level — there is no project-scoped security check for state operations.
The permission logic (`canManageState`) is already implemented for `ProjectAccess` in `RbacV2Service`
(Admin and Write both return true, consistent with org-level behavior),
but the corresponding security check class has not been added yet.

**Should we:**
1. Keep State management at Organization level only (current behavior)
2. Add a project-scoped security check to enable Manage State for Admin and Write
   (consistent with organization-level behavior)
3. Something else?

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/security/groups/GroupService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/security/groups/GroupService.java
@@ -15,4 +15,6 @@ public interface GroupService {
 
      boolean isMemberWithLimitedAccessV2(User user, Organization organization);
 
+     boolean isMemberWithProjectAccess(User user, Organization organization);
+
 }
--- a/api/src/main/java/io/terrakube/api/plugin/security/groups/dex/DexGroupServiceImpl.java
+++ b/api/src/main/java/io/terrakube/api/plugin/security/groups/dex/DexGroupServiceImpl.java
@@ -12,7 +12,9 @@ import org.springframework.security.oauth2.server.resource.authentication.JwtAut
 import org.springframework.stereotype.Service;
 import io.terrakube.api.plugin.security.groups.GroupService;
 import io.terrakube.api.repository.AccessRepository;
+import io.terrakube.api.repository.ProjectAccessRepository;
 import io.terrakube.api.rs.Organization;
+import io.terrakube.api.rs.project.access.ProjectAccess;
 import io.terrakube.api.rs.workspace.Workspace;
 import io.terrakube.api.rs.workspace.access.Access;
 
@@ -29,6 +31,8 @@ public class DexGroupServiceImpl implements GroupService {
     RedisTemplate redisTemplate;
 
     AccessRepository accessRepository;
+
+    ProjectAccessRepository projectAccessRepository;
 
     FederatedRepository federatedRepository;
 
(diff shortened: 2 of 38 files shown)


alfespa17:
Thank you, I will try to check during this week when I have some free time from work

alfespa17:
This type of changes are a little difficult to validate but you could add some test cases like this [example](https://github.com/terrakube-io/terrakube/blob/33fc58e76a082249d8c212e8bf89ee4655b76722/api/src/test/java/io/terrakube/api/AccessTests.java#L128)

It will be something like 
- create an organization
- create the teams
- creating a project
- grant the access 
- do a search using the get workspace API
- validate the response contains the workspace where the project grant access

 You could even add some information using liquibase 

https://github.com/terrakube-io/terrakube/blob/main/api/src/main/resources/db/changelog/demo-data/gcp.xml

Updating the demo data examples

https://github.com/terrakube-io/terrakube/blob/33fc58e76a082249d8c212e8bf89ee4655b76722/api/src/main/resources/db/changelog/changelog-demo.xml#L13

hirokimatsueda:
Thank you for the feedback! I've added integration tests as suggested — the test cases cover the following scenarios:

- A user without project access cannot see a workspace in that project
- After granting project-level access, the user can see the workspace and manage it according to their role (read, write, admin)
- A project admin can manage ProjectAccess entries but cannot manage project access with a write role
- I also added demo data via Liquibase following the pattern in gcp.xml

While working on this, I noticed a potential security concern around delegating project management to teams: if a workspace is removed from a project (i.e., project = null), the current implementation falls back to org-wide membership checks, making the workspace visible to all org members — including those who should not have access.

To address this, I'm currently exploring an approach using a Default Project per organization:

- Workspaces without an explicit project are automatically assigned to the Default Project, which is open to all org members (preserving backward compatibility)
- Workspaces in non-default projects are only accessible through explicit project-level access grants
- Moving workspaces between projects is restricted to org-level admins only

I'm still thinking through whether this is the right approach, and will continue to explore the design.

hirokimatsueda:
**Update:** After further consideration, I decided not to pursue the Default Project approach.

Here's how the concerns were resolved:

- **Workspaces with no project assigned** remain visible to all org members. This preserves backward compatibility and is kept as intentional behavior.
- **Workspaces assigned to a project** are only accessible to users who have explicit project-level access or org-level Write+ membership. The security gap I was concerned about does not apply to project-assigned workspaces.
- **Workspace-project reassignment** requires either org-level Write+ (`team manage workspace`) or project-level Admin/Write on the *destination* project. This allows project admins to pull workspaces into their project, while preventing arbitrary moves across projects without appropriate authority.

I think this is simpler than the Default Project approach and avoids the schema/migration complexity that would come with it.

I plan to do a bit more end-to-end testing before marking this ready for review.

hirokimatsueda:
Update: End-to-end testing is complete. Marking this as ready for review.

alfespa17 (approved):
Amazing job!

PR #3187, opened 2026-05-13
fix(executor,api): make dynamic credentials path consistent

pierrickorni:
The executor writes dynamic-credentials files to a random tmp dir (SetupWorkspaceImpl.setupWorkspaceDirectory creates `tmp*` under ~/.terraform-spring-boot/executor), but the API hardcoded a different deterministic path (`~/.terraform-spring-boot/executor/{orgId}/{wsId}`) both inside the credential_source.file of the GCP config JSON and in the GOOGLE_APPLICATION_CREDENTIALS / AWS_WEB_IDENTITY_TOKEN_FILE env vars. PR #3173/#3176 attempted to repair this by rewriting the env vars to the Terraform working directory + `_gcp.json` suffix in TerraformExecutorServiceImpl.loadTempEnvironmentVariables, but the writer kept using the clone-root path with no suffix, so the env var and the file on disk still never agree.

Fix:

- API (DynamicCredentialsService): emit credential_source.file with a `${WORKSPACE_DIRECTORY}` placeholder and stop pre-setting the GOOGLE_APPLICATION_CREDENTIALS / AWS_WEB_IDENTITY_TOKEN_FILE env vars (only the executor can know the real path).
- Executor (SetupWorkspaceImpl): substitute `${WORKSPACE_DIRECTORY}` with the absolute workspace clone path when writing the GCP config, and set the env vars to the absolute path of the file just written.
- Executor (TerraformExecutorServiceImpl): remove the now-redundant post-hoc env-var rewrite in loadTempEnvironmentVariables.

Also extend SetupWorkspaceTest to assert AWS_WEB_IDENTITY_TOKEN_FILE is set and add a GCP test covering placeholder substitution and GOOGLE_APPLICATION_CREDENTIALS.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/token/dynamic/DynamicCredentialsService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/token/dynamic/DynamicCredentialsService.java
@@ -166,7 +166,8 @@ public class DynamicCredentialsService {
 
         workspaceEnvVariables.put("TERRAKUBE_AWS_CREDENTIALS_FILE", awsWebIdentityToken);
         workspaceEnvVariables.put("AWS_ROLE_ARN", workspaceEnvVariables.get("WORKLOAD_IDENTITY_ROLE_AWS"));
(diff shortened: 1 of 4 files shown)


alfespa17:
I am closing this PR, I created a new https://github.com/terrakube-io/terrakube/pull/3189 with a few adjustments

PR #3193, opened 2026-05-17
feat(ui): add inline role editing and permission controls to project access tab

hirokimatsueda:
## Summary

Follow-up to #3178 (project-level team permissions).

- **Inline role editing:** Added a "Change" button to each row in the team access table, allowing admins to update a team's role without removing and re-adding them
- **Role descriptions in dropdowns:** Both the "Change" role select and the "Add Team" role select now show a short description of each role in the dropdown options, making it easier to choose the right role without prior knowledge
- **Role tooltip:** Hovering over a role tag in the table shows the same description as a tooltip

## Screenshots

image

image

image


alfespa17 (approved):
Looks good

PR #3194, opened 2026-05-17
fix for IdP auth for AWS

jplorier:
Version 1.31.1 fails to authenticate AWS provider via IdP.
This fixes the issue


alfespa17:
This should be fixed here https://github.com/terrakube-io/terrakube/releases/tag/2.31.2-beta.2

jplorier:
closing after @alfespa17 comment

PR #3199, opened 2026-05-19
feat(ui): implement Edit Client button for VCS providers

### LINKED ISSUE #1096: Settings - VCS Providers - Edit Client does nothing
### Bug description 🐞

Subj

### Steps to reproduce

1. Setup VCS connection to GitLab for example.
2. Try to edit it by going to Settings - VCS Providers - Edit Client.
3. Nothing happens

### Expected behavior

Some UI for editing VCS client is appearing

### Example repository

_No response_

### Anything else?

Also nothing in browser console, no network request, seems that event on that button is not working. Tested in both Firefox and Chrome clone.

hirokimatsueda:
## Summary

Closes #1096 

- Implement the "Edit Client" button in the organization VCS Provider settings (`/organizations/:orgid/settings/vcs`), which was rendered but had no `onClick` handler
- Add new `EditVCS` component that pre-fills editable fields (`name`, `clientId`, `endpoint`, `apiUrl`) from the existing VCS record and submits a PATCH request on save
- `clientSecret` / `privateKey` fields are intentionally left empty on load (the API never returns secrets); users can enter a new value to update, or leave blank to keep the existing one
- VCS type and connection type are inferred from stored data and shown as read-only, with a link to the provider-specific Terrakube documentation

## Screenshot

image

### DIFF
--- a/ui/src/domain/Settings/EditVCS.tsx
+++ b/ui/src/domain/Settings/EditVCS.tsx
@@ -1,0 +1,337 @@
+import { Button, Form, Input, Space, Spin, Typography, message } from "antd";
+import { useEffect, useState } from "react";
+import { HiOutlineExternalLink } from "react-icons/hi";
+import { useParams } from "react-router-dom";
+import axiosInstance, { getErrorMessage } from "../../config/axiosConfig";
+import { VcsConnectionType, VcsType, VcsTypeExtended } from "../types";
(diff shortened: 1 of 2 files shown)


alfespa17 (approved):

alfespa17:
thank you @hirokimatsueda

PR #3209, opened 2026-05-22
fix(ui): support subpath deployments and OIDC callback routing

mabunixda:
Implemented subpath support for the UI so Terrakube can run reliably under paths like `ui` without breaking routes, static assets, or provider icons. The authentication flow was updated to use runtime-aware redirect URIs and a custom sign-in callback, preventing OIDC logins from being rewritten to `/` and keeping users on the configured base path after Dex redirects. The changes remain backward-compatible with root deployments through runtime configuration (`REACT_APP_BASE_PATH` and `REACT_APP_REDIRECT_URI` ).

* make UI build/path handling compatible with subpath deployments (for example /ui/)
*  preserve SPA routing after OIDC callback so users stay on the configured base path
* ensure provider/static asset paths resolve correctly when hosted under a subpath
* keep root-path deployments supported via runtime config


alfespa17:
Thank you, I will check later when I have some free time

alfespa17 (approved):

PR #3213, opened 2026-05-25
Allow air gapped agents

jplorier:
In order to have air gapped agents on satellite clusters without needing to have direct storage bucket access, this PR implements new API endpoints and use existing ones to manage state, plan and outputs via the API directly.
It also implements a checksum validation to warranty consistency


jplorier:
will be splited in smaller PRs

Issue #1, opened 2021-03-20
aks-server API definition


alfespa17:
Create the API definition for the server

PR #3214, opened 2026-05-25
feat(api): HTTP backend for executors + sha256 integrity + workspace state locks

jplorier:
## Summary

- Adds the API surface for running executors in air-gapped environments: a new `/tfstate/v1/http-backend/**` controller that implements the Terraform `http` backend protocol, so executors no longer need direct egress to the configured object store.
- Adds an `X-Content-Sha256` header check on state/plan/output PUTs (`UploadIntegrity`). On mismatch the request is rejected with `409`. Header is optional, preserving backward compatibility with existing archive PUTs.
- Persists workspace state locks in a new `workspace_state_lock` table with `WorkspaceStateLock` entity, `WorkspaceStateLockRepository`, and `changelog-2.33.0-workspace-state-lock.xml`.
- Authentication is enforced in-controller via HTTP Basic (Terraform's `http` backend cannot send custom headers but supports user/pass) — the password is validated as an HMAC-signed internal JWT. The `/tfstate/v1/http-backend/**` path is added to `DexWebSecurityAdapter`'s CSRF-ignore + permitAll lists so the controller can own auth.
- Extends `StorageTypeService` with `uploadStepOutput`, `uploadTerraformPlan`, and `deleteCurrentTerraformState`, implemented across local, AWS, Azure, and GCP backends so the API can act as a write-through proxy for executors.
- Adds new history-state read/write endpoints on `TerraformStateController` (raw + JSON) used by the executor for upload + read-back verification.

This is the backend half of the air-gapped-executor feature. The executor side that consumes these endpoints is in a separate PR.

## Tests

- `UploadIntegrityTest` — sha256 helper, header-absent/blank/match/mismatch/uppercase/whitespace paths, 409 body shape.
- `TerraformHttpBackendControllerTest` — Basic + Bearer auth, 401s on missing/invalid/unknown-scheme auth, lock/unlock state machine (new lock, idempotent re-lock by same holder, conflict for different holder, force-unlock by mismatched id, race-on-insert), state GET 404/200, state POST lock enforcement, delete.
- `LocalStorageTypeServiceImplTest` — added cases for `uploadStepOutput`, `uploadTerraformPlan`, `deleteCurrentTerraformState` (present + missing).

35 tests pass locally (`mvn -pl api -am test -Dtest='UploadIntegrityTest,TerraformHttpBackendControllerTest,LocalStorageTypeServiceImplTest'`).

## Test plan

- [ ] CI green
- [ ] Smoke-test from an executor against an API with this PR (covered by the companion executor PR)
- [ ] Verify the liquibase change applies cleanly on a fresh DB


alfespa17:
From what I can see here https://opentofu.org/docs/language/settings/backends/http/

You only need to define the lock and unlock endpoint

```terraform
terraform {
  backend "http" {
    address = "http://myrest.api.com/foo"
    lock_address = "http://myrest.api.com/foo"
    unlock_address = "http://myrest.api.com/foo"
  }
}
```

The controller is exposing this methods `/organization/{organizationId}/workspace/{workspaceId}/lock`  you only need the workspace id, there is a field inside the workspace table to enable or disable the "lock",  `workspace_state_lock` do we need this table?

Maybe you can change this to "unlock"

@DeleteMapping(value = "/organization/{organizationId}/workspace/{workspaceId}/lock")


I am not sure if this is the best way to validate basic auth 🤔

```java
private boolean isAuthorized(HttpServletRequest request) {
        String header = request.getHeader("Authorization");
        if (header == null) {
            return false;
        }
        String token;
        if (header.startsWith("Basic ")) {
            try {
                String decoded = new String(Base64.getDecoder().decode(header.substring("Basic ".length())), StandardCharsets.UTF_8);
                int colon = decoded.indexOf(':');
                token = colon >= 0 ? decoded.substring(colon + 1) : decoded;
            } catch (IllegalArgumentException e) {
                return false;
            }
        } else if (header.startsWith("Bearer ")) {
            token = header.substring("Bearer ".length());
        } else {
            return false;
        }
        if (token == null || token.isBlank()) return false;
        try {
            Jwts.parser().verifyWith(internalJwtKey).build().parseSignedClaims(token);
            return true;
        } catch (Exception e) {
            log.warn("Rejected http-backend request: invalid token ({})", e.getMessage());
            return false;
        }
    }
```

Macbet:
>Persists workspace state locks in a new workspace_state_lock table with WorkspaceStateLock entity, WorkspaceStateLockRepository, and changelog-2.33.0-workspace-state-lock.xml.

Isn't it better to store locks inside Redis/Valkey?  
For example, we can set a lock key with auto-expire, and if the agent fails (for example, due to an infrastructure problem), the lock expires after 2–3 minutes.

alfespa17:
> >Persists workspace state locks in a new workspace_state_lock table with WorkspaceStateLock entity, WorkspaceStateLockRepository, and changelog-2.33.0-workspace-state-lock.xml.
> 
> Isn't it better to store locks inside Redis/Valkey?  
> For example, we can set a lock key with auto-expire, and if the agent fails (for example, due to an infrastructure problem), the lock expires after 2–3 minutes.


I like this idea

jplorier:
> > Persists workspace state locks in a new workspace_state_lock table with WorkspaceStateLock entity, WorkspaceStateLockRepository, and changelog-2.33.0-workspace-state-lock.xml.
> 
> Isn't it better to store locks inside Redis/Valkey? For example, we can set a lock key with auto-expire, and if the agent fails (for example, due to an infrastructure problem), the lock expires after 2–3 minutes.

not sure I follow, but if you fix a lock time, you will not wait for the plan/apply to complete? or is it just a auto unlock after a timeout without a heartbit or something similar?

jplorier:
Thanks for the review. Pushed [625cc249](https://github.com/terrakube-io/terrakube/pull/3214/commits/625cc249) addressing both threads:

## Lock storage → Redis with TTL

Dropped the `workspace_state_lock` SQL table, `WorkspaceStateLock` entity, `WorkspaceStateLockRepository`, and the `changelog-2.33.0-workspace-state-lock.xml` changeset. Replaced with a small `WorkspaceLockService` that stores the raw lock JSON body terraform sends under `tflock:{workspaceId}` in Redis, using `SET NX EX` for atomic acquire.

Behaviour:
- TTL is configurable via `io.terrakube.api.http-backend.lock-ttl-minutes` (default 30). A crashed agent never leaves a workspace stuck.
- The TTL is refreshed on every matching-lock state POST, so a long apply that finishes within the TTL still holds its lock all the way through the operation.
- The `SET NX EX` semantics give us the race resolution that the prior `saveAndFlush`/catch-on-conflict path was simulating.

Re: the workspace table's existing `locked` / `lock_description` columns — those are a separate UI-driven lock toggle (manual workspace freeze) and don't carry a `lock_id`, so I kept them as-is and didn't try to fold the terraform-protocol lock into them. Happy to wire `WorkspaceLockService.tryAcquire` to also flip `workspace.locked` for UI visibility if you'd prefer; that would be a follow-up commit.

## JWT validation

Stopped rolling our own with `Jwts.parser().verifyWith(...)`. The controller now builds a `NimbusJwtDecoder` once via the same wiring [`DexAuthenticationManagerResolver.getJwtEncoder`](https://github.com/terrakube-io/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/security/authentication/dex/DexAuthenticationManagerResolver.java#L89) uses for the `TerrakubeInternal` issuer — `NimbusJwtDecoder.withSecretKey(secret).macAlgorithm(HS256).build()` — and calls `decoder.decode(token)`. So we share the Spring Security JWT path with the rest of the API.

The Basic-auth-with-JWT-password unwrap is still in the controller because terraform's `http` backend cannot send custom headers (`Authorization: Bearer` is unavailable to it) — it can only carry credentials in HTTP Basic. So we extract the token from the Basic header and hand it to the same decoder the Bearer path would have used.

## Tests

- **New** `WorkspaceLockServiceTest` (11 cases): `tryAcquire` uses `SET NX EX` with the configured TTL, returns `false` on loser-of-race, defensively returns `false` when Redis returns `null`; `refresh` extends TTL on the key; `release` deletes; `readLockId` handles valid JSON, blank, malformed, and missing-`ID` inputs.
- **Reworked** `TerraformHttpBackendControllerTest` (17 cases): mocks the new `WorkspaceLockService` instead of a JPA repository; covers Basic + Bearer auth, 401 on missing/invalid/unknown-scheme auth, the lock state machine (new acquire, idempotent re-lock by same holder, conflict for different holder, fallback `{}` body when the winning entry's TTL expires between SETNX and read-back, force-unlock by empty body / mismatched id), state GET 404/200, state POST lock enforcement, and that `postState` calls `refresh` on the matching-lock path.
- Existing `UploadIntegrityTest` (7) and the `LocalStorageTypeServiceImplTest` additions (11 total) unchanged.

46 tests pass locally: `mvn -pl api -am test -Dtest='UploadIntegrityTest,TerraformHttpBackendControllerTest,LocalStorageTypeServiceImplTest,WorkspaceLockServiceTest'`.

The companion executor PR (#3215) needs no changes — the http-backend URL shape is unchanged.

jplorier:
Switched the lock storage to a heartbeat model so a long apply (Helm charts, big terraform stacks) doesn't get cut off by a fixed TTL.

Default TTL is now 5 min (io.terrakube.api.http-backend.lock-ttl-seconds=300). The executor pings a new endpoint once a minute to refresh:


POST /tfstate/v1/http-backend/organization/{org}/workspace/{ws}/lock/heartbeat
200 when the TTL was extended
410 Gone when there's no lock to refresh (executor reads that as "abort the run, the API will hand the lock to whoever asks next")
401 if the JWT doesn't check out
Healthy runs stay locked as long as the executor keeps the heartbeat going. If the executor dies or the network drops, the workspace frees itself in ~5 min.

WorkspaceLockService.refresh now returns whether the key existed so the heartbeat endpoint can give the executor a real signal instead of always 200.

3 new tests on the controller (200/410/401 on heartbeat), 2 more on the service (refresh returns true / false / null defensively), plus the existing post-state test now stubs refresh returning true.

p2pdkivenko (commented):
Really solid feature — the lock state machine, Basic-auth parsing, and integrity check are well-built, and the test coverage (auth negative paths, lock conflict/idempotency/expiry, integrity match/mismatch) is genuinely thorough. A few items worth addressing, the main one being a cross-tenant authorization gap.

### 🔴 Blocking

**1. Missing `@PreAuthorize` on the six new endpoints → cross-tenant state access.** Every sibling method on `TerraformStateController` carries `@PreAuthorize("@stateService.hasManageStatePermission(authentication, #organizationId, #workspaceId)")`, but these new ones don't:
- `TerraformStateController`: `uploadTerraformPlanBinary`, `getHistoryState`, `getHistoryStateJson`, `uploadHistoryState`, `uploadHistoryStateJson`
- `TerraformOutputController`: `uploadStepOutput`

These sit under the authenticated `/tfstate/v1/**` and `/tfoutput/v1/**` paths (not the `permitAll` http-backend path), so with method security enabled the omission silently falls through to `anyRequest().authenticated()`. Any valid principal (any org's user, PAT, or team token) can then read/overwrite **another org's** state, plan binary, history, and step output using the org/workspace/history UUIDs, which aren't secret. Please add the per-workspace `@PreAuthorize` to all six (and a job/workspace-scoped check for `uploadStepOutput`).

**2. `TerraformHttpBackendController.isAuthorized` accepts any internal-signed JWT with no scope.** It validates signature + expiry only — no `iss`/`aud`, and no binding to the `{organization}/{workspace}` in the path. Combined with the companion executor PR minting a 1-day, non-workspace-scoped token (see #3215), holding the internal secret signature is effectively god-mode over all state. Suggest validating `iss`/`aud` via a `JwtValidator` and checking a `workspaceId` claim against the path variable.

### 🟠 Recommended before merge

- **Unlock is delete-by-key, not compare-and-delete.** `WorkspaceLockService.release` deletes by key without checking the holder id; a TOCTOU between `getLockInfo` and `release` lets a late unlock remove a *different* executor's freshly-acquired lock → two concurrent writers. A Lua `if get(k)==id then del(k)` closes it. Short TTL + heartbeat mitigates but doesn't eliminate.
- **Force-unlock (empty/mismatched body id) is unconditional** for any holder of a valid internal token — only safe once the token is workspace-scoped (above).
- **Optional integrity header = silent downgrade.** Omitting `X-Content-Sha256` skips the check; fine for accidental corruption, but if it's a security control consider a `require-integrity-header` flag and document that absent = unverified.

### 🟡 Minor
- Empty `byte[0]` reads return 404, which terraform treats as a fresh/empty workspace — but backends also return `byte[0]` on read *errors*, so a transient storage failure can masquerade as "no state." Pre-existing swallow pattern, newly exposed by these endpoints.
- The four `deleteCurrentTerraformState` impls catch `Exception` and return 200 — a failed delete is reported as success; also please log with the throwable, not just `getMessage()`.
- Please confirm Azure `deleteCurrentTerraformState` uses the same blob key as Azure's `uploadState` (couldn't verify from the diff; AWS/GCP keys check out against their own constants).

Cross-PR note: item 2 here and the token scoping in #3215 need to land together to actually close the hole.

p2pdkivenko:
@jplorier i make small review plz check

jplorier:
🔴 Blocking — fixed
#1 — Missing @PreAuthorize → cross-tenant access. Added the per-workspace check to all five new TerraformStateController endpoints (uploadTerraformPlanBinary, getHistoryState, getHistoryStateJson, uploadHistoryState, uploadHistoryStateJson). For uploadStepOutput (org/job/step, no workspace in the path) I added StateService.hasManageStatePermissionByJob(...), which resolves the job → its workspace, rejects if the job's org ≠ the path org, then reuses the existing manage-state check. Internal tokens still short-circuit, so the executor keeps working. Covered by a new StateServiceTest (not-found / numeric / cross-org / internal-happy-path).

#2 — isAuthorized had no scope. Now, on top of signature+expiry, it requires issuer TerrakubeInternal, audience TerrakubeInternal, and a workspaceId claim that matches the path workspace — on every http-backend endpoint. Four new negative tests (wrong workspace, missing claim, wrong issuer, wrong audience).

⚠️ Cross-PR coupling — needs the executor heartbeat fix. I checked feat/executor-api-routed-state: the backend password (used by terraform for state + lock/unlock) is workspace-scoped, but the heartbeat still mints generateAccessToken(5) (unscoped) at ApiTerraformStateImpl.java:182. With strict enforcement, the heartbeat will 401 every minute and abort runs until that one line becomes generateAccessToken(workspaceId, 5) and a generateAccessToken(String, int) overload exists. As you noted, a "lenient when claim absent" bridge wouldn't actually close the hole (attacker just omits the claim), so strict is the only real fix — and these two PRs must land together. 

🟠 Recommended — fixed
Unlock TOCTOU. Added WorkspaceLockService.releaseIfMatches(...) — an atomic Lua if get==expected then del. Normal unlock now compares-and-deletes against the exact value it just read, so a late unlock can't remove a freshly-acquired lock. Force-unlock (empty body) still releases unconditionally — and that path is now safe because #2 scopes the token to the workspace, so it's no longer org-wide.
Optional integrity header. Added a require-integrity-header flag (io.terrakube.api.upload.require-integrity-header, default false). When true, an absent X-Content-Sha256 on the air-gapped endpoints is rejected with 400 instead of silently downgrading to unverified.
🟡 Minor
Delete logging now logs the throwable (log.error("…", e)) with key/org/ws in AWS/GCP/Azure. (The local impl uses deleteQuietly — no catch.) I kept the best-effort 200-on-failure semantics rather than change the StorageTypeService contract — flagging that as a deliberate choice; say the word if you want delete failures surfaced to the caller.
Azure delete key verified consistent with its uploadState live key (org/ws/terraform.tfstate) — no change needed.
byte[0] → 404 masquerade: left as-is (pre-existing storage-layer swallow, not introduced here) — noting it rather than reworking the storage contract.

jplorier:
@p2pdkivenko @alfespa17 any chance to review the PR to get this moving forward? I'm sorry I have not used the codespace new feature to test, being really busy

p2pdkivenko:
## Code review — lock protocol, state-read safety, and auth placement

Solid feature with good test coverage — the lock service uses proper atomic Redis primitives (SET NX EX, compare-and-delete Lua). However, the lock protocol has two confirmed holes that together allow exactly the stale-write corruption the machinery was built to prevent. Findings ranked most-severe first; each was independently verified against the diff and the base branch.

### 1. Heartbeat refreshes the lock TTL without verifying the caller owns the lock — **confirmed**
`TerraformHttpBackendController` heartbeat → `WorkspaceLockService.refresh(workspaceId)` is a bare `EXPIRE` on the workspace key. If executor A's lock expired and executor B acquired it, A's heartbeat finds B's key, returns **200 instead of the documented 410**, and extends B's lock on its behalf. This silently defeats the "410 → abort the in-flight operation" contract in the endpoint's own javadoc — a bare `EXPIRE` can't distinguish "my lock still exists" from "someone's lock exists."

**Fix:** compare-and-refresh only when the stored lock ID matches the caller's (the heartbeat currently doesn't even read the caller's lock ID from the request).

### 2. `postState` writes state unconditionally when no lock is held — **confirmed**
The lock-match check only runs inside `if (heldLockInfo.isPresent())`. Sequence: A's lock TTL elapses mid-apply → B locks, applies, releases → A's late `POST …/state?ID=a` finds no lock held and silently overwrites B's newer state. The `?ID=` param is discarded in the no-lock branch, but *ID present + no lock held* unambiguously identifies a stale writer and can be rejected with 409 without breaking legitimate `-lock=false` writes (which send no ID). Combined with #1, executor A never receives the 410 that would have made it abort.

### 3. Transient storage-read failures surface as 404, which Terraform treats as "no state" — **confirmed**
All four `getCurrentTerraformState` impls catch broad `Exception` and return `new byte[0]` (e.g. `AwsStorageTypeServiceImpl#downloadObjectFromBucket`), so an S3/Azure/GCP timeout or 5xx is indistinguishable from "state never existed." `getState` maps empty → 404 → Terraform plans against an empty state → on an auto-approved run it recreates (or destroys-and-recreates) already-managed infrastructure. The GET path should distinguish not-found from read-error and return 5xx on the latter.

### 4. PR description promises a `workspace_state_lock` DB table; the diff implements Redis-TTL-only locks — **confirmed**
No `WorkspaceStateLock` entity, repository, or `changelog-2.33.0-workspace-state-lock.xml` exists in the diff (or repo). Beyond the description drift: a Redis restart/failover/eviction silently drops every held lock mid-apply, letting a second executor run a concurrent apply against the same state. Either implement the promised persistent lock (Redis as fast path) or update the description and document the Redis durability requirement.

### 5. `isAuthorized` binds the token to `workspaceId` but never validates the path `organizationId` — plausible
The org flows unvalidated into storage keys (`tfstate/{org}/{ws}/…`), so a caller can read/write its own workspace's state under an arbitrary org prefix — forked state that canonical readers never see. No cross-org data exposure (workspace UUIDs are pinned by the token), so this is defense-in-depth: resolve the workspace and check its real org, as this PR's own `StateService.hasManageStatePermissionByJob` already does for the job path.

### 6. `deleteState` performs no lock check, unlike `postState` — plausible
Any workspace-scoped token can hard-delete the live `terraform.tfstate` while another executor holds the lock mid-apply. DELETE isn't on Terraform's normal locked path, so the realistic trigger is out-of-band cleanup racing an in-flight run — but the write path honors the lock and the destroy path should too.

### 7. `/tfstate/v1/http-backend/**` is `permitAll` + CSRF-exempt with auth hand-rolled per handler — design concern
All six current handlers do call `isAuthorized()`, but the invariant now lives in per-method boilerplate instead of the filter chain: any future handler added under this path that forgets the call is anonymous read/write of Terraform state (fail-open). The controller also duplicates `DexAuthenticationManagerResolver`'s HS256 decoder construction byte-for-byte. Deeper fix: a dedicated `SecurityFilterChain` scoped to this path with a Basic→JWT converter reusing a shared internal-`JwtDecoder` bean, so auth fails closed.

### 8. The read-body → verify → 409 → upload → 201 sequence is copy-pasted 4× — cleanup
Three methods in `TerraformStateController` plus `TerraformOutputController.uploadStepOutput`, differing only in the storage call; the text variants additionally hold both the `byte[]` and a decoded `String` of a potentially large state in memory at once (pre-existing endpoints held one copy). Extract one `readVerifiedBody(request)` helper.

---

**Minor notes** (verified): Azure's new uploads add a per-call `containerClient.exists()` round-trip the pre-existing uploads don't do; `WorkspaceLockService.getLockId()` is dead code (no callers); `sha256Hex` could be `HexFormat.of().formatHex(...)`; the 409 mismatch body interpolates the raw client header into JSON unescaped (malformed error body only, no security impact); lock values are stored JDK-serialized because the shared `RedisTemplate` sets no serializers — works today, but opaque to `redis-cli` and fragile if the template config changes.

**Suspected issues that verification cleared** (no action needed):
- The internal-token contract is satisfiable: `executor/…/WorkspaceSecurityImpl.generateAccessToken()` mints the required camelCase `TerrakubeInternal` iss/aud, `workspaceId` claim, 30-day expiry. One caveat: the API decodes with `io.terrakube.token.internal` while the executor signs with `io.terrakube.client.secretKey` — these must resolve to the same value in deployment config.
- The new authenticated PUTs do **not** need CSRF-ignore entries — existing bearer POSTs to non-ignored endpoints already pass in the checked-in integration tests (`TfcApiTests`).
- The empty-body force-unlock branch correctly matches Terraform's real force-unlock behavior (fresh process, `jsonLockInfo` unset → empty body).

**Recommendation:** hold merge on findings 1–3. The heartbeat ownership check and stale-writer rejection are small, well-localized fixes that close a real state-corruption sequence, and the 404-on-error path is a one-line-per-backend hazard with destructive blast radius.

PR #3215, opened 2026-05-25
feat(executor): API-routed terraform state backend for air-gapped agents

jplorier:
## Summary

- Adds a new `TerraformStateType` (`ApiTerraformStateImpl`) that routes every state, plan, and step-output write through the Terrakube API over HTTPS instead of talking to GCS/S3/Azure/local directly. This lets executors run inside air-gapped networks where only the API endpoint is reachable.
- Emits an HCL override that points terraform at the API's `/tfstate/v1/http-backend/**` endpoints (Basic auth carrying a JWT in the password field, since the http backend can't send custom headers).
- Plan/state/output PUTs include `X-Content-Sha256` so the API can verify what it received before committing. On successful PUT the executor (when `verifyReadBack=true`, the default) GETs the same object back and re-hashes it — closing the gap the announced-hash header can't, since on its own that header proves transit integrity but not that the storage backend persisted the right bytes.
- `StateUploadFailedException` is thrown when upload retries (3 attempts, exponential backoff) are exhausted. The message is user-facing and gets surfaced into the job step output / error log via the new `surfaceUploadFailure` helper in `TerraformExecutorServiceImpl`, so failed runs show a concrete reason instead of silently losing data.
- `UpdateJobStatusImpl` swallows step-output upload failures (output is diagnostic, not infrastructure) but flips the step to failed and clears the URL so the UI doesn't link to a half-uploaded artifact.
- New feature flag `io.terrakube.executor.plugin.tfstate.api.verifyReadBack` (env: `TerraformStateApiVerifyReadBack`, default `true`) lets cost-sensitive workspaces opt out of the GET back-channel.
- README documents the new `TerraformStateType` and feature flag.

This is the executor half of the air-gapped feature; it depends on the companion API PR that adds the `/tfstate/v1/http-backend/**` controller, the `X-Content-Sha256` check, the new history state endpoints, and the `workspace_state_lock` table.

## Tests

- `ApiTerraformStateImplTest` (11 cases) spins up an in-process `com.sun.net.httpserver.HttpServer` and exercises HCL override generation, plan save with integrity + read-back, retry-then-fail on persistent 500s, 409 sha-mismatch surfacing, read-back tampering detection, step output upload, history state (raw + json) uploads, and plan download with / without a URL.
- `UpdateJobStatusImplTest` covers the new catch behavior: step is marked failed and output URL is cleared on `StateUploadFailedException`; normal uploads still record the returned URL.
- `TerraformExecutorServiceImplTest` adds a case where `saveTerraformPlan` throws and the run completes with exit 1 and the error reason in `outputErrorLog`.

Full executor test suite passes locally: 52 tests, 0 failures.

## Test plan

- [ ] CI green
- [ ] End-to-end smoke: spin up an executor pointed at an API with companion PR merged; run plan + apply against a workspace configured with `TerraformStateType=ApiTerraformStateImpl`; confirm state, plan binary, and step output are persisted via the API and that the workspace is locked while running.
- [ ] Repeat with `TerraformStateApiVerifyReadBack=false` to confirm the header-only fast path still works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ The key features of Terrakube are:
 
 - **Remote Backend:** Terrakube supports both `remote backend` and `cloud` block so you can run your workflow directly from the Terraform / OpenTofu CLI.
 
+- **API-Routed Agents (no direct bucket access):** Run executors (agents) that need only outbound HTTPS to the Terrakube API — the agent never connects to GCS / S3 / Azure Blob directly. Terraform uses an `http` backend pointing at the API, and all plan binaries, state history, and step output are streamed through the API which is the only side that talks to the configured object store. Includes a DB-backed workspace lock, sha-256 integrity on every upload, optional read-back verification, and 3-attempt retries before the job is failed with a visible error. Useful when the agent runs in a different cloud / VPC than the server.
+
 ### Getting Started
 
 ### Installation
@@ -52,6 +54,28 @@ The key features of Terrakube are:
 - [Test Terrakube using Minikube](https://docs.terrakube.io/getting-started/deployment/minikube-+-https)
 - [Test Terrakube using Gitpod](https://docs.terrakube.io/getting-started/getting-started)
 - [Develop Terrakube using VS Code Devcontainers](.devcontainer/README.md)
+
+### Enabling API-Routed Agents
+
+This is opt-in per executor. The API server keeps its existing `TerraformStorageType=GCP|AWS|AZURE|LOCAL` — it remains the only process that talks to the bucket. On each executor that should run "air-gapped" from the bucket, set:
+
+```
+TerraformStateType=ApiTerraformStateImpl
+TerrakubeApiUrl=https://<your-terrakube-api>
+InternalSecret=<same value as the API server>
+# Optional, default true. When true the executor GETs each just-uploaded
+# object back from the API and re-hashes it locally to confirm what the
+# storage backend persisted matches what was produced. Disable only if the
+# extra GET traffic is undesirable.
+TerraformStateApiVerifyReadBack=true
+```
+
+What this changes:
+- Terraform uses an `http` backend pointing at the API, so `terraform plan/apply` itself never opens a connection to GCS / S3 / Azure Blob.
+- Plan binaries, state history (raw + JSON) and step output are PUT to the API; the API is the only side that writes to the configured object store.
+- Concurrent runs on the same workspace serialise through a DB-backed lock (`workspace_state_lock` table, added by the bundled Liquibase migration on first start).
+- Every upload carries an `X-Content-Sha256` header. The API rejects payloads whose received hash doesn't match the announced hash (HTTP 409) without writing anything to storage. The executor retries up to 3 times before failing the job with a visible error in the step output — your infrastructure is never silently shifted by a half-transferred state.
+- Existing executors that talk directly to GCS / S3 / Azure / local disk are unaffected; the new mode is selected only when `TerraformStateType=ApiTerraformStateImpl`.
 
 ### Documentation
 To learn more about Terrakube [go to the complete documentation.](https://docs.terrakube.io/) 
--- a/executor/src/main/java/io/terrakube/executor/plugin/tfstate/StateUploadFailedException.java
+++ b/executor/src/main/java/io/terrakube/executor/plugin/tfstate/StateUploadFailedException.java
@@ -1,0 +1,18 @@
+package io.terrakube.executor.plugin.tfstate;
+
+/**
+ * Thrown when the executor exhausts its retries trying to deliver a state,
+ * plan, or output payload to the Terrakube API. The message is user-facing —
+ * it is written to the job step output so the run shows up as failed with a
+ * concrete reason instead of silently losing data.
+ */
+public class StateUploadFailedException extends RuntimeException {
+
+    public StateUploadFailedException(String message) {
+        super(message);
+    }
+
+    public StateUploadFailedException(String message, Throwable cause) {
+        super(message, cause);
+    }
+}
--- a/executor/src/main/java/io/terrakube/executor/plugin/tfstate/TerraformState.java
+++ b/executor/src/main/java/io/terrakube/executor/plugin/tfstate/TerraformState.java
@@ -15,4 +15,27 @@ public interface TerraformState {
     void saveStateJson(TerraformJob terraformJob, String applyJSON, String rawState);
 
     String saveOutput(String organizationId, String jobId, String stepId, String output, String outputError);
+
+    /**
+     * Begin a heartbeat that keeps the API-side workspace lock alive while
+     * this executor holds it. Default no-op for state backends that don't
+     * lock at the API layer (local/AWS/Azure/GCP).
+     */
+    default void startHeartbeat(String organizationId, String workspaceId) {}
+
+    /**
+     * Stop the heartbeat started by {@link #startHeartbeat(String, String)}.
+     * Idempotent. Must be invoked in a {@code finally} block so a crashed
+     * executor still tears the scheduler down cleanly.
+     */
+    default void stopHeartbeat() {}
+
+    /**
+     * Throws {@link StateUploadFailedException} when the heartbeat has been
+     * failing long enough that the API lock TTL is about to (or already has)
+     * elapsed. The executor calls this between polls of long-blocking
+     * terraform operations so the run aborts before the API frees the lock
+     * to a different caller.
+     */
+    default void checkHeartbeatHealth() {}
 }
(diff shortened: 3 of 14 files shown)


jplorier:
Executor side of the heartbeat (companion to #3214). `ApiTerraformStateImpl` now runs a single-thread scheduler while a run is in flight and PUTs `.../lock/heartbeat` every 60s so the API-side TTL is refreshed for as long as terraform is actually doing something — Helm-chart waits and big modules don't get cut off mid-apply.

If the heartbeat fails for too long (default 180s = 3 missed pings, configurable via `io.terrakube.executor.plugin.tfstate.api.heartbeatAbortAfterSeconds`), the next health check throws `StateUploadFailedException` and the executor cancels the in-flight terraform future. That leaves a ~120s margin under the API's 300s TTL — the run aborts before the lock is reassigned to whoever asks next. The error surfaces through the same `surfaceUploadFailure` banner used for upload failures, so the user sees a concrete "Lost contact with the Terrakube API for 180s" in the job logs.

`TerraformState` gained three `default` methods — `startHeartbeat`, `stopHeartbeat`, `checkHeartbeatHealth`. The local/AWS/Azure/GCP impls inherit the no-op so they're unaffected.

`TerraformExecutorServiceImpl` brackets `plan` / `apply` / `destroy` with `startHeartbeat` and `finally { stopHeartbeat() }`. The long-blocking `terraformClient.X(...).get()` calls are replaced with `awaitWithHeartbeat(...)` — a tiny helper that polls the health check every 2s and cancels the future if the heartbeat dies. For non-API state backends this behaves identically to a plain `.get()`.

Two new properties (defaults documented inline in `application.properties`):
- `io.terrakube.executor.plugin.tfstate.api.heartbeatIntervalSeconds` (default 60)
- `io.terrakube.executor.plugin.tfstate.api.heartbeatAbortAfterSeconds` (default 180)

Tests: 6 new cases on `ApiTerraformStateImplTest` (heartbeat ping shape, healthy keeps health quiet, abort threshold trips, 410 doesn't count as success, start/stop idempotency), 3 new on `TerraformExecutorServiceImplTest` (start-then-stop bracket on success, stop-on-init-failure, surface + cancel on heartbeat-loss). 61 executor tests pass locally.

p2pdkivenko (commented):
Nice work — retry/heartbeat lifecycle is clean (idempotent `compareAndSet`, teardown in `finally`), no secrets are logged, and the "swallow step-output failure but mark the step failed" handling is the right call since output is diagnostic, not infra state. Tests are strong. A couple of security items to address, mostly tied to the companion API PR.

### 🔴 Blocking

**1. Backend token is not workspace-scoped and lives 1 day in a plaintext file.** `getBackendStateFile` mints the password via `generateAccessToken(TOKEN_LIFETIME_MINUTES)` (1440 min). The `int` overload of `WorkspaceSecurityImpl.generateAccessToken` doesn't set the `workspaceId` claim that the `String` overload does — so the token written into `api_backend_override.tf` authorizes the bearer against *every* workspace's http-backend endpoints for a full day. Anything that can read the working dir during a run (a malicious provider, `local-exec`, an `external` data source) captures an org-wide state credential. Please mint a workspace-scoped, short-lived token (lifetime ≈ lock TTL), and have the API (#3214) enforce that scope.

**2. No TLS enforcement.** `apiUrl` comes from `TerrakubeApiUrl` with no scheme check; an `http://` value sends the JWT and full state in cleartext. For an air-gapped-over-HTTPS feature, suggest hard-failing on non-`https://` when `ApiTerraformStateImpl` is selected, with an explicit dev-only opt-out flag.

### 🟠 Recommended before merge

- **Orphaned history rows on upload failure.** `saveStateJson` calls `createHistory(...)` *before* the PUTs; if the PUTs exhaust retries, a history row points at an object that was never stored. Create the row after uploads succeed, or roll it back on failure.
- **Read-back GET failure re-PUTs the whole body and can fail a job whose PUT already succeeded.** With `verifyReadBack=true` (default), a transient GET hiccup is treated as a retryable upload failure. Decouple them: retry only the GET, and reserve job failure for an actual hash *mismatch* (the announced-hash header already covers transit integrity).
- **Whole state/plan buffered in memory (twice with read-back).** Regression vs the AWS impl, which streamed from file — large states can OOM a constrained agent. `setFixedLengthStreamingMode` + `DigestInputStream` would stream both directions.
- **Confirm the sha256 contract matches #3214 byte-for-byte** (header name/case, lowercase-hex vs base64, raw received bytes vs re-serialized string). Worth an end-to-end test, since a mismatch 409s every upload.

### 🟡 Minor
- Backend file persists for the run and survives a crash/cleanup-IOException — worth noting for non-ephemeral agents.
- `Charset.defaultCharset()` for the HCL write → use `StandardCharsets.UTF_8`.
- Dead `basicAuthHeader` helper (`@SuppressWarnings("unused")`) — uploads use Bearer, not Basic; removing it avoids implying otherwise.
- `getBackendStateFile` returns `null` on IOException → silent fallback to local/no backend; for air-gapped mode a hard failure is safer.

The PR description says uploads use Basic-auth-with-JWT, but only the terraform http backend block does — the executor's own PUT/GET/heartbeat calls use `Authorization: Bearer`. Minor doc nit.

p2pdkivenko:
@jplorier small review plz check

jplorier:
> @jplorier small review plz check

thanks so much for the review! I'm working on the improvements and will push and tag you as soon as done

jplorier:
@p2pdkivenko 

I didn't address the in memory recommendation as it required a major rework and it might be better for a more experienced contributor.


🔴 #1 — workspace-scoped, shorter-lived backend token

New WorkspaceSecurity.generateAccessToken(workspaceId, minutes) overload that sets the workspaceId claim (the old int overload now delegates to it with workspaceId=null).
[getBackendStateFile](vscode-webview://07tn0t4aotsdo7t58g6hqb2ednric1u4nm99amduds81tkt4gopg/executor/src/main/java/io/terrakube/executor/plugin/tfstate/api/ApiTerraformStateImpl.java#L197) now mints a workspace-scoped token with a configurable lifetime, default 60 min. The TOKEN_LIFETIME_MINUTES=1440 constant is gone.
⚠️ Cross-branch caveat: I confirmed the API's /http-backend/** handler on origin/feat/api-http-backend-air-gapped-state validates only the JWT signature — it never inspects workspaceId. The scoping is inert until that branch enforces token.workspaceId == path.workspaceId. That enforcement is the other half of this fix and lives in #3214.
🔴 #2 — TLS enforcement

[requireSecureApiUrl](vscode-webview://07tn0t4aotsdo7t58g6hqb2ednric1u4nm99amduds81tkt4gopg/executor/src/main/java/io/terrakube/executor/plugin/tfstate/configuration/TerraformStateAutoConfiguration.java) hard-fails bean creation when ApiTerraformStateImpl is selected with a non-https:// URL, with a dev-only opt-out (...api.allowInsecureApiUrl, default false). Enforced at wiring time, so unit tests that build the impl directly against http://127.0.0.1 are unaffected.
🟠 Orphaned history rows — verified the API history PUT stores by id-derived path and needs no pre-existing row, so [saveStateJson](vscode-webview://07tn0t4aotsdo7t58g6hqb2ednric1u4nm99amduds81tkt4gopg/executor/src/main/java/io/terrakube/executor/plugin/tfstate/api/ApiTerraformStateImpl.java#L280) now creates the row after both uploads succeed. New test asserts the row is never created when upload fails.

🟠 Read-back GET coupled to PUT — split into [verifyReadBack](vscode-webview://07tn0t4aotsdo7t58g6hqb2ednric1u4nm99amduds81tkt4gopg/executor/src/main/java/io/terrakube/executor/plugin/tfstate/api/ApiTerraformStateImpl.java): the GET retries independently, a transient read failure no longer re-PUTs the body or fails the job (we accept on the announced hash), and only a confirmed hash mismatch fails. New test covers the transient-GET path.

🟠 sha256 contract — verified byte-for-byte against the API branch: X-Content-Sha256, lowercase-hex, over raw received bytes. Already matches; no change.

🟡 Minor — StandardCharsets.UTF_8 for the HCL write; dead basicAuthHeader + Base64 import removed; getBackendStateFile now throws StateUploadFailedException on IOException instead of silently returning null (caller ignores the return value, so it fails the job hard — correct for air-gapped).

Deferred / not done 
🟠 Whole-state buffering / streaming — deferred to a follow-up, as agreed.
🟡 Backend file persists across crash — left as-is (a note, not a fix).

jplorier:
@p2pdkivenko @alfespa17 any chance to review the PR to get this moving forward? I'm sorry I have not used the codespace new feature to test, being really busy

alfespa17:
> @p2pdkivenko @alfespa17 any chance to review the PR to get this moving forward? I'm sorry I have not used the codespace new feature to test, being really busy

I have been busy with my job too, take your time once you can test it a little bit in codespaces you can add some screenshots  to make sure everything is working correctly

PR #3223, opened 2026-05-26
deps(docker): drop Bitnami images for official upstream

p2pdkivenko:
## Summary

Bitnami deprecated its free image catalog (several images were already pinned to the `bitnamilegacy` tag), so this moves every container image in the repo's compose/dev stacks to official upstream equivalents. **No application code changes** — all swaps are at the image/config layer.

| Service | Before | After |
|---|---|---|
| Redis | `bitnami(legacy)/redis` | `valkey/valkey:8-alpine` |
| PostgreSQL | `bitnami(legacy)/postgresql` | `postgres` |
| MinIO | `bitnami(legacy)/minio` | `minio/minio` + `minio/mc` bucket-init sidecar |
| nginx | `bitnami/nginx` | `nginx:alpine` |
| OpenLDAP | `bitnami(legacy)/openldap` | `osixia/openldap:1.5.0` |

## Notes per service

- **Redis → Valkey**: wire-compatible; password via `--requirepass`, `FLUSHDB`/`FLUSHALL` disabled via `--rename-command`.
- **PostgreSQL → official**: env renamed to `POSTGRES_USER/PASSWORD/DB`; data dir `/var/lib/postgresql/data`. `PGDATA` set to a `pgdata/` subdir so `initdb` targets a clean path even over a pre-existing volume.
- **MinIO → official**: replaces Bitnami's `MINIO_DEFAULT_BUCKETS` with a `minio/mc` init sidecar that creates the bucket; adds `command: server /data --console-address ":9001"`.
- **nginx → official**: conf mounted at `/etc/nginx/conf.d/default.conf`, port `80`.
- **OpenLDAP → osixia**: env model `LDAP_DOMAIN`/`LDAP_ORGANISATION`; Dex repointed `1389 → 389`; LDIF root `dc=` entry dropped (osixia auto-creates the base DN). Seed LDIF is mounted as a **directory** (`ldif/`) with `LDAP_REMOVE_CONFIG_AFTER_SETUP=false` — a single-file bind mount breaks osixia's bootstrap (`sed`/`rm` "Device or resource busy"). Kept on stable osixia v1 (`1.5.0`); the only 2.6.x osixia tag is a breaking alpha.

Spans `docker-compose/`, `.devcontainer/`, `telemetry-compose/`, `scripts/`, `.devfile.yaml`, and `.gitpod.yml`.

## Verification

- All modified compose files pass `docker compose config`.
- OpenLDAP fix verified live: osixia boots, seeds users/groups, serves `:389`; bind `cn=admin/admin` returns `cn=lester` and group `TERRAKUBE_ADMIN`.
- No Bitnami images remain except the intentional `osixia` swap target and the unrelated `bitnami.conf` filename.

## Test plan

- [x] Bring up the devcontainer stack and confirm **Dex login** works (LDAP via osixia on 389).
- [x] Confirm the MinIO `sample` bucket is created by the init sidecar and state/output writes succeed.
- [x] Confirm the API reaches a fresh Postgres and demo data loads.

> Note: the `.devfile.yaml` (Che/Gitpod) path lost Bitnami's env-configurable custom-LDIF dir; osixia v1 has no equivalent, so that flow won't auto-seed users without a follow-up. The docker-compose/devcontainer login path is fully wired.


alfespa17:
I always wanted to update this but I never found the time, thank you @p2pdkivenko, I will check later this week

alfespa17 (approved):

PR #3226, opened 2026-05-29
fix(ui): refresh workspace state after settings save without page reload

### LINKED ISSUE #2900: Settings are not applied to a workspace until page refresh
### Bug description 🐞

If I change the default template or branch of a vcs workspace, settings are not reflected until I do a page reload in the browser

### Steps to reproduce

Edit workspace's settings. Change the default template and/or branch, trigger a new run. The old values are set as defaults until you refresh the page.

### Expected behavior

Values are updated in the run frame without requiring an extra manual refresh

### Example repository

_No response_

### Anything else?

_No response_

hirokimatsueda:
## Summary

Closes #2900

- Workspace settings (General, SSH Key, Webhook, State Shared) now call `onWorkspaceUpdate` after a successful save, triggering `loadWorkspace` in the parent and refreshing React state immediately
- The "Run now" modal (`CreateJob`) now re-fetches `defaultTemplate` and `branch` each time it opens, and applies them via `form.setFieldsValue()` so stale mount-time values are no longer shown

## Root cause

`WorkspaceGeneral` (and other settings tabs) were never wired to the `onWorkspaceUpdate` callback that `Details.tsx` provides. As a result, the parent workspace state was never refreshed after a save — stale values persisted until a full page reload.

Separately, `Create.tsx` fetched workspace defaults only once on mount and relied on `Form.Item initialValue`, which Ant Design only applies at form initialization. Opening the modal after a settings change always showed the original values.

---
Note: The build-jammy CI failure is unrelated to this PR. These tests are in the Java API module. This PR only touches TypeScript/React files under ui/src/.

### DIFF
--- a/ui/src/domain/Jobs/Create.tsx
+++ b/ui/src/domain/Jobs/Create.tsx
@@ -45,6 +45,7 @@ export const CreateJob = ({ changeJob, planJob = true }: Props) => {
       const { branch, defaultTemplate } = response.data.data.attributes;
       setDefaultTemplate(defaultTemplate);
       setBranchName(branch);
+      form.setFieldsValue({ templateId: defaultTemplate, branchName: branch });
     });
   };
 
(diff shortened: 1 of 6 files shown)


alfespa17 (approved):

alfespa17:
thank you @hirokimatsueda

Issue #3224, opened 2026-05-27
Structured plan output seems not to work with OpenTofu


themaroqa:
### Bug description 🐞

After upgrading to 2.31.2 I expected the Structured Plan Output feature (https://github.com/terrakube-io/terrakube/pull/3133) to work out of the box, but it didn't.

After a brief Slack discussion it seems the feature hasn't been completely tested with OpenTofu.

In the executor logs I have `WARN 1 --- [lTaskExecutor-1] i.t.e.s.t.PlanStructuredOutputService    : Unable to get plan json for job 6009 step 0cd513e8-098f-42ef-b758-79e187f435e1. Error:` 
the error is empty so not very useful.

FIY @alfespa17 @arunesh90 

### Steps to reproduce

- Upgrade to version 2.31.2
- Create an OpenTofu (version > 1.11.3) workspace of type Version Control Flow 
- Perform a run of the simplest workspace you can think of the creation of a null resource is enough

### Expected behavior

It should show the structured output

alfespa17:
I think this is more a UI issue, I am testing with this repo https://github.com/terrakube-io/terrakube-docker-compose

For example I have this 2 workspaces.

## Opentofu 

I dont see the structured output

<img width="1214" height="968" alt="Image" src="https://github.com/user-attachments/assets/6488e623-e770-48c0-b1e7-88b16b6227c7" />

Checking the context API, I can see the information related to structured output 

```
GET https://terrakube-api.platform.local/context/v1/1
Authorization: Bearer (PAT)
{
    "planStructuredOutput": {
        "fbd8d308-8e54-49c8-be77-36a3a80f5ab1": [
            {
                "afterSensitive": {},
                "address": "module.time_module.random_integer.time",
                "moduleAddress": "module.time_module",
                "afterUnknown": {
                    "id": true,
                    "result": true
                },
                "before": null,
                "action": "create",
                "resourceName": "time",
                "after": {
                    "min": 1,
                    "seed": null,
                    "max": 5,
                    "keepers": null
                },
                "beforeSensitive": false,
                "actions": [
                    "create"
                ],
                "resourceType": "random_integer"
            },
            {
                "afterSensitive": {},
                "address": "null_resource.next",
                "moduleAddress": null,
                "afterUnknown": {
                    "id": true
                },
                "before": null,
                "action": "create",
                "resourceName": "next",
                "after": {
                    "triggers": null
                },
                "beforeSensitive": false,
                "actions": [
                    "create"
[...]

PR #3230, opened 2026-06-04
fix(executor): fixed structured output for OpenTofu workspaces

arunesh90:
## Summary
Fixes:
- #3224 

This PR fixes the issue by making sure the tofu binary is being used to retrieve the plan's json output when applicable.

Made sure to have Claude do a quick E2E test, including a negative test (so without the fix) to confirm that these changes resolve the issues that were reported when trying to use OpenTofu with structured output.
Tests were added as well to make sure this does not break in the future.


alfespa17 (approved):

alfespa17:
thank you @arunesh90

PR #3249, opened 2026-06-16
fix: use correct casing for third party project and service names

qubeena07:
## Summary

Fixes inconsistent casing of third party product names across the UI, documentation and backend code as reported in #3233.

Changes made:
- `Github` → `GitHub` in UI labels, buttons, dropdowns, README and Java log messages
- `Gitlab` → `GitLab` in the same areas
- `Minio` → `MinIO` in docs
- `DEX` → `Dex` in prose (README and development.md)
- `Gitbook` → `GitBook` in README sponsors section
- `Minikube` → `minikube` in README
- `devcontainer` / `Dev container` → `Dev Containers` in docs
- `Vcs Provider` / `Vcs Identifier` → `VCS Provider` / `VCS Identifier` in workspace import table

The `JobVia` enum values (`Github`, `Gitlab`) in both Java and TypeScript are intentionally left unchanged since they are stored as database records and changing them would require a data migration.

## Test plan

- [ ] Open the VCS settings page and verify GitHub and GitLab button labels display correctly
- [ ] Create a workspace and confirm GitHub and GitLab dropdown options show correct casing
- [ ] Create a module and confirm GitHub and GitLab button text is correct
- [ ] Check README renders correctly on GitHub
- [ ] Check workspace import table shows VCS Provider and VCS Identifier column headings

Closes #3233

PR #3250, opened 2026-06-16
perf: replace hcl2-parser with inline parser to eliminate 3.1 MB bundle chunk

### LINKED ISSUE #3212: Reduce hcl2-parser bundle weight (currently 3.1 MB / 753 KB gzipped)
### Feature description 💡

`hcl2-parser` is the largest single chunk in the UI bundle: **3 118 KB minified / 753 KB gzipped**, bigger than `antd` (1 676 KB). Loaded on every view that parses HCL.

Possible directions:
- Switch to a smaller HCL parser
- Move parsing to the API and return a JSON AST instead of raw HCL
- Trim `hcl2-parser` to only the productions actually used (subset grammar)

### Anything else?

Measured during #3211. Chunk: `hcl-parser-CG3MgaFC.js`. Size unchanged between `main` and the PR branch — i.e. orthogonal to the other bundle work in #3211.

qubeena07:
## Summary

Closes #3212

hcl2-parser was the largest single chunk in the UI bundle at 3118 KB minified / 753 KB gzipped, larger than antd. It was only used in one place: Modules/Details.tsx, to extract variable, output and resource blocks from Terraform HCL files for display in the module details tabs.

This PR replaces it with a small inline parseHcl function (~50 lines) that handles the three block types and attribute patterns the component actually uses. The hcl-parser chunk no longer appears in the build output at all.

### What the parser handles

variable, output and resource block headers with their labels
Single line string values (key = "value")
Bare values including booleans, null, numbers and type expressions like string and list(string)
Strips //, # and block comments before parsing

### What it does not handle (acceptable tradeoffs)

Heredoc descriptions will be blank rather than shown
Multi line default values such as lists or maps spanning multiple lines will not be parsed; the cell will be empty

These edge cases affect display only and are far less common than the straightforward patterns above.

## Test plan

Open a module in the registry and check the Inputs tab shows variable names, types, descriptions and defaults
Check the Outputs tab shows output names and descriptions
Check the Resources tab shows resource type and name pairs
Confirm no hcl-parser chunk appears in the network tab when loading the module details page
Run yarn build and confirm no hcl-parser file in build/assets/

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -13,7 +13,6 @@
     "cron-to-quartz": "^1.4.13",
     "cronstrue": "^3.14.0",
     "github-markdown-css": "^5.9.0",
-    "hcl2-parser": "^1.0.3",
     "html-react-parser": "^6.1.3",
     "html-to-image": "^1.11.13",
     "luxon": "^3.7.2",
--- a/ui/src/domain/Modules/Details.tsx
+++ b/ui/src/domain/Modules/Details.tsx
@@ -57,6 +57,58 @@ const Markdown = lazy(async () => {
   return { default: MarkdownWithPlugins };
 });
 
+type HclAttrs = Record<string, unknown>;
+type HclObject = Record<string, Record<string, unknown>>;
+
+function parseHcl(input: string): [HclObject] {
+  const result: HclObject = {};
+
+  const text = input
+    .replace(/\/\/[^\n]*/g, "")
+    .replace(/#[^\n]*/g, "")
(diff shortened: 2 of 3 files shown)


alfespa17:
@qubeena07 can you help me updating the PR to only include the changes for the hcl2-parser, it also including the changes for the fixes in the text from [here](https://github.com/terrakube-io/terrakube/pull/3249)

qubeena07:
Hi @alfespa17 I have updated the PR. It now only includes the changes for the hcl2-parser. The casing fixes remain in #3249.

PR #3253, opened 2026-06-17
feat: support for semantic version

### LINKED ISSUE #2449: Need support for semantic version for terraform
### Bug description 🐞

While working with an imported workspace, terraform cloud supports the use of ~> in the terraform version number.
The workspace with the imported version fails to run in terrakube and the run stucks. To know it's not going to progress, I checked the executor log and found the issue

```
[threadPoolTaskExecutor-1] INFO io.terrakube.terraform.TerraformDownloader - Downloading terraform version ~>1.11.0 architecture amd64 Type Linux
[threadPoolTaskExecutor-1] ERROR org.springframework.aop.interceptor.SimpleAsyncUncaughtExceptionHandler - Unexpected exception occurred invoking async method: public void io.terrakube.executor.service.executor.ExecutorJobImpl.createJob(io.terrakube.executor.service.mode.TerraformJob)
java.lang.IllegalArgumentException: Invalid Terraform Version
	
```

### Steps to reproduce

Import a workspace with semantic version to use ~> in the version. Run a plan.

### Expected behavior

An error should be sent to the plan and make the plan fail or allow this kind of version definition in the workspace.

### Example repository

_No response_

### Anything else?

_No response_

hirokimatsueda:
## Summary

Closes #2449

Adds support for Terraform/OpenTofu semantic version constraints in workspace settings. Previously, users were forced to select an exact version from a fixed dropdown. This PR replaces the dropdown with a free-text `AutoComplete` input that validates the entered value and still surfaces available versions as suggestions.

### UI

- Replaced the `<Select>` dropdown for Terraform version in **Create Workspace** and **Workspace Settings → General** with `<AutoComplete>`, allowing users to type a constraint while browsing available versions as suggestions.
- Added `validateTerraformVersion` (versionValidation.ts) — a curried validator factory `validateTerraformVersion(availableVersions?)` that applies two layers of checks:
    1. **Format check** — accepts any of the following:
        - Plain semver: `1.11.0`, `1.11`, `1`
        - NPM x-range: `*`, `1.x`, `1.2.*` *(bare `x`/`X` are not accepted — see Known Limitations)*
        - NPM tilde/caret/hyphen: `~1.2.3`, `^1.2.3`, `1.2.3 - 2.3.4`
        - Primitive operators and CocoaPods pessimistic: `>=1.5.7 <1.9.0`, `~>1.11.0`
        - Ivy notation: `[1.0,2.0]`, `[1.7.0,)`, `(,1.14.0]`
    2. **Existence check** — when an exact version (e.g. `1.11.0`) is entered and `availableVersions` is non-empty, validates that the version is actually present in the available releases list. Constraint strings (non-exact) skip this check intentionally.
- Updated help text to inform users that version constraints are supported.

### Executor (AWS S3 backend)

- Fixed `AwsTerraformStateImpl.getBackendStateFile()` to correctly select the S3 backend HCL format when a version constraint is used. Terraform 1.6.0 changed the S3 backend configuration syntax — the `endpoint` attribute was replaced with an `endpoints` block, and `force_path_style` was renamed to `use_path_style` — so Terrakube generates different HCL depending on which syntax the target version requires. Previously this selection only worked with an exact version string; this fix extends it to handle constraints:
    - For range constraints, the **largest version number found literally in the constraint string** is used as a reference for backend selection (e.g. `~>1.11.0` → `1.11.0`, `>=1.5.7 <1.9.0` → `1.9.0` as the upper boundary marker). This is a best-effort heuristic: the code does not resolve the constraint to an actual release, but the highest number in the string is a reliable indicator of the target version range.
    - For X-Range wildcards (`*`, `1.x`, `1.*`), the constraint is treated as "latest" and the new-style backend block is used.

## Tests

- **UI:** Vitest unit test suite covering all supported formats, invalid inputs, and the existence-check behavior with and without an `availableVersions` list (versionValidation.spec.ts).
- **Executor:** JUnit 5 parameterized tests verifying that each constraint style produces the correct backend block, including boundary cases around the 1.6.0 threshold (AwsTerraformStateImplTest.java).

## Known Limitations

- **Bare `x` and `X` wildcards are not supported.** Using `x` or `X` alone as the Terraform version (e.g. entering `x` to mean "any version") results in a validation error in the UI. This is a limitation of `terraform-spring-boot-starter` v1.4.0: `TerraformDownloader` rejects these as an invalid version range at runtime. Use `*` for "any version", or specify a major prefix (e.g. `1.x`, `1.10.x`).

### DIFF
--- a/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java
+++ b/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java
@@ -32,6 +32,8 @@ import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 @Builder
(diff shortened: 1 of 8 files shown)


alfespa17:
thank you @hirokimatsueda 

I did a quick test and it is working correctly

imagen

I can see from the logs that is piking the right version

```shell
2026-06-17T14:28:53.783Z  INFO 9636 --- [lTaskExecutor-1] i.t.terraform.TerraformDownloader        : Downloading terraform version " >=1.5.7 <1.9.0 " architecture amd64 Type Linux
2026-06-17T14:28:53.785Z  INFO 9636 --- [lTaskExecutor-1] i.t.terraform.TerraformDownloader        : Terraform version is " 1.8.5 "
2026-06-17T14:28:53.785Z  INFO 9636 --- [lTaskExecutor-1] i.t.terraform.TerraformDownloader        : terraform_1.8.5_linux_amd64.zip terraform already exists
2026-06-17T14:28:53.963Z  INFO 9636 --- [lTaskExecutor-1] i.t.e.s.t.PlanStructuredOutputService    : 
```

PR #3255, opened 2026-06-17
fix: cascade delete steps and addresses when deleting a job

### LINKED ISSUE #3016: Error "NULL VALUE" When Deleting a Job via API
### Description
An error occurs when attempting to delete a Terrakube job using the DELETE API endpoint, following the official documentation:
https://docs.terrakube.io/api/methods/jobs

#### Request
```shell
curl -L \
  --request DELETE \
  --url "$TERRAKUBE_DEV_URL/api/v1/organization/b74a235d-f2ec-4e04-9312-00bdc9a770f2/job/1565" \
  --header 'Accept: */*' \
  --header "Authorization: Bearer $TERRAKUBE_DEV_TOKEN"
```

#### Error Response
```json
{
  "errors": [
    {
      "detail": "ERROR: null value in column \"job_id\" of relation \"step\" violates not-null constraint\n  Detail: Failing row contains (c0548608-e1ba-433b-85e9-8275417d4a34, 100, https://terrakube-api.apps.xxxx-dev.k8s.app.si.test.org/t..., completed, null, Drift Detection - Auto Prefix Variables, null)."
    }
  ]
}
```

### Observed Behavior

The DELETE request fails with a database constraint error indicating that the job_id column in the step table cannot be null. The error suggests that during the deletion process, a related step record is being updated or orphaned in a way that sets job_id to null, violating the NOT NULL constraint.

### Expected behavior

The job should be deleted successfully, including proper handling (deletion or cascading) of related step records without violating database constraints.

### Example repository

_No response_

### Question
What could be causing this issue?
Thanks

qubeena07:
## Problem

Deleting a job via `DELETE /api/v1/organization/{orgId}/job/{jobId}` fails with:

```
ERROR: null value in column "job_id" of relation "step" violates not-null constraint
```

Hibernate's default behaviour for `@OneToMany` without cascade options is to null out child FK columns before removing the parent row. Since `step.job_id` and `address.job_id` are `NOT NULL`, the update fails before the delete even runs.

## Fix

**JPA (`Job.java`)** — added `cascade = CascadeType.REMOVE` and `orphanRemoval = true` to the `step` and `address` relationships. Hibernate now deletes child rows instead of nullifying their FK.

**Liquibase migration** — drops and recreates `fk_job_step` and `fk_job_address` with `ON DELETE CASCADE` so that direct SQL deletes behave consistently too.

## Test plan

- [ ] `DELETE /api/v1/organization/{orgId}/job/{jobId}` returns 204 with no errors
- [ ] Verify `step` rows for that job are removed from the database
- [ ] Verify `address` rows for that job are removed from the database
- [ ] Deleting a job with no steps/addresses still works

Fixes #3016

### DIFF
--- a/api/src/main/java/io/terrakube/api/rs/job/Job.java
+++ b/api/src/main/java/io/terrakube/api/rs/job/Job.java
@@ -16,6 +16,7 @@ import com.yahoo.elide.annotation.LifeCycleHookBinding;
 import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.annotation.UpdatePermission;
 
+import jakarta.persistence.CascadeType;
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
 import jakarta.persistence.EnumType;
@@ -109,10 +110,10 @@ public class Job extends GenericAuditFields {
     private Workspace workspace;
 
     @UpdatePermission(expression = "user is a super service")
-    @OneToMany(mappedBy = "job")
+    @OneToMany(mappedBy = "job", cascade = CascadeType.REMOVE, orphanRemoval = true)
     private List<Step> step;
 
-    @OneToMany(mappedBy = "job")
+    @OneToMany(mappedBy = "job", cascade = CascadeType.REMOVE, orphanRemoval = true)
     private List<Address> address;
 
 }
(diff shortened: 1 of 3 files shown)


alfespa17:
@qubeena07 can you only include the specific commit for this change please, you are including the previous one too

qubeena07:
@alfespa17 Done! I rebased the branch so now only the cascade delete commit is included. Sorry about that, it was accidentally branched off another fix I had locally.

alfespa17:
@qubeena07  this still include other changes take a look here https://github.com/terrakube-io/terrakube/pull/3255/changes

qubeena07:
@alfespa17 Fixed! The branch was accidentally based on another local branch I had, which pulled in the hcl2 parser removal and the capitalization fixes. I rebased directly onto upstream main and cherry picked only the cascade delete commit so now the PR includes just the three files related to the job delete fix.

alfespa17 (approved):

alfespa17:
Now it was easier to review your PR

PR #3256, opened 2026-06-18
fix(api): mark VCS connection as ERROR when token refresh fails

### LINKED ISSUE #2083: GitLab VCS token expired and not renewed
### Bug description 🐞

Hi team,
I noticed that the GitLab OAuth tokens were not refreshed as expected. The tokens had expired while my cluster was undergoing maintenance, and no automatic refresh was triggered afterwards.
When i made plan on workspace connected to VCS i have plan without commitID and investigating we found problem with expire of GitLab access Token.
Image

To resolve the issue, I had to manually update the status of the tokens to PENDING directly in the database. After that, the reconnection worked without needing to change the access or secret tokens.

It seems like the system didn't retry the refresh once the cluster came back online.
Would it be possible to implement a retry mechanism or a check for expired tokens after downtime?

Let me know if you need any additional logs or details.

Thanks!

Image

### Steps to reproduce

Configure GitLab as VCS and try to change expire in database. 

### Expected behavior

When use vps if token it's expired refresh it, or allow to disconnect and reconnect to vpc from Web-UI

### Example repository

_No response_

### Anything else?

_No response_

mvanhorn:
## What

When an OAuth-based VCS connection (GitLab and others) misses its scheduled token-refresh window — for example because the cluster was down when the hourly Quartz job was meant to run — the access token expires. The next refresh attempt then fails, and `TokenService.refreshAccessToken(...)` returns an empty map. `ScheduleVcs.execute()` previously did nothing in that case, so the connection's `status` stayed `COMPLETED` forever. The connection was permanently broken with no way to reconnect from the UI, and the only known recovery was a manual database edit (set `status` back to `PENDING` and clear the token fields).

This adds an `else` branch in the refresh path: when the refresh returns an empty map, the connection is loaded and its `status` is set to `VcsStatus.ERROR`, then saved. A `WARN` log line records the transition.

## Why

`VcsStatus` already defines `PENDING`, `COMPLETED`, and `ERROR`. Surfacing the failed connection as `ERROR` in the existing status field lets users see the broken connection and re-run the OAuth connect flow instead of silently failing every plan with a missing commitID — no database surgery required.

## How

- `ScheduleVcs.execute()` now sets `VcsStatus.ERROR` and persists it when `refreshAccessToken` returns an empty map. The change is localized to the refresh-failure branch; the success path and the `AZURE_SP_MI` early return are untouched, and the `TokenService` return contract is unchanged.
- Added `ScheduleVcsTest` covering the failure path (`COMPLETED` -> `ERROR`), the success path (tokens updated, status stays `COMPLETED`), and the status guard (only `COMPLETED` connections enter the refresh path).

Closes #2083

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleVcs.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleVcs.java
@@ -87,6 +87,11 @@ public class ScheduleVcs implements org.quartz.Job {
                 tempVcs.setRefreshToken((String) newTokenInformation.get("refreshToken"));
                 tempVcs.setTokenExpiration((Date) newTokenInformation.get("tokenExpiration"));
                 vcsRepository.save(tempVcs);
+            } else {
+                Vcs tempVcs = vcsRepository.getReferenceById(vcs.getId());
+                tempVcs.setStatus(VcsStatus.ERROR);
+                vcsRepository.save(tempVcs);
+                log.warn("VCS connection {} marked ERROR after failed token refresh", vcs.getId());
             }
         }
     }
--- a/api/src/test/java/io/terrakube/api/plugin/scheduler/ScheduleVcsTest.java
+++ b/api/src/test/java/io/terrakube/api/plugin/scheduler/ScheduleVcsTest.java
@@ -1,0 +1,145 @@
+package io.terrakube.api.plugin.scheduler;
+
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


alfespa17:
Thank you @mvanhorn

mvanhorn:
Thanks @alfespa17 for the merge. Marking the VCS connection as ERROR when token refresh is missed surfaces the failure instead of hiding it.

mvanhorn:
Thank you @alfespa17. Marking the VCS connection as ERROR when token refresh fails makes a silent failure visible.

PR #3257, opened 2026-06-18
fix: make group membership checks case insensitive

### LINKED ISSUE #2930: Permission issues when groups are lowercase
### Bug description 🐞

I'm testing terrakube with a keycloak integration. The groups sent by keycloak are lower case (so, for example, terrakube_admin).
I've created those teams in terrakube in lowercase. It allows me to login and run some tasks, but some stuff are failing: 

- approval
- team creation
Looking in the browser console I can see the API is returning a 403 code.

I've worked around that issue by creating a role mapper in keycloak which sends the upper group with a role mapper.

### Steps to reproduce

Create teams in lowercase, have the user log with those lower case groups provided to Dex.

### Expected behavior

_No response_

### Example repository

_No response_

### Anything else?

_No response_

qubeena07:
## Summary

Fixes #2930

Identity providers like Keycloak can send group names in lowercase while the configured owner group or team names in Terrakube use a different casing. The case sensitive `equals` check in `isMember` and `isServiceMember` caused 403 errors on operations requiring superuser access such as job approval and team creation.

Changes:
- `isMember` and `isServiceMember` in `DexGroupServiceImpl` now use `equalsIgnoreCase` so the comparison works regardless of casing from the identity provider
- Added case insensitive JPQL queries to `AccessRepository` and `ProjectAccessRepository` for the repository backed membership checks
- JWT groups are normalized to lowercase before being passed to those queries so all four membership code paths handle case mismatches consistently

## Test plan

- Configure an identity provider to send group names in lowercase (e.g. `terrakube_admin`)
- Set `io.terrakube.owner` to the uppercase equivalent (e.g. `TERRAKUBE_ADMIN`)
- Verify that job approval and team creation no longer return 403
- Verify that existing users with matching case are not affected

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/security/groups/dex/DexGroupServiceImpl.java
+++ b/api/src/main/java/io/terrakube/api/plugin/security/groups/dex/DexGroupServiceImpl.java
@@ -21,6 +21,7 @@ import io.terrakube.api.rs.workspace.access.Access;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 @AllArgsConstructor
 @Slf4j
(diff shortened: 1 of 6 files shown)


alfespa17:
I will not merge this, the group name from the JWT  token should match the configuration in Terrakube.

PR #3262, opened 2026-06-19
fix(api): add missing indexes on foreign-key columns

klinux:
## Problem

Several foreign-key columns that are filtered on every request have no index, so the lookups run as **sequential scans**. Opening a workspace loads its jobs, each of which lazily loads `step`/`address` (N+1), and also looks up `history`/`webhook` by workspace. On a workspace with ~1000 jobs this produces **thousands of sequential scans per page load**, pinning the database CPU at 100% even though the tables are small.

Confirmed with `EXPLAIN ANALYZE` on a real instance:

```
step  WHERE job_id=?       -> Seq Scan, Rows Removed by Filter: 10196
job   WHERE workspace_id=? -> Seq Scan, Rows Removed by Filter: 6045
```

## Changes

Add a Liquibase changeset creating indexes on the FK columns:

- `step.job_id`
- `address.job_id`
- `job.workspace_id`
- `history.workspace_id`
- `webhook.workspace_id`
- `workspacetag.workspace_id`
- `module_version.module_id`

Each `createIndex` is guarded by a `not indexExists` precondition with `onFail="MARK_RAN"`, so the changeset is idempotent on databases where the index was already created by hand.

## Impact (measured on a production instance)

After adding these indexes, the per-query cost dropped from a full table scan (e.g. ~6–22 ms) to an index scan (~0.2 ms), and the database CPU went from saturated back to idle.

## Notes / scope

- This PR is intentionally **indexes only**. Enabling API pagination on `Job`/`History` (to stop clients fetching the full history) is a separate, larger change because the bundled UI currently fetches those collections without pagination or sort, and a small default page size would silently truncate the Runs/States tabs and the computed "current state". That belongs in a follow-up that updates the API **and** the UI together.
- The indexes use plain `createIndex` (Liquibase runs it in a transaction). The affected tables are small, so the lock is momentary. Instances with very large tables that need a zero-lock build can create the indexes out of band first — the preconditions will then skip them.

### DIFF
--- a/api/src/main/resources/db/changelog/changelog.xml
+++ b/api/src/main/resources/db/changelog/changelog.xml
@@ -104,4 +104,5 @@
     <include file="db/changelog/local/changelog-2.31.0-repo-webhooks.xml"/>
     <include file="/db/changelog/local/changelog-2.32.0-project-access.xml"/>
     <include file="/db/changelog/local/changelog-2.32.0-job-step-cascade.xml"/>
+    <include file="/db/changelog/local/changelog-2.32.0-performance-fk-indexes.xml" />
 </databaseChangeLog>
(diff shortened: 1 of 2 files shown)


alfespa17:
If you have a workspace with 1000 jobs you should consider using KEEP_JOB_HISTORY environment variable to delete the old runs

PR #3263, opened 2026-06-19
feat: paginate workspace runs/states (job & history)

klinux:
> **Draft** — the backend change is straightforward, but the UI change needs manual QA on a running instance (see test plan). Opening as draft for review/QA before merge.

## Problem

Opening a workspace loads its jobs and state history through the workspace `include` (`?include=job,history,...`) and the UI derives "last run" and "current state" from those full arrays. On workspaces with many runs/states this fetches the entire history at once (>1 MB, thousands of relationship lookups), making the screen slow.

The `Job` and `History` entities are also not annotated with `@Paginate`, so the API rejects pagination outright:

```
GET .../workspace/{id}/job?page[size]=10  ->  400 {"errors":[{"detail":"Cannot paginate job"}]}
```

## Changes

**API**
- Annotate `Job` and `History` with `@Paginate(defaultPageSize = 100, maxPageSize = 1000)` so clients can request a bounded, sorted page.

**UI**
- Stop requesting `job` and `history` in the workspace `include`.
- Load them via the dedicated sub-collection endpoints sorted newest-first:
  `/workspace/{id}/job?page[size]=100&sort=-id` and the same for `/history`.
- Derive **last run** and **current state** from the most recent item of the sorted result (instead of computing a max over the full include), which also fixes a latent correctness issue if the collections were ever paginated.
- Both lists refresh on the existing 10s polling cycle.

> Note: this loads the most recent page (100) of runs/states. Full page navigation / "load more" controls in the Runs and States tabs can be a follow-up; the sub-collection endpoints already accept `page[number]`.

## Why pagination + UI together

Enabling `@Paginate` alone would change the default behavior of the existing UI (which fetched everything unsorted) and could silently truncate the Runs/States tabs and the computed current state. So the UI is updated in the same PR to fetch these collections explicitly, sorted and bounded.

## Validation done
- `mvn -pl api compile` — OK
- UI `tsc --noEmit` — 0 errors
- UI `eslint` on the changed files — no new errors introduced (pre-existing lint findings only)

## Test plan (needs a running instance — not yet executed)
1. Open a workspace with **>100 runs**: Runs tab shows the 100 most recent (newest first); the latest run status updates live (polling).
2. Open the States tab: shows the most recent states; the **current state** (resources/outputs) matches the latest apply.
3. Trigger a new run → after it completes, the new run appears at the top and current state/resources update.
4. Workspace with **few runs** behaves exactly as before.
5. `GET /workspace/{id}/job?sort=-id&page[size]=10&page[number]=2` returns the second page (no "Cannot paginate" error).


adeeelsameer:
We just hit exactly this in a real deployment, and can confirm both the problem and its blast radius. Adding our experience in case it helps prioritize.

Context: we run Terrakube (2.31.x-based) internally, and recent load testing left several workspaces with 400–500 retained runs. Shortly after, during a period of database pressure, the workspace page's include=job,...,history,... fetch became the query that took the whole UI down:

Every workspace page issues that include-everything query, and it was by far the heaviest query in the system — large response bodies and thousands of relationship lookups, exactly as this PR's description says.
Under database pressure it exceeded the server's response window (~5s in our case): the API closed connections or returned 503s for this query specifically, while lighter endpoints kept serving. Net effect: the UI was effectively unavailable for all workspaces and all users, because there is no workspace page that doesn't issue this query.
We reproduced it outside the browser: the same GET with the full include list failed on every workspace — including nearly-empty ones — while the same fetch without include returned fine. So under load, the unbounded include is the failure mode itself, not any particular workspace's history size.

In other words, this PR fixes not just a performance papercut but what was, for us, a UI availability incident.

@alfespa17  — would love to see this revived. If the remaining blocker is manual QA against an instance with >100-run workspaces: that's precisely what we have (test workspaces with 500+ retained runs, plus a load harness that can generate arbitrarily many), and I'd be happy to run a build of this branch against them and report results back here.

klinux:
Thanks @adeeelsameer — that's a much sharper description of the blast radius than the original PR gave, particularly the part about it failing on nearly-empty workspaces under load. That reframes it: the unbounded `include` is the failure mode by itself, independent of any workspace's history size.

I've rebased this onto current `main` (133 commits had landed since June), resolved the conflicts, and taken it out of draft.

**How the conflicts were resolved**

`Details.tsx` — `main` meanwhile made template loading conditional/cached (`_loadTemplates`). Kept that in full and re-applied this branch's change on top, so the only remaining difference is dropping `job,history` from the include and fetching them afterwards.

`workspaceDataUtils.ts` — `main` had added a `manageState` permission check before `loadState`. Worth flagging since it's easy to lose in a merge: this branch already does the equivalent check inside `loadWorkspaceHistory`, and sources `organizationId`/`workspaceId` from its arguments rather than `sessionStorage`. No behaviour was dropped.

I also removed the eight parameters of `setupWorkspaceIncludes` that became unused once job/history/state moved out, plus a now-orphaned `WORKSPACE_ARCHIVE` import — without that the merge introduced 8 `TS6133` errors.

**Verification**

- `tsc` reports the same 87 pre-existing errors as `upstream/main`, with none in the touched files. I diffed against a baseline run rather than eyeballing it, which is how the 8 above surfaced.
- `eslint`: no new errors. One extra `no-console` warning, from the `console.error` handlers in the new catch blocks, matching the existing convention in that file.
- `vite build` and `mvn -pl api compile` both pass.

**What I could not verify**

Runtime. I have no instance with >100-run workspaces, and that was the original reason this sat in draft — so the offer to run it against your 500+ run workspaces is exactly what it needs. The things most worth exercising:

- Runs/States tabs on a workspace with >100 runs — both lists request `page[size]=100&sort=-id`, so they should show the 100 *most recent*, not the oldest.
- "Last run" and "current state" on the workspace header, since both are now derived from element `[0]` of the sorted sub-collection rather than from the include.
- The 10s polling refresh, and a workspace with zero runs (empty-state path).

@alfespa17 — the API side is just `@Paginate(defaultPageSize = 100, maxPageSize = 1000)` on `Job` and `History`. Worth noting `maxPageSize` caps what a client can request; anything already asking for more than 1000 in one page would start getting an error.

PR #3265, opened 2026-06-20
fix: speed up TFC-compatible workspace list endpoint (N+1 in RemoteTfeService.listWorkspace)

mvanhorn:
## Summary

Speeds up `GET /remote/tfe/v2/organizations/{org}/workspaces` (used by `terraform workspace list`) on organizations with many workspaces by removing two N+1 database patterns in `RemoteTfeService.listWorkspace()`. The JSON:API response is unchanged.

## Why this matters

On large orgs this endpoint took 16s+ to respond (#3091). The hot path re-fetched data the loop already had:

1. For every matched workspace, `getWorkspace(orgName, name, ...)` re-loaded the same workspace from the database by name, even though `listWorkspace()` already held the loaded `Workspace` entity.
2. In the tag-filtered search, `tagRepository.getReferenceById(...)` ran once per workspace tag inside a nested loop.

## Changes

- Added a `getWorkspace(Workspace, Map, JwtAuthenticationToken)` overload that builds the `WorkspaceData` from an already-loaded entity. The existing String-based `getWorkspace(orgName, name, ...)` now resolves the workspace once and delegates to it, so the produced attributes, permissions, and relationships are identical.
- `listWorkspace()` calls the new overload with the workspace it already has, removing the per-workspace re-fetch.
- Replaced the per-tag `getReferenceById` calls with a single org-scoped lookup into a `tagId -> tagName` map, via a new `TagRepository.findByOrganizationName(String)` derived query (org-scoped so it does not scan tags from other organizations).
- Permission checks still run per workspace; nothing is cached across workspaces, so authorization behavior is unchanged.

## Testing

Added `RemoteTfeServiceTest` (Mockito unit tests) covering:
- name-prefix search returns the matching workspaces and does not re-fetch each by name (`getByOrganizationNameAndName` not called per workspace),
- tag-filtered search returns only workspaces matching all requested tags and builds the tag map without per-tag `getReferenceById`,
- a caller without permission still sees the workspaces listed but with the `can-*` permission flags false.

```
cd api && mvn -Dtest=RemoteTfeServiceTest test
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Fixes #3091

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
@@ -425,35 +425,38 @@ public class RemoteTfeService {
 
     WorkspaceData getWorkspace(String organizationName, String workspaceName, Map<String, Object> otherAttributes,
                                JwtAuthenticationToken currentUser) {
-        Optional<Workspace> workspace = Optional
-                .ofNullable(workspaceRepository.getByOrganizationNameAndName(organizationName, workspaceName));
+        Workspace workspace = workspaceRepository.getByOrganizationNameAndName(organizationName, workspaceName);
+        return getWorkspace(workspace, otherAttributes, currentUser);
+    }
 
-        if (workspace.isPresent()) {
-            log.info("Found Workspace Id: {} Terraform: {} Global Remote State: {}", workspace.get().getId().toString(),
-                    workspace.get().getTerraformVersion(), workspace.get().isGlobalRemoteState());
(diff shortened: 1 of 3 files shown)


alfespa17:
Hello @mvanhorn, I think this looks good just one quick comment about the test cases, can mark as "deleted" the workspaces that you created at the end of each test. 

There are some test cases that uses the demo data and if you add some workspace those will start to fail because the size of the response wont match.

alfespa17:
My mistake you are using mockito

PR #3269, opened 2026-06-23
feat: shallow git clone (--depth 1) for VCS workspace checkout

### LINKED ISSUE #3238: git clone taking long time on monorepo
### Feature description 💡

git clone taking long time on monorepo, would be good to have something like https://git-scm.com/docs/git-sparse-checkout so it does not clone entire monorepo but just one/multiple folders needed 

### Anything else?

_No response_

mvanhorn:
## Why

Cloning a workspace repository takes a long time for large monorepos because the executor performs a full clone of the entire repository history on every job (reported in #3238). For a normal plan/apply only the requested branch tip is needed, so the full-history transfer is wasted work. The reporter noted that `git clone --depth 1` should be relatively easy, and the maintainer agreed it was worth trying.

## What

Performs a shallow git clone (`--depth 1`) of the requested branch tip when checking out a workspace's VCS repository, instead of always cloning the entire repository history.

jgit's `CloneCommand` supports `setDepth(int)`, so the change is small and contained to the executor clone helper.

## How

In `SetupWorkspaceImpl.downloadWorkspaceGit(...)`:

- A depth-1 shallow clone cannot check out an arbitrary historical commit, so the optimization is applied **only when `terraformJob.getCommitId()` is null or blank**.
- When a specific `commitId` is requested, the clone falls back to a full clone (no `setDepth`) so the subsequent `Git.open(...).checkout().setName(commitId)` still resolves the historical commit. The existing checkout block is unchanged.
- The gating is applied to both the SSH clone path and the credentials clone path. `setBranch(...)` and `setCloneSubmodules(true)` are unchanged, so the branch tip (and submodules) are still fetched.

## Testing

- Added `performsShallowCloneWhenNoCommitIdRequested` to `SetupWorkspaceTest`, which asserts the clone is shallow (the jgit `.git/shallow` marker file is present) for the no-`commitId` happy path.
- The existing `failsWhenAskedToCheckoutABadCommit` test continues to cover the full-clone fallback: a bad commit id still throws `RefNotFoundException`, which only succeeds because a `commitId` job performs a full clone.
- `mvn -Dtest=SetupWorkspaceTest test` passes.

Fixes #3238

### DIFF
--- a/executor/src/main/java/io/terrakube/executor/service/workspace/SetupWorkspaceImpl.java
+++ b/executor/src/main/java/io/terrakube/executor/service/workspace/SetupWorkspaceImpl.java
@@ -11,6 +11,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import com.azure.core.credential.AccessToken;
 import com.azure.core.credential.TokenRequestContext;
@@ -25,10 +26,15 @@ import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.eclipse.jgit.api.CloneCommand;
+import org.eclipse.jgit.api.FetchCommand;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.errors.MissingObjectException;
+import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.revwalk.RevWalk;
 import org.eclipse.jgit.transport.CredentialsProvider;
+import org.eclipse.jgit.transport.RefSpec;
 import org.eclipse.jgit.transport.SshTransport;
 import org.eclipse.jgit.transport.UsernamePasswordCredentialsProvider;
 import org.eclipse.jgit.transport.sshd.JGitKeyCache;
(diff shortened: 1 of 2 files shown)


alfespa17:
@mvanhorn you need to consider one thing, imaging you are running a plan->apply template

The first time the executor clone the repo it will do a shallow git clone (running plan), it will get the commit id information in this part of the code.

https://github.com/terrakube-io/terrakube/blob/9283073c19026af40c8efb2076025624d8e9cd94/executor/src/main/java/io/terrakube/executor/service/workspace/SetupWorkspaceImpl.java#L185.

Later It will save the commit id inside the job definition here

https://github.com/terrakube-io/terrakube/blob/9283073c19026af40c8efb2076025624d8e9cd94/executor/src/main/java/io/terrakube/executor/service/status/UpdateJobStatusImpl.java#L47

https://github.com/terrakube-io/terrakube/blob/9283073c19026af40c8efb2076025624d8e9cd94/executor/src/main/java/io/terrakube/executor/service/status/UpdateJobStatusImpl.java#L52

When the executor runs the apply it will checkout all the history because the commit id will be update in the job definition.

So this PR partially fix the problem, the first time it will do a `shallow git clone` but later executions will still do a `full git clone`

mvanhorn:
Great catch, that scenario would have broken applies whenever the branch advanced between plan and apply. Fixed in 69eed08: the checkout path now checks whether the recorded commit id exists in the shallow clone; if not, it fetches that SHA directly at depth 1, and if the server rejects SHA-in-want it falls back to unshallowing before checkout. The happy path (tip checkout, or commit already present) stays depth-1, so the clone-speed win from this PR is preserved. Added tests that simulate exactly your plan->apply case: shallow clone, advance the source branch, then check out the older recorded commit.

alfespa17 (approved):

alfespa17:
@mvanhorn I did some test and it is working correctly, thank you for your help

mvanhorn:
Thanks for testing and merging the shallow clone support, @alfespa17.

mvanhorn:
Thanks @alfespa17! Shallow clones for VCS workspace checkout should speed up big-repo workspaces noticeably.

PR #3291, opened 2026-07-02
fix(executor): surface init failures on remote-content runs

### LINKED ISSUE #3290: Executor: terraform init failures invisible on remote-content (CLI-driven) runs — empty step log
### Bug description 🐞

When a job runs as a remote operation (CLI-driven run, workspace branch / override branch `remote-content`), `terraform init` output is not forwarded to the step output. If init fails, the job fails with a **completely empty step log**: nothing in the stored step output, nothing streamed back to the terraform CLI. The only place the error appears is the executor pod log.

### Where

`ExecutorService` disables headers for remote operations:

https://github.com/terrakube-io/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java (`setShowHeader(false)` for `remote-content`)

`TerraformExecutorServiceImpl.executeTerraformInit` then reuses that flag to decide where the whole init stream goes — with `showHeader == false` both stdout and stderr consumers are `log.info` lambdas, so the step output consumer never sees a single line:

```java
if (terraformJob.isShowHeader()) {
    initSuccessful = Boolean.TRUE.equals(terraformClient.init(terraformProcessData, output, errorOutput).get());
} else {
    initSuccessful = Boolean.TRUE.equals(terraformClient.init(terraformProcessData, s -> {
        log.info(s);
    }, s -> {
        log.info(s);
    }).get());
}
```

Suppressing the Terrakube banner for remote runs makes sense (the CLI draws its own frame), but suppressing init **errors** does not.

### Steps to reproduce

1. VCS workspace with any config error that breaks `terraform init` (e.g. a module using a non-hashicorp provider without `required_providers`, so terraform resolves `hashicorp/<name>` and gets a registry 404).
2. `terraform plan` against the workspace via the cloud/remote backend (CLI-driven run).
3. Job fails. CLI shows the run errored with no diagnostics, UI step log is empty, `tfoutput` endpoint returns an empty body.
4. Executor pod log contains the full init error.

The same config pushed via VCS webhook shows the init error in the step log as expected (`showHeader == true` path).

### Expected behavior

Init failures are visible to the user: in the stored step output and in the terraform CLI stream, same as VCS-driven runs.

### Versions

- Observed on a 2.3x self-hosted instance (helm chart, remote executor)
- Code path unchanged on current `main`

mikluko:
## Description

Fixes #3290.

Remote operations (CLI-driven runs, `showHeader == false`) routed the entire `terraform init` stream to the executor log only. When init failed, the step output stayed empty: nothing in the stored step log, nothing streamed back to the terraform CLI — the failure was diagnosable only from the executor pod log.

This change keeps init quiet on success for remote operations (the terraform CLI draws its own frame, and TFC behaves the same way), but buffers the init lines and flushes them to the step output consumer when init fails. Stderr is merged into stdout by `setRedirectErrorStream(true)`, so a single buffered consumer covers both.

VCS-driven runs (`showHeader == true`) are unchanged.

## Testing

- Reproduced on a self-hosted instance: CLI-driven plan with a module missing a `required_providers` entry for a non-hashicorp provider → init fails at provider resolution → empty step log before the change; with the change the init error reaches the step output.
- VCS webhook runs on the same config keep showing the full init section as before.

### DIFF
--- a/executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
+++ b/executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
@@ -478,11 +478,20 @@ public class TerraformExecutorServiceImpl implements TerraformExecutor {
         if (terraformJob.isShowHeader()) {
             initSuccessful = Boolean.TRUE.equals(terraformClient.init(terraformProcessData, output, errorOutput).get());
         } else {
-            initSuccessful = Boolean.TRUE.equals(terraformClient.init(terraformProcessData, s -> {
+            // Remote operations (CLI-driven runs) keep init quiet on success, but the
+            // stream must still reach the step output when init fails; otherwise the
+            // error is only visible in the executor log and the client sees an empty
+            // step. Buffer the lines (stderr is merged into stdout via
+            // setRedirectErrorStream) and flush them on failure.
+            TextStringBuilder initOutput = new TextStringBuilder();
+            Consumer<String> quietOutput = s -> {
                 log.info(s);
-            }, s -> {
-                log.info(s);
-            }).get());
+                initOutput.appendln(s);
+            };
+            initSuccessful = Boolean.TRUE.equals(terraformClient.init(terraformProcessData, quietOutput, quietOutput).get());
+            if (!initSuccessful) {
+                output.accept(initOutput.toString());
+            }
         }
 
         log.warn("Terraform init Executed Successfully: {}", initSuccessful);


alfespa17 (approved):

mikluko:
@alfespa17 Thank you very much for your close-to-instant reviews! These are extremely valuable these days, when the usual feedback loop might take days and weeks.

alfespa17:
> @alfespa17 Thank you very much for your close-to-instant reviews! These are extremely valuable these days, when the usual feedback loop might take days and weeks.

Thank you for sending small pull request with a good description of the change, those are easy to review 😅.

PR #3303, opened 2026-07-09
Allow region configuration for custom S3 clients

scheibling:
Allows region selection for non-AWS clients, setting the region to "auto" can mess with access tokens for certain S3 providers and/or servers like garage.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.java
+++ b/api/src/main/java/io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.java
@@ -76,7 +76,7 @@ public class StorageTypeAutoConfiguration {
                             .build();
 
                     s3client = S3Client.builder()
-                            .region(Region.of("auto"))
+                            .region(awsStorageTypeProperties.getRegion())
                             .credentialsProvider(StaticCredentialsProvider.create(getAwsBasicCredentials(awsStorageTypeProperties)))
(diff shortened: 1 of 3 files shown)


scheibling:
This is only a small subset of the changes proposed by @SolomonHD, but those seem to be a bit stuck :-)

alfespa17:
I think you will need to update the default value in all the properties files (api,executor and registry)  to "auto"

Example for the API:

https://github.com/terrakube-io/terrakube/blob/14a0e66d43232d7ca8a9788678de4adf660d8ab6/api/src/main/resources/application.properties#L56

https://github.com/terrakube-io/terrakube/blob/14a0e66d43232d7ca8a9788678de4adf660d8ab6/api/src/main/resources/application-demo.properties#L56

The helm chart does not include the value for region, so you will still need to update the helm chart so it does not use auto by default

Examples: 

https://github.com/terrakube-io/terrakube-helm-chart/blob/d65a0ec7e278d2a15297dad324576129814c0039/charts/terrakube/values.yaml#L140

https://github.com/terrakube-io/terrakube-helm-chart/blob/d65a0ec7e278d2a15297dad324576129814c0039/charts/terrakube/templates/secrets-api.yaml#L42

scheibling:
I'll put this back to draft, for some reason I haven't been able to find yet those changes don't work as consistently as I'd like, the executor and/or registry are having some issues... and a quick fix turned into another project 😁 

I'll have to come back to it after coming back from vacation

alfespa17:
Fixed in https://github.com/terrakube-io/terrakube/pull/3473

Issue #2255, opened 2025-06-30
Can't Import Terraform Cloud Workspace into Bitbucket VCS Terrakube workspace


yeahdj:
### Bug description 🐞

You get an error when trying to import a TFC workspace into Terrakube with VCS workflow enabled.

### Steps to reproduce

Workspaces > Import Workspace > Terraform Cloud > VCS Workflow > Select Bitbucket > Provide TFC credentials (org name and token) 

Produces this error

```
"workspace
There was an error creating the workspace:could not execute statement [ERROR: null value in column "source" of relation "workspace" violates not-null constraint Detail: Failing row contains (eff94235-b8d3-4dd5-ad75-b7db30823f45, workspace, null, main, ~>1.9.0, 3d772bd9-c7df-42bb-b675-8cd5520c4894, bdb56b06-20e2-485d-904d-c18522f4fe7c, null, null, null, f, f, 2025-06-30 16:01:31.736, 2025-06-30 16:01:31.736, admin@example.com, admin@example.com, remote, null, terraform, null, null, null, f).] [insert into workspace (agent_id,allow_remote_apply,branch,created_by,created_date,default_template,deleted,description,execution_mode,folder,iac_type,lock_description,locked,module_ssh_key,name,organization_id,source,ssh_id,terraform_version,updated_by,updated_date,vcs_id,id) values (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)]; SQL [insert into workspace (agent_id,allow_remote_apply,branch,created_by,created_date,default_template,deleted,description,execution_mode,folder,iac_type,lock_description,locked,module_ssh_key,name,organization_id,source,ssh_id,terraform_version,updated_by,updated_date,vcs_id,id) values (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)]; constraint [source" of relation "workspace]"
```

### Expected behavior

Import completes successfully.

alfespa17:
Maybe you can add some screenshots, looks like there is some value that is empty when creating the workspace, not sure if you are creating a VCS workspace or CLI workspace

yeahdj:
It's a VCS workspace - using bitbucket. I can't really share any screenshots as it's corporate data. I click import workspace, provide my terraform cloud organization name and token, from there I get a list of TFC workspaces to choose from, and when I click any one - I get the error mentioned in the original post (although i've changed the name to workspace)

[...]

PR #3308, opened 2026-07-14
fix: widen vcs access_token/refresh_token columns for oversized OAuth tokens

mvanhorn:
## What

Adds a Liquibase migration (`changelog-2.33.0-vcs-token-size.xml`) that widens `vcs.access_token` and `vcs.refresh_token` to `varchar2(4096)`, and registers it in the master `changelog.xml`.

## Why

When configuring a Bitbucket Cloud (and other OAuth) VCS provider, saving the `vcs` row fails with:

```
value too long for type character varying(2048)
```

Atlassian and other providers now issue OAuth access/refresh tokens longer than the current column widths. `access_token` was previously widened from `varchar2(1024)` to `varchar2(2048)` in `changelog-2.26.0-size-access-token.xml`, but `refresh_token` is still `varchar2(1024)`, and even 2048 is no longer enough for current tokens. A reporter confirmed the fix by manually running `ALTER TABLE public.vcs ALTER COLUMN access_token TYPE varchar(4096)` (and the same for `refresh_token`).

This follows the maintainer's suggestion on the issue to send a Liquibase migration to increase the column sizes.

## How

- New file `api/src/main/resources/db/changelog/local/changelog-2.33.0-vcs-token-size.xml` with two `modifyDataType` changeSets (`2-33-0-1`, `2-33-0-2`) widening `vcs.access_token` and `vcs.refresh_token` to `varchar2(4096)`, following the exact shape of the existing `changelog-2.26.0-size-access-token.xml`.
- One `<include>` line appended to the master `changelog.xml` after the `2.32.0-*` entries.

The abstract Liquibase `varchar2` type maps to `varchar` on Postgres and `varchar2` on Oracle, matching how the existing size migration already ships cross-DB without a Postgres-specific variant. No entity or Java change is needed since the JPA `String` fields are unbounded at the app layer; the constraint was purely the DB column width.

## Scope

This addresses the column-size half of the issue (the `varchar(2048)` overflow the maintainer invited a migration for). The separately reported `null value in column "source"` workspace-import error is a distinct problem in the TFC import path and is out of scope here.

Refs #2255

### DIFF
--- a/api/src/main/resources/db/changelog/changelog.xml
+++ b/api/src/main/resources/db/changelog/changelog.xml
@@ -105,4 +105,5 @@
     <include file="/db/changelog/local/changelog-2.32.0-project-access.xml"/>
     <include file="/db/changelog/local/changelog-2.32.0-job-step-cascade.xml"/>
     <include file="/db/changelog/local/changelog-2.32.0-performance-fk-indexes.xml" />
+    <include file="/db/changelog/local/changelog-2.33.0-vcs-token-size.xml"/>
 </databaseChangeLog>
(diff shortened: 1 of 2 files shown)


alfespa17:
@mvanhorn can you update the liquibase author name 😅, maybe you can put your GitHub username

mvanhorn:
Done - set both changeSet authors to `mvanhorn` in 632b0b8. Thanks!

mvanhorn:
Thanks for merging, @alfespa17 - oversized OAuth tokens will not truncate in the vcs columns anymore.

PR #3336, opened 2026-07-24
fix: validate GitHub App token expiry and clean up orphaned VCS refs

### LINKED ISSUE #3317: GitHub App installation token served from cache without expiry validation (stale-token 401s)
## Summary

GitHub App installation tokens are cached in the `github_app_token` table keyed by `(appId, owner)` and returned on read **without any expiry validation and without re-fetching on auth failure**. Because GitHub App installation tokens expire after 1 hour, any gap in the background refresh job leaves a stale token that is served indefinitely, causing every clone for that owner to fail with `not authorized` until the next successful refresh — with no self-healing on the read path.

## Affected code

`api/src/main/java/io/terrakube/api/plugin/vcs/provider/github/GitHubTokenService.java` (as of `main` @ `22a4748c`):

```java
public GitHubAppToken getGitHubAppToken(Vcs vcs, String[] ownerAndRepo) ... {
    GitHubAppToken gitHubAppToken = gitHubAppTokenRepository.findByAppIdAndOwner(vcs.getClientId(), ownerAndRepo[0]); // ~L121
    if (gitHubAppToken == null) {                          // only re-fetches when NO row exists
        gitHubAppToken = fetchGitHubAppInstallationToken(vcs, ownerAndRepo);
    }
    return gitHubAppToken;                                 // ~L130 — returned regardless of token age/validity
}
```

The token is refreshed only by the periodic `ScheduleGitHubAppToken` Quartz job (`repeatSecondlyForever`, 3300s). The `GitHubAppToken` entity stores no `expiresAt`/`fetchedAt`, so the read path has no way to know the token is stale.

## Impact

- After any refresh gap (API downtime spanning the interval, a refresh that throws — see companion issue on `refreshAccessToken` and a missing VCS — clock skew, etc.), the cached token expires and **all** git operations for that owner fail:
  ```
  Failed to prepare work dir
  org.eclipse.jgit.api.errors.TransportException: https://github.com/<owner>/<repo>.git: not authorized
  ```
- There is no recovery on read: the row still exists, so `getGitHubAppToken` keeps returning the expired token. Only a successful background refresh (or manual deletion of the row) fixes it.

## Steps to reproduce

1. Configure a STANDALONE GitHub App VCS and run a workspace so a `github_app_token` row is created.
2. Prevent the refresh job from succeeding for >1 hour (e.g., stop the API, or trigger the companion `refreshAccessToken` failure by deleting/recreating the VCS).
3. Run the workspace → clone fails with `not authorized`, and keeps failing even though the App/installation/permissions are all valid (verified with a manually minted installation token).

## Suggested fix

- Persist token expiry (`expires_at`, returned by `POST /app/installations/{id}/access_tokens`) on `GitHubAppToken`.
- In `getGitHubAppToken`, treat an expired/near-expiry cached token as a miss and re-fetch synchronously.
- Optionally, re-fetch once on a 401 from the git transport instead of relying solely on the scheduled refresh.

### LINKED ISSUE #3318: refreshAccessToken NPEs and leaves a stale GitHub App token when no VCS matches the appId (deleted/recreated VCS)
## Summary

`GitHubTokenService.refreshAccessToken()` looks up the VCS by app/client id and dereferences the result without a null check. If no VCS matches that id — e.g. a GitHub App VCS was **deleted and recreated** (the `github_app_token` row survives, keyed by `appId`), or otherwise removed — the lookup returns `null`, the refresh throws, the token is never updated, and the stale token is left in place. Combined with the read path not validating token expiry (companion issue), this produces a permanent `not authorized` state that only manual DB cleanup resolves.

## Affected code

`api/src/main/java/io/terrakube/api/plugin/vcs/provider/github/GitHubTokenService.java` (as of `main` @ `22a4748c`):

```java
public String refreshAccessToken(GitHubAppToken gitHubAppToken) ... {
    Vcs vcs = vcsRepository.findFirstByClientId(gitHubAppToken.getAppId()); // ~L112 — can return null
    String jws = generateJWT(vcs.getClientId(), vcs.getPrivateKey());       // NPE if vcs == null
    return fetchGitHubAppInstallationToken(gitHubAppToken.getInstallationId(), vcs.getApiUrl(), jws, gitHubAppToken.getOwner());
}
```

The scheduled refresh job swallows the exception and, on a null token, keeps the previous value:

```java
// ScheduleGitHubAppToken.execute()
token = gitHubTokenService.refreshAccessToken(appToken);  // throws -> caught, logged
...
if (token != null) { appToken.setToken(token); save(); }  // stale token retained on failure
```

## Impact

- Deleting/recreating a GitHub App VCS connection (same underlying App ID) leaves an **orphaned/mismatched `github_app_token` row**. During the window with no matching VCS, every refresh throws (NPE), the token expires after 1h, and clones fail with:
  ```
  org.eclipse.jgit.api.errors.TransportException: https://github.com/<owner>/<repo>.git: not authorized
  ```
- Because the read path (`getGitHubAppToken`) never validates expiry, the stale token is served indefinitely. Recovery requires manually deleting the `github_app_token` row.

## Steps to reproduce

1. Set up a STANDALONE GitHub App VCS, run a workspace (creates a `github_app_token` row + a 55-min refresh job).
2. Delete the VCS and recreate it (or delete it entirely). The `github_app_token` row persists.
3. Wait for the refresh job to fire → it NPEs on the null VCS; token is not updated. After 1h the token expires → clones return `not authorized` with no self-recovery.

## Suggested fix

- Null-check the VCS in `refreshAccessToken`; if no VCS matches the `appId`, delete the orphaned `github_app_token` row and its refresh job (Quartz trigger) instead of throwing.
- On refresh failure, consider invalidating/removing the stale token rather than retaining it, so the read path re-fetches (paired with the expiry-check fix in the companion issue).

jakegroves:
## Description

GitHub App installation tokens were cached indefinitely with no expiry check on read, so any gap in the background refresh job left every clone for that owner permanently failing with `not authorized`. Separately, if a GitHub App VCS was deleted and recreated, `refreshAccessToken` NPE'd on the missing VCS instead of cleaning up the orphaned token row, compounding the same failure with no self-recovery.

## How solved

- Added `expires_at` to `GitHubAppToken` (new Liquibase migration), populated from GitHub's installation-token response.
- `getGitHubAppToken` now treats a missing/near-expiry token as a cache miss and refreshes it in place (same row, no duplicates) instead of blindly returning it.
- `refreshAccessToken` null-checks the VCS lookup; on an orphaned token (no matching VCS) it deletes the row and cancels its Quartz refresh trigger instead of throwing.

## Tests added

- `GitHubTokenServiceTest` (6 cases): expired/missing-expiry tokens trigger a re-fetch, valid cached tokens are served with zero HTTP calls, and orphaned-VCS refresh cleans up the token row + schedule (including when trigger deletion itself fails).

Fixes #3317
Fixes #3318

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/vcs/provider/github/GitHubTokenService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/vcs/provider/github/GitHubTokenService.java
@@ -7,6 +7,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Base64;
@@ -48,6 +49,9 @@ import reactor.netty.transport.ProxyProvider;
 public class GitHubTokenService implements GetAccessToken<GitHubToken> {
 
     private static final String DEFAULT_ENDPOINT = "https://github.com";
+    // Treat a cached token as expired slightly before GitHub actually expires it,
+    // so a request never races the real expiry.
+    private static final Duration EXPIRY_SAFETY_BUFFER = Duration.ofMinutes(1);
 
     @Autowired
     ObjectMapper objectMapper;
@@ -110,9 +114,23 @@ public class GitHubTokenService implements GetAccessToken<GitHubToken> {
     public String refreshAccessToken(GitHubAppToken gitHubAppToken)
             throws NoSuchAlgorithmException, InvalidKeySpecException, JsonMappingException, JsonProcessingException {
         Vcs vcs = vcsRepository.findFirstByClientId(gitHubAppToken.getAppId());
+        if (vcs == null) {
+            log.warn("No Vcs found for GitHub App id {}, removing orphaned GitHubAppToken {} for owner {}",
+                    gitHubAppToken.getAppId(), gitHubAppToken.getId(), gitHubAppToken.getOwner());
+            try {
+                scheduleGitHubAppTokenService.deleteTask(gitHubAppToken.getId().toString());
+            } catch (SchedulerException e) {
+                log.error("Failed to delete refresh schedule for orphaned GitHubAppToken {}, error {}",
+                        gitHubAppToken.getId(), e);
+            }
+            gitHubAppTokenRepository.delete(gitHubAppToken);
+            return null;
+        }
         String jws = generateJWT(vcs.getClientId(), vcs.getPrivateKey());
-        return fetchGitHubAppInstallationToken(gitHubAppToken.getInstallationId(), vcs.getApiUrl(), jws,
-                gitHubAppToken.getOwner());
+        GitHubAppInstallationToken installationToken = fetchGitHubAppInstallationToken(
+                gitHubAppToken.getInstallationId(), vcs.getApiUrl(), jws, gitHubAppToken.getOwner());
+        gitHubAppToken.setExpiresAt(installationToken.expiresAt());
+        return installationToken.token();
     }
 
     public GitHubAppToken getGitHubAppToken(Vcs vcs, String[] ownerAndRepo)
@@ -122,6 +140,15 @@ public class GitHubTokenService implements GetAccessToken<GitHubToken> {
         if (gitHubAppToken == null) {
             log.info("No token found in GitHubAppToken table, fetching new token");
             gitHubAppToken = fetchGitHubAppInstallationToken(vcs, ownerAndRepo);
+        } else if (isExpired(gitHubAppToken.getExpiresAt())) {
+            log.info("Cached GitHub App token for user/organization {} is expired or missing an expiry, refreshing",
+                    ownerAndRepo[0]);
+            String jws = generateJWT(vcs.getClientId(), vcs.getPrivateKey());
+            GitHubAppInstallationToken installationToken = fetchGitHubAppInstallationToken(
+                    gitHubAppToken.getInstallationId(), vcs.getApiUrl(), jws, ownerAndRepo[0]);
+            gitHubAppToken.setToken(installationToken.token());
+            gitHubAppToken.setExpiresAt(installationToken.expiresAt());
+            gitHubAppToken = gitHubAppTokenRepository.save(gitHubAppToken);
         }
 
         log.info("Token fetched for user/organization {}", ownerAndRepo[0]);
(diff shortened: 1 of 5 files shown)


alfespa17:
One of the issues with Terrakube is that the vcs token is refresh by a schedule job every X minutes, I did some code in the past to update the value dynamically but I never finish  the code because I didnt find a way to refresh the token dynamically when someone is using a vcs token to download a module 

https://github.com/terrakube-io/terrakube/pull/2962

With the above code the token expiration was evaluated before running a workspace if I remember correctly

jakegroves:
> One of the issues with Terrakube is that the vcs token is refresh by a schedule job every X minutes, I did some code in the past to update the value dynamically but I never finish the code because I didnt find a way to refresh the token dynamically when someone is using a vcs token to download a module
> 
> #2962
> 
> With the above code the token expiration was evaluated before running a workspace if I remember correctly

thanks  for flagging #2962, that context helps,

to be clear on scope: this PR fixes #3318 and #3317, both in `GitHubTokenService` on the api side.

- #3318 - `getGitHubAppToken` returned the cached token with no expiry check, only re-fetching when the row didn't exist yet. Any gap in the scheduled refresh job (api downtime, a refresh that throws, etc) left a stale token served indefinitely, no self-healing on read.
- #3317 - `refreshAccessToken` NPE'd if the VCS behind a cached token got deleted/recreated (`vcsRepository.findFirstByClientId` returning null), so the row was stuck forever with no way to recover.

Fix adds `expires_at` on `github_app_token`, and `getGitHubAppToken` now treats a missing/near-expiry token as a miss and refreshes it synchronously instead of trusting the background job. `refreshAccessToken` cleans up orphaned rows + their quartz trigger instead of throwing.

That check-before-use logic runs on:
- workspace runs, through `ExecutorService` -> `TokenService` -> `GitHubTokenService.getGitHubAppToken`
- the module refresh job (`ModuleRefreshJob`), same `TokenService` path

So for those two, it's basically the same result you were after in #2962, just via an expiry timestamp instead of an ls-remote check.

one thing this doesn't touch, and I think it's the same gap you're describing from #2962: the module download path through the registry service. Traced it through - `ModuleServiceImpl.getAccessToken` never goes near `GitHubTokenService` at all:

```java
// registry/src/main/java/io/terrakube/registry/service/module/ModuleServiceImpl.java
List<GitHubAppToken> gitHubAppTokens =
    terrakubeClient.getGitHubAppTokenByVcsIdAndOwner(owner, vcsClientId).getData();
```

`terrakubeClient` is a generic JSON:API client hitting the api's Elide endpoint directly, so it reads whatever's sitting in the `token` column with no validation at all. Which lines up with what you said - refreshing dynamically when a module is being downloaded is the part that was never finished.

It seems a bit more involved than the workspace-run case because of two things: Elide (7.1.17) only has lifecycle hooks for `CREATE/UPDATE/DELETE`, no `READ`, so there's no way to hook the registry's `GET` and force a refresh the way we can elsewhere. And the registry talks to the api through `io.terrakube.client:terrakube-spring-boot-starter`, which is its own published artifact, so a real fix probably means a new endpoint in api plus a change on that client and a version bump in registry.

So ... given that I'd treat it as its own piece of work rather than doing the fix of that into this one. Happy to have a look at one also once this one is in

thoughts?

alfespa17 (approved):

alfespa17:
> And the registry talks to the api through io.terrakube.client:terrakube-spring-boot-starter, which is its own published artifact, so a real fix probably means a new endpoint in api plus a change on that client and a version bump in registry.

You are correct, the API will need to expose some endpoint to refresh the token and the io.terrakube.client:terrakube-spring-boot-starter will have to read the expiration date to trigger the refresh.

One thing to consider is the registry is not always calling the API there is some cache that improves registry performance for some operations like the following.

https://github.com/terrakube-io/terrakube/blob/7574faee35f0bdd1651b14a96ff7cad77a0e8f90/registry/src/main/java/io/terrakube/registry/service/module/ModuleServiceImpl.java#L62

https://github.com/terrakube-io/terrakube/blob/7574faee35f0bdd1651b14a96ff7cad77a0e8f90/registry/src/main/java/io/terrakube/registry/service/module/ModuleServiceImpl.java#L88

PR #2956, opened 2026-02-12
refactor(registry): redesign UI with Leaa-inspired cards and add provider auto-refresh

p2pdkivenko:
### ✨ Key Features

**Frontend Improvements**
- 🎯 Centered layout with max-width 1200px for all registry pages
- 🎴 HCP-style single-column card list for modules and providers
- 🎨 Leaa-inspired styling (soft shadows, 8px border-radius, clean typography)
- ⚙️ Inline version selector and sidebar usage cards
- 🔍 Public Registry Search page for browsing and importing from registry.terraform.io
- 💨 Lazy tab loading & pagination (5 items per page)

**Backend Improvements**
- 🔄 Provider auto-refresh via Quartz scheduler (every 24 hours)
- 📦 Queries `registry.terraform.io/v1/providers/{namespace}/{name}/versions`
- 🚀 Parallel platform downloads using Promise.allSettled
- 📊 Sparse fieldsets to reduce API payload by ~98%

### 🔧 Bug Fixes

**Provider Import Naming** (Latest)
- ✅ Fixed provider source format from invalid `registry.host/org/hashicorp/azurerm` to valid `registry.host/org/azurerm`
- Added `imported` boolean flag for tracking providers imported from public registry
- Added `registryNamespace` field to store original namespace separately
- Provider name now stores only the short name (e.g. `azurerm`)
- Updated all backend logic to use new fields instead of parsing name with `/`

**Previous Fixes**
- ✅ Extended provider name column from 32 to support full names during import
- ✅ Changed refresh interval from 5 minutes to 24 hours to reduce API load
- ✅ Restored `preLaunchTask: yarn-install` in VS Code launch configuration

### 📋 Scope

- **Files Modified**: Backend (4 files) + Frontend (5 files)
- **DB Migration**: Added changelog-2.30.0-provider-imported.xml
- **New Features**: Public Registry Search, Provider Auto-Refresh
- **Breaking Changes**: None (backward compatible with new DB fields)

### ✅ Testing Checklist

- [ ] Import provider from public registry - verify Terraform source format is correct
- [ ] Check provider details page - verify namespace displays correctly
- [ ] Test provider refresh job scheduling
- [ ] Verify existing providers still work (manually created ones)
- [ ] Database migration runs successfully

---

**Design References**: Inspired by [Leaa](https://github.com/SolidZORO/leaa) and HCP Terraform UI patterns

PR #3339, opened 2026-07-25
feat(ui): modernize organization and workspace list views

jakegroves:
## Description

Modernizes the Organizations and Workspaces list views in the UI. Adds a condensed, more scannable "New" layout for both lists behind a persisted **New/Legacy** toggle, so the existing views stay available while the new ones bake in. Also fixes a couple of unrelated bugs surfaced while building this (OpenTofu logo, organization-switch navigation).

This helps with the TACOS approach of a collaboration platform but becomes difficult when platforms have many/multiple orgs, and hundreds of workspaces, as the current card approach is quite large.

## What's changed

**Organizations list ("New" view)**
- Condensed, sortable table (name, execution mode) instead of the card grid
    - in 'Group by project' mode it will show 15 workspaces per project, and a toggle appears to show more or show less 
- Workspace count badge per organization
- Permission-gated settings shortcut per row (only shown if you can manage that org)
- Search by name/description

**Workspaces list ("New" view)**
- Custom-built condensed list (replacing antd's `Table` internals) — single header, real column widths, consistent theming
- Click-to-filter: click a tag or project directly on a row to filter by it; removable "Filtering by tag: X ×" chips in the filter bar
- Project quick-filter chips with a search box (scales to orgs with many projects) alongside the existing status filter
- **Group by project** toggle — rows cluster under project headers, with per-group pagination (a "Show N more" expand) so large orgs don't render hundreds of rows at once
- Description + tags shown on a second line under the workspace name (tags capped at 3 + a "+N" tooltip, monochrome pills with a small deterministic colored dot)
- Lock icon for locked workspaces
- Sortable column headers (Name/Status/Last run/Version/Source) now share the same sort state as the existing "Sort by" dropdown, instead of running independently

**New/Legacy toggle**
- Segmented control in the page header of both list pages, persisted in `localStorage`, defaults to "New"
- Legacy view is functionally and visually unchanged

**Bug fixes (unrelated to the above, found along the way)**
- OpenTofu logo wasn't rendering — `IacTypeLogo` checked for `"opentofu"` but the real `iacType` value used elsewhere in the codebase is `"tofu"`
- Switching organizations always redirected to Workspaces regardless of which page you were on, and didn't refresh data if you were already on the Workspaces page — now it stays on the same section (workspaces/registry/settings/projects) and refetches correctly
- workspace search was not doing search as you type, now does that.

**Accessibility**
- The new custom list's clickable rows, sortable headers, and expand controls are keyboard-operable (Enter/Space) with visible focus outlines

## Tests

- 24 test suites / 195 tests passing (`yarn jest`)
- `tsc --noEmit` clean
- New/updated coverage: `WorkspaceTable`, `WorkspaceFilter`, `WorkspaceGroupedList` logic (now folded into `WorkspaceTable`), `OrganizationTable`, `OrgSettingsButton`, `useWorkspaceFilterState`, `filterWorkspaces`, `workspaceSort` comparators, `organizationService`/`workspaceService` GraphQL mapping, `ListViewToggle`, `IacTypeLogo`
- Manual verification in a running dev stack (toggle switching, grouping, filtering, sorting, pagination, org-switch navigation)

## Screenshots

**Organizations list**
legacy: terrakube-org-legacy
new: terrakube-org-new

**Workspaces list — flat**
legacy: terrakube-workspaces-legacy-1
new: terrakube-workspaces-new-1

**Workspaces list — grouped by project**
terrakube-workspaces-new-group-by-project-1
terrakube-workspaces-new-group-by-project-2


jakegroves:
created as a draft, as I have some questions:
- naming of: legacy vs new ... should this be something like 'cards vs compact' ?
- is this wanted? personally coming from HCP/TFE the HCP/TFE UI is alot easier to navigate than TK right now as the cards are large
    - goes against the styling of stuff that was done like https://github.com/terrakube-io/terrakube/pull/2956 
- anything that should change?

alfespa17:
- naming of: legacy vs new ... should this be something like 'cards vs compact' ?

I like "cards vs compact"

- is this wanted? personally coming from HCP/TFE the HCP/TFE UI is alot easier to navigate than TK right now as the cards are large

I like the idea; it will look good for people with hundreds of workspaces 

- anything that should change?

1) when you are reading information for the UI always try to use the graphql API
2) you can include "description" column in the organization
3) maybe you can check how this looks when you are using the "purple" and "dark" theme

alfespa17:
By the feature to group workspaces by projects looks great

alfespa17:
Will you add more changes to this branch? Or can we move it for "ready for review"?

jakegroves:
> Will you add more changes to this branch? Or can we move it for "ready for review"?

made some adjustments to the discussion on slack:
- default color now the terrakube (purple)
- made workspaces poll their status on workspace list
- moved sidemenu for better workspace/org listing capability
    - collapsible side menu
    - nested settings menu for org and workspaces rather than having multi settings menu on one page
- improved org/workspace listing
- made form pages more consistent

terrakube-ui-1
terrakube-ui-2
terrakube-ui-4
terrakube-ui-5

alfespa17:
I was testing this in my dev environment and it looks good and I like the way yo manage to handle "duplicated" org and workspaces names with the hooks.

There is one little error when you open the registry, the version is showing "vv", we can fix this later in another PR

imagen

You added a lot of test cases for the UI, the pipeline is not really running those right now it is only running the build, in the future we could change this too.

https://github.com/terrakube-io/terrakube/blob/main/.github/workflows/pull_request_ui.yml

PR #3341, opened 2026-07-27
feat: run onReject commands when an approval step is rejected

### LINKED ISSUE #3340: Support `onReject` commands on approval steps (run cleanup/notifications when a run is rejected)
# Feature description 💡

## Problem

Templates can react to almost every outcome of a run except rejection:

- `commands` on an `approval` step run after the run is **approved** (the approval flow type is executed like `customScripts` in `ExecutorJobImpl`)
- `onFailure` runs when a terraform/customScripts step **fails** in the executor

But when a user **rejects** a run at the approval step, nothing user-defined can execute. `ScheduleJob`'s `rejected` case only does teardown (marks remaining steps failed, updates the VCS commit status, deschedules). The job never reaches the executor, so `onFailure` never fires either.

This makes it impossible to notify a Slack channel / ticketing system, clean up temporary resources, or record an audit event when an approver rejects a plan — common needs for teams using approval gates.

## Proposal

Add an optional `onReject` command list to the `approval` flow step, symmetric with the existing `onFailure`:

```yaml
flow:
  - type: "terraformPlan"
    step: 100
  - type: "approval"
    step: 150
    team: "TERRAKUBE_ADMIN"
    onReject:
      - runtime: "BASH"
        priority: 100
        after: true
        script: |
          curl -X POST $SLACK_WEBHOOK -d "{\"text\": \"Run rejected for $workspaceName\"}"
  - type: "terraformApply"
    step: 200
```

### Proposed semantics

- Fire-and-forget: the hook's success/failure never changes the job's final `rejected` status
- The job **stays** `rejected` throughout — the hook must not resurrect the run
- Script output is attached to the step so it is visible in the UI, like any other command output
- No impact on `autoApply` jobs (they never enter `waitingApproval`) or on existing templates (field is optional)

## Implementation sketch

I traced the code and believe this is a small change (happy to submit the PR):

1. **`Flow.java`** (api): add `List<Command> onReject;` — SnakeYAML binds it automatically
2. **`ScheduleJob.java`** (api): in `case rejected:`, before teardown, resolve the current flow via `tclService.getNextFlow(job)` (the approval step is still the first pending step at that point, since rejection is only reachable from `waitingApproval`), set `flow.setCommands(flow.getOnReject())`, and dispatch via `executorService.execute(job, stepId, flow)` — reusing the exact path the approval type already uses on approve. Do **not** set the job to `queue`
3. **`UpdateJobStatusImpl.java`** (executor): extend the existing cancelled-job guard (`isJobCancelled`) to also cover `rejected`, in both `setRunningStatus` and `setCompletedStatus`. Without this, the executor callback would PATCH the rejected job back to `running`/`pending`, and with all steps torn down the scheduler would then mark a rejected run `completed` (including a success commit status to the VCS). Step-level log updates keep working so the hook output is still visible

Plus tests in `TclServiceTest`, `ScheduleJobTest`, and the executor's status-update tests, and a docs PR for the templates page.

## Open questions

1. Naming: `onReject` (mirrors `onFailure`) vs `onRejection`?
2. Should the hook output land on the approval step (simplest), or should a dedicated "On Reject" step record be created for cleaner UI display?
3. Note for the changelog: templates using `onReject` on an older server will fail YAML parsing into the `yamlError` flow, so this needs a minimum-server-version callout in the docs

### Anything else?

I'd like to contribute this myself — if maintainers agree with the direction (and settle the naming/step-display questions above), I'll open a PR against `main` covering the api + executor changes and tests.

adeeelsameer:
Closes #3340

## Problem

Templates can react to almost every outcome of a run except rejection: `commands` on an `approval` step run after the run is approved, and `onFailure` runs when a step fails in the executor. But when a user rejects a run at the approval step, the job is torn down without ever reaching the executor, so nothing user-defined can execute — no Slack/ticket notification, no cleanup, no audit event.

## What this PR does

Adds an optional `onReject` command list to the `approval` flow step, symmetric with the existing `onFailure`:

```yaml
flow:
  - type: "terraformPlan"
    step: 100
  - type: "approval"
    step: 150
    team: "TERRAKUBE_ADMIN"
    onReject:
      - runtime: "BASH"
        priority: 100
        after: true
        script: |
          curl -X POST $SLACK_WEBHOOK -d "{\"text\": \"Run rejected for $workspaceName\"}"
  - type: "terraformApply"
    step: 200
```

Semantics: fire-and-forget. The job stays `rejected` throughout; the hook's success or failure never changes the final job status. Script output is attached to the approval step so it is visible in the UI.

## How

- **api / `Flow.java`**: new optional `onReject` command list (SnakeYAML binds it automatically; existing templates unaffected)
- **api / `ScheduleJob.java`**: when a job lands in `rejected`, and the current flow step is an `approval` with `onReject` commands, dispatch them to the executor through the same path the approval type already uses on approve — but without moving the job to `queue`, so it stays `rejected`. Dispatch failures are logged and never block the normal teardown
- **executor / `UpdateJobStatusImpl.java`**: extend the existing cancelled-job guard to also cover `rejected`, in both `setRunningStatus` and `setCompletedStatus`. Without this the executor callback would PATCH the rejected job back to `running`/`pending`, and with all steps already torn down the scheduler would then mark a rejected run `completed` (including a success commit status to the VCS). Step-level log updates still run so the hook output is visible

## Testing

- `TclServiceTest`: `onReject` parses from template YAML into the flow model
- `ScheduleJobTest`: rejected job dispatches its `onReject` commands to the executor with the job left in `rejected` and teardown still performed; no executor call when `onReject` is absent; teardown survives a dispatch failure
- `UpdateJobStatusImplTest` (new): executor skips job status mutation for `rejected` (and `cancelled`) jobs in both callbacks while still updating step logs, and continues to update active jobs normally

## Notes

- Open questions from #3340 (naming `onReject` vs `onRejection`, whether hook output should get a dedicated step instead of the approval step) — happy to adjust based on maintainer preference
- Templates using `onReject` on an older server will fail YAML parsing into the `yamlError` flow, so this needs a minimum-server-version callout in the docs; I'll open a docs PR for the templates page once the field name is settled


alfespa17:
There is one small issue

When using the following

imagen

The status for the approval step is "green"

imagen

If you compare with an approval without "onReject", it should be marked with "red" and rejected like this

imagen

adeeelsameer:
Good catch, thank you for testing it! Fixed in the latest commit.

**Root cause:** teardown marks the approval step `failed` (that's the red ✗ you see without `onReject`), but the hook run then reports its own result against that same step — so a successful script repainted the step to `completed`.

**Fix:** in the executor's completion callback, when the job is `rejected` the step is now always written back as `failed` — the `onReject` command output is still saved to the step logs, but the script's success can no longer change the step status. Cancelled jobs keep their existing behavior (step status still follows the actual command result).

Added a unit test asserting the step is written back as `failed` even when the onReject command succeeds (and a companion test locking the unchanged cancelled behavior). I don't have a full environment handy to re-run your exact template end to end, so if you get a chance to re-test the flow from your screenshot, that would be appreciated.

alfespa17:
You can simply use codespaces to test, you can have an ephemeral environment really quickly.

https://docs.terrakube.io/getting-started/development/github-codespaces

By the way your PR has merge conflicts

adeeelsameer:
Fixed merge conflicts. I tested it and it looks good now.

When using the following template:

image


I get this:

image

adeeelsameer:
@alfespa17 any updates?

alfespa17:
I have been quite busy I will check later this week

alfespa17 (approved):

alfespa17:
I merged this one to handle on reject commands but for a simple webhook message to slack you can use [this](https://github.com/terrakube-io/terrakube/pull/3401) in version 2.33.0 that will be created in 2 weeks

PR #3344, opened 2026-07-28
fix(webhook): serialize shared-webhook sync via a Quartz job, fix hook registration, orphan cleanup, and UI delete

jakegroves:
## Problem

Concurrent workspace webhook create/update/delete requests sharing a repository URL raced on `RepoWebhookService.getOrCreateRepoWebhook` when setting up the shared repo-level webhook. A first fix using a Postgres advisory lock (`pg_advisory_xact_lock`) worked but wasn't portable — Terrakube also supports SQL Azure and other datasources where that function doesn't exist. So @alfespa17 reccomended looking at Quartz replacement instead.

While building and live-testing the Quartz-based replacement, three further bugs surfaced:

- The shared-webhook sync job never fired for webhook create/update at all — only delete — so brand new or updated webhooks never actually synced their GitHub-side hook, despite every existing unit test passing.
- Reverting the last workspace still using the shared webhook on a given repo left the shared `RepoWebhook` (both the DB row and the live GitHub webhook) permanently orphaned.
- The UI's "disable webhook" and "delete event" actions were both broken: DELETE requests failed against the API, and the UI silently reported success regardless of the actual result.

## How it was reproduced

- **Race condition**: unit tests plus a Testcontainers-backed integration test (`RepoWebhookSyncCoalescingIntegrationTest`) firing N concurrent `scheduleSync()` calls for the same repo URL against a real Postgres instance and a real JDBC-backed Quartz scheduler — this is what caught Quartz's own `scheduleJob()` pre-check not being atomic under concurrent transactions (the losing caller got a raw `JobPersistenceException` instead of the expected `ObjectAlreadyExistsException`).
- **Missing sync on create/update**: live-tested end to end against a running devcontainer instance — created workspaces sharing a repo via Terraform, watched the API logs and queried the database directly, and confirmed no `RepoWebhook` row and no GitHub webhook ever got created despite the Terraform apply succeeding.
- **Orphaned webhook on revert**: reverted the last migrated-v2 workspace on a shared repo via the UI and confirmed via direct database queries and GitHub API calls that the shared webhook was never cleaned up.
- **UI delete/disable bugs**: reproduced directly against the API with `curl`, isolating the exact request shape (`Content-Type: application/vnd.api+json` on a DELETE) that Elide's Spring routing misinterprets, then confirmed the same failure via the actual UI.

## How it was solved

- Replaced the advisory-lock approach with `RepoWebhookSyncScheduler` + `RepoWebhookSyncJob`: a single-flight Quartz job keyed deterministically by a hash of the normalized repo URL. Quartz's own cluster-safe JDBC job store coalesces concurrent triggers for the same repo onto one execution — nothing here reimplements locking.
- `WebhookManageHook` now schedules the sync from `POSTCOMMIT` (after the workspace's migrated `Webhook`/`WebhookEvent` data is durably committed) instead of calling `RepoWebhookService` synchronously from `PRECOMMIT`.
- Added a `JobPersistenceException` catch alongside the existing `ObjectAlreadyExistsException` one, since both represent the same "someone else already scheduled this" outcome under real concurrent load.
- Registered the missing `@LifeCycleHookBinding` for `CREATE+POSTCOMMIT` and `UPDATE+POSTCOMMIT` on `Webhook` — previously only `DELETE+POSTCOMMIT` was registered, so the sync job was silently unreachable for create/update.
- Changed the sync-scheduling condition to fire on every GitHub-backed create/update, not only when still `migratedV2`, so a revert now correctly triggers a resync that notices the workspace no longer shares the repo and cleans up the shared webhook if it was the last one.
- Fixed the UI's webhook and webhook-event DELETE calls to explicitly drop the `Content-Type` header, avoiding Elide's relationship-delete routing ambiguity.
- Replaced the disable-webhook flow's fire-and-forget delete (fake immediate success message, no error handling) with a proper `await`/`catch` so real failures are surfaced.

## Tests added

- `RepoWebhookSyncJobTest`, `RepoWebhookSyncSchedulerTest` — unit coverage for the scheduler's coalescing behavior and the job's orphan-cleanup vs. sync branch logic.
- `RepoWebhookSyncCoalescingIntegrationTest` — real Postgres + real Quartz JDBC job store, proving N concurrent `scheduleSync()` calls for the same URL result in exactly one execution and no propagated exceptions. Replaces the deleted `RepoWebhookConcurrencyTests`, which tested the now-abandoned advisory-lock layer directly.
- `WebhookLifeCycleHookBindingTest` — asserts the actual `@LifeCycleHookBinding` annotations on `Webhook`, since `WebhookManageHookTest` calls `execute()` directly and can't catch a missing registration.
- `WebhookManageHookTest` — new cases for revert-still-schedules-sync and non-GitHub-skips-sync.
- All fixes verified with the standard revert-fix-confirm-fails / restore-confirm-passes pattern, and live-verified end to end against a running devcontainer (Terraform apply/destroy/taint, direct API calls, and manual UI testing).

## Companion PR

Goes along with [terraform-provider-terrakube#153](https://github.com/terrakube-io/terraform-provider-terrakube/pull/153) — the provider had matching bugs on the `terrakube_workspace_webhook_v2` / `terrakube_workspace_webhook_event` resources (unknown-after-apply attributes, a malformed `Update()` payload, and drift/replace handling) that only fully surfaced once this backend redesign was in place.


alfespa17:
@jakegroves the build is failing

alfespa17 (approved):

PR #3353, opened 2026-07-31
feat(executor): support running multiple executor replicas, fix VCS-status crash & plan-only completion, add Queued filter

jakegroves:
### What was added and why

Terrakube's executor only processes one job at a time internally, but nothing stopped a busy pod from still receiving new job dispatches when `executor.replicaCount > 1` — the Service just round-robins across all pods regardless of load. 
Alternatively people were instructed to have ephemeral agents, but people may want to have a pool of 'warm ready' agents, and this could ultimately be extended in the future to do autoscaling with KEDA in future improvements

This adds readiness gating so a busy pod drops out of rotation, making it safe to actually run multiple executor replicas concurrently. Also adds a "Queued" filter to the workspace list, since `JobStatus.Queue` already existed in the data model but wasn't exposed anywhere in the UI.

### How it was done

- `ExecutorJobImpl.createJob()` publishes `REFUSING_TRAFFIC` for the duration of a job and `ACCEPTING_TRAFFIC` once it's free, so the k8s readiness probe pulls a busy pod out of the Service's endpoints. Ephemeral executors skip the restore step since the pod shuts down right after.
- `JobExecutionWatchdog` is a scheduled backstop (60s interval): if a job never reaches the `finally` block (hung terraform process, looping hook script), it marks the pod's liveness `BROKEN` after `io.terrakube.executor.job.maxDurationMinutes` (default 360) so k8s restarts it instead of the pool silently shrinking forever.
- UI: added `JobStatus.Queue` to the shared status palette, color map, icon switch, and tag label — the same places every other status is already wired through.

### Any tests added

- `UpdateJobStatusImplTest.java` (new): asserts a plan-only job (single step) with detected changes now reaches `completed`; asserts a Plan+Apply job's plan step still correctly stays `pending`.
- `JobExecutionWatchdogTest.java` / `ExecutorJobImplTest.java`: cover the readiness-toggling and watchdog-scheduling behavior.
- No new test needed for the `ScheduleJob` fix — `ScheduleJobTest.expiredJobsAreDescheduledEvenIfVcsIntegrationFails` already covers "VCS status call throws" for a related code path; confirmed it still passes.

### What was fixed/added along the way and why

- **`ScheduleJob.updateJobStatusOnVcs`**: a failure notifying the VCS of a job's status (expired token, provider outage, or a job with no resolvable commit SHA) was an unhandled exception that crashed the *entire* Quartz job execution — including from the job-completion path, so a job that had actually finished successfully could get stuck instead of ever reaching a terminal status. Wrapped in try/catch so it's logged, not fatal.
- **`UpdateJobStatusImpl`**: a plan-only job whose plan found changes was left `pending` forever — correct when a following apply step is waiting for approval, wrong when there's no further step in the template. Now completes when this is the job's only step.
- Found but **not** fixed here, flagged for follow-up: a race condition creating duplicate `GitHubAppToken` rows under concurrent dispatch to the same VCS connection (worked around via manual DB cleanup during testing, no code fix yet), and a related gap where a job whose step actually *failed* could still get marked `completed` instead of `failed` (`ScheduleJob.executePendingJob`'s completion branch doesn't check prior step outcomes before calling `completeJob()`).

This is paired with https://github.com/terrakube-io/terrakube-helm-chart/pull/288

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
@@ -494,19 +494,26 @@ public class ScheduleJob implements org.quartz.Job {
             return;
         }
 
-        switch (job.getWorkspace().getVcs().getVcsType()) {
-            case GITHUB:
-                gitHubWebhookService.sendCommitStatus(job, jobStatus);
-                break;
-            case GITLAB:
-                gitLabWebhookService.sendCommitStatus(job, jobStatus);
-                break;
-            case AZURE_DEVOPS:
-            case AZURE_SP_MI:
-                azDevOpsWebhookService.sendCommitStatus(job, jobStatus);
-                break;
-            default:
-                break;
+        // Notifying the VCS is a side effect of the job, not part of it: a VCS-side failure
+        // (expired token, provider outage, rate limit, missing commit) must never abort the
+        // job itself, since callers here include the job-completion path.
+        try {
+            switch (job.getWorkspace().getVcs().getVcsType()) {
+                case GITHUB:
+                    gitHubWebhookService.sendCommitStatus(job, jobStatus);
+                    break;
+                case GITLAB:
+                    gitLabWebhookService.sendCommitStatus(job, jobStatus);
+                    break;
+                case AZURE_DEVOPS:
+                case AZURE_SP_MI:
+                    azDevOpsWebhookService.sendCommitStatus(job, jobStatus);
+                    break;
+                default:
+                    break;
+            }
+        } catch (Exception e) {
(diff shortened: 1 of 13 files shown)


alfespa17:
@jakegroves I think this part looks good but I think it will require some changes in the API too, for example if we have 10 executor replicas that are running some jobs the readiness probe for all will be "false" and if the API send a new request here

https://github.com/terrakube-io/terrakube/blob/95d7529e26109af1be5dbb4ab963dffe375c8a93/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/persistent/PersistentExecutorService.java#L84

The job will break instead of waiting for a "free" executor with "service not available"  🤔 I guess

PR #3372, opened 2026-08-05
fix(redis): enable Jedis connection pooling

### LINKED ISSUE #3363: Redis log writes open a new connection per line (Jedis pooling not enabled)
### Bug description 🐞

**Versions:** Terrakube 2.32.0, Helm chart 4.7.4, ephemeral executors, external Redis (ElastiCache Valkey 8.0.1) with TLS.

# Problem

Log shipping writes one `XADD` per terraform output line, synchronously, on the thread draining terraform's stdout:

- [`LogsConsumer.accept()`](https://github.com/AzBuilder/terrakube/blob/main/executor/src/main/java/io/terrakube/executor/service/logs/LogsConsumer.java) — called once per line, calls `sendLogs()` inline
- [`LogsServiceRedis.sendLogs()`](https://github.com/AzBuilder/terrakube/blob/main/executor/src/main/java/io/terrakube/executor/service/logs/LogsServiceRedis.java) — `redisTemplate.opsForStream().add(...)`

Because the connection factory is not pooled, each of those writes appears to open and close its own connection:

- [`RedisAutoConfiguration`](https://github.com/AzBuilder/terrakube/blob/main/executor/src/main/java/io/terrakube/executor/configuration/RedisAutoConfiguration.java) builds `JedisClientConfiguration.builder()...build()` without `usePooling()`
- [`StreamingConfiguration`](https://github.com/AzBuilder/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/streaming/StreamingConfiguration.java) (API) does the same, so the `log-via-api` path has it too — [`LogsService.appendLogs()`](https://github.com/AzBuilder/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/logs/LogsService.java) loops per line
- `commons-pool2` is not declared in `pom.xml`, `executor/pom.xml`, or `api/pom.xml`

With TLS enabled, every line therefore pays a full TCP + TLS handshake. Since the writes are serial, the drain rate is capped at roughly `1 / connection-setup-time`.

# Evidence

Measured on a single plan (~40,000 output lines):

- **~19 new Redis connections/sec sustained**, matching the observed **~17.5 log lines/sec** — 1:1 with lines
- CloudWatch `NewConnections` ~17,000 per 15 min while `CurrConnections` stayed at **4–7** (created and closed immediately, not held)
- **~1,100–2,500 sockets in `TIME_WAIT` to :6379** inside one job pod
- The terraform process sat in **`pipe_write`** (`/proc/<pid>/task/*/wchan`) — blocked writing stdout, waiting on the log consumer

Net effect: the plan took **~40 minutes instead of ~7**, because terraform is rate-limited by log shipping rather than by terraform itself. Long runs then exceed credential lifetimes — for us, EKS tokens (15 min hard cap) expired mid-plan, producing ~900 `Unauthorized` errors.

We reduced connection-setup latency in our environment and the same plan went to **87 lines/sec** (terraform CPU 51% → 165%), which supports connection setup being the constraint rather than Redis throughput or terraform itself.

### Possible direction

Enabling pooling (`usePooling()` on the client configuration, which would need `commons-pool2` on the classpath) would let a job reuse one connection instead of one per line. We may be missing context on why it was left off, though — and other approaches could fit better, e.g. pipelining/batching the `XADD` calls, holding a connection for the duration of a step, or making `sendLogs` non-blocking. Happy to test a patch if that's useful.

### Steps to reproduce

Observe new connection metrics on redis cluster, perform large plan.

### Expected behavior

Connections are re-used. I believe there is an option we can configure that would sync to redis every ~5m but that does not provide the same level of durability/debugability.

### Example repository

_No response_

### Anything else?

_No response_

Guflly:
Enable Jedis pooling for the API and executor Redis connection factories and add the required pool dependency.

Fixes #3363

Tests:
- mvn -pl executor -Dtest=RedisAutoConfigurationTest,RedisPropertiesTest,SpringAsyncAutoConfigurationTest test
- mvn -pl api -Dtest=StreamingConfigurationTest,StreamingPropertiesTest,StreamingServiceTest,RedisStreamReaderTest test

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -139,6 +139,10 @@
             <type>jar</type>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-pool2</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.yahoo.elide</groupId>
             <artifactId>elide-spring-boot-starter</artifactId>
             <version>${elide.version}</version>
--- a/api/src/main/java/io/terrakube/api/plugin/streaming/StreamingConfiguration.java
+++ b/api/src/main/java/io/terrakube/api/plugin/streaming/StreamingConfiguration.java
@@ -57,6 +57,7 @@ public class StreamingConfiguration {
     @Bean
     JedisConnectionFactory jedisConnectionFactory(StreamingProperties props, SSLSocketFactory sslSocketFactory) {
         JedisClientConfiguration.JedisClientConfigurationBuilder clientConfigBuilder = JedisClientConfiguration.builder();
+        clientConfigBuilder.usePooling();
 
         if (props.isSsl()) {
             log.info("Setup Redis connection using SSL");
--- a/api/src/test/java/io/terrakube/api/plugin/streaming/StreamingConfigurationTest.java
+++ b/api/src/test/java/io/terrakube/api/plugin/streaming/StreamingConfigurationTest.java
@@ -1,0 +1,25 @@
+package io.terrakube.api.plugin.streaming;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import javax.net.ssl.SSLContext;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.data.redis.connection.jedis.JedisConnectionFactory;
+
+class StreamingConfigurationTest {
+
+    @Test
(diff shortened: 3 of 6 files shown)


alfespa17:
thank you for your help @Guflly

PR #3375, opened 2026-08-05
fix(executor): restrict the shallow clone to the requested ref

siler:
currently, even with `setDepth(1)`, jgit defaults to fetching all branches. tested when fetching against `torvalds/linux` I see the following results:

| Clone | Wall | `.git` |
| --- | --- | --- |
| full, all branches | 22.9 min | 6.77 GB |
| `depth=1`, all branches (today) | 9.1 min | 3.32 GB |
| `depth=1`, one branch (this PR) | **50 s** | **306 MB** |

tests run from a macbook pro (m4)

# changes in this pr
* fetch just one ref (using `setBranchesToClone()`)
* if provided with a tag, triggers an `ls-remote` round trip to resolve it to a branch or a tag
  * this is best effort - a failed probe assumes it's a branch, and if that fails the clone reports it
* credentials are resolved only once per job. without some refactoring that would have been duplicated
* the ssh session factory for fetch is closed, the clone factories are left open
  * fixing the clone factory close would ideally involve changing where the key is placed as `CloneCommand` only works on an empty directory

# risk
the checkout now only has a single branch and no tags. if anything (git hooks, etc.) depends on more being present it will be broken.

# tests
added `org.eclipse.jgit.junit.ssh` at matching jgit version for offline ssh tests. wrote tests for various failure cases this change could introduce as well as some pre-existing behavior. if you don't want the new test dep, we can remove it and associated tests.

i did not add tests for http paths (github, etc.) or submodules

### DIFF
--- a/executor/pom.xml
+++ b/executor/pom.xml
@@ -141,6 +141,12 @@
 			<version>${jgit.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>org.eclipse.jgit</groupId>
+			<artifactId>org.eclipse.jgit.junit.ssh</artifactId>
+			<version>${jgit.version}</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.apache.groovy</groupId>
 			<artifactId>groovy-all</artifactId>
 			<version>${groovy.version}</version>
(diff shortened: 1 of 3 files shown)


jakegroves:
you should be able test with https://docs.terrakube.io/getting-started/development/dev-container or https://docs.terrakube.io/getting-started/development/github-codespaces

alfespa17 (approved):

PR #3379, opened 2026-08-07
fix(api): evaluate federated membership outside the groups loop

### LINKED ISSUE #3378: Federated identity can read but cannot create jobs: isServiceMember nests the federated check inside the groups loop
### What happens

A federated identity (`federated` table + claim matchers) authenticates fine and can **read** workspaces, templates and jobs, but gets denied when creating a job.

### Root cause

`DexGroupServiceImpl.isServiceMember` evaluates the federated check **inside** the loop over the `groups` claim:

```java
public boolean isServiceMember(User user, String group) {
    JwtAuthenticationToken principal = ((JwtAuthenticationToken) user.getPrincipal());
    boolean isMember = principal.getTokenAttributes().get("iss").equals("TerrakubeInternal") ? true : false;
    boolean isFederated = isFederatedAccount(user);
    if (!isMember) {
        for (String groupName : toStringArray((ArrayList) principal.getTokenAttributes().get("groups"))) {
            if (groupName.equals(group))
                isMember = true;
            if (isFederated && isFederatedMember(user, group))   // <-- inside the loop
                isMember = true;
        }
    }
    return isMember;
}
```

Tokens from an external OIDC provider generally have **no `groups` claim** — a GitHub Actions OIDC token certainly doesn't. `toStringArray(null)` returns an empty array, the loop body never executes, and the federated check never runs, so membership resolves to `false`.

Read paths work because `MembershipService.checkMembership` calls `isFederatedMember` directly, outside any loop. So the identity looks half-working: it lists workspaces, then fails on `POST /organization/{id}/job` (`team plan job` / `team limited plan job`).

### Suggested fix

Hoist the federated check out of the loop:

```java
if (!isMember && isFederated && isFederatedMember(user, group)) {
    isMember = true;
}
```

`isMember(User, String)` (the non-service path) has the same shape and may deserve the same treatment for consistency.

### Environment

Terrakube 2.32.2, Dex auth, federated identity with issuer `https://token.actions.githubusercontent.com` and claim matchers on `repository` / `repository_owner` / `ref`.

### Workaround in the meantime

The token-minting path handles federation correctly (`TeamTokenService.getCurrentGroups` returns the federated name), so a federated caller *can* mint a short-lived team token via `POST /access-token/v1/teams` and use that token — it carries a real `groups` claim — for the job APIs. That works today, it is just an extra round trip that the fix above would remove.

jvdrag:
Fixes #3378

### Problem

A federated identity can authenticate and read, but gets denied when creating a job.

In `DexGroupServiceImpl.isServiceMember` the federated check sits inside the loop over the `groups` claim:

```java
for (String groupName : toStringArray((ArrayList) principal.getTokenAttributes().get("groups"))) {
    if (groupName.equals(group))
        isMember = true;
    if (isFederated && isFederatedMember(user, group))
        isMember = true;
}
```

Tokens issued by an external provider usually have no `groups` claim, and a GitHub Actions OIDC token never has one. `toStringArray(null)` returns an empty array, the loop body never runs, so the federated check never executes and membership resolves to false.

Read paths work because `MembershipService.checkMembership` calls `isFederatedMember` directly, outside any loop. That is why the identity looks half working: it lists workspaces and templates, then fails on `POST /organization/{id}/job`.

### Fix

Move the federated check out of the loop. The loop keeps behaving the same for tokens that do carry `groups`.

### Test

`DexGroupServiceTests` covers three cases:

- federated token with no `groups` claim and matching claims is a member. This one fails without the fix.
- federated token whose claims do not match the registered ones is not a member.
- token carrying a `groups` claim still resolves membership the same way.

The last two are there on purpose. For an authorization fix, the risk is not that it keeps denying, it is that it starts allowing too much.

Result with the fix, and with the fix reverted:

```
Tests run: 3, Failures: 0   (with fix)
Tests run: 3, Failures: 1   (without fix)
  DexGroupServiceTests.federatedTokenWithoutGroupsClaimIsMember:37 expected: <true> but was: <false>
```

### Notes

`isMember` has a similar shape but no federated branch at all, so it is left unchanged. Federated principals reach `isServiceMember` because `isServiceAccount` already returns true for them.

Reproduced on 2.32.2 with a federated identity using issuer `https://token.actions.githubusercontent.com` and claim matchers on `repository` and `repository_owner`. As a workaround until this lands, a federated caller can mint a short lived team token through `POST /access-token/v1/teams`, since `getCurrentGroups` handles federation correctly, and use that token for the job APIs.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/security/groups/dex/DexGroupServiceImpl.java
+++ b/api/src/main/java/io/terrakube/api/plugin/security/groups/dex/DexGroupServiceImpl.java
@@ -56,10 +56,13 @@ public class DexGroupServiceImpl implements GroupService {
         boolean isMember = principal.getTokenAttributes().get("iss").equals("TerrakubeInternal")? true: false;
         boolean isFederated = isFederatedAccount(user);
         if(!isMember) {
+            // Federated tokens are issued by an external provider and usually carry no
+            // "groups" claim, so this check cannot live inside the loop below.
+            if (isFederated && isFederatedMember(user, group))
+                isMember = true;
+
             for (String groupName : toStringArray((java.util.ArrayList) principal.getTokenAttributes().get("groups"))) {
                 if (groupName.equals(group))
-                    isMember = true;
-                if (isFederated && isFederatedMember(user, group))
                     isMember = true;
             }
             log.debug("{} is member {} {}", principal.getTokenAttributes().get("name"), group, isMember);
--- a/api/src/test/java/io/terrakube/api/DexGroupServiceTests.java
+++ b/api/src/test/java/io/terrakube/api/DexGroupServiceTests.java
@@ -1,0 +1,94 @@
+package io.terrakube.api;
+
+import com.yahoo.elide.core.security.User;
+import io.terrakube.api.plugin.security.groups.dex.DexGroupServiceImpl;
+import io.terrakube.api.repository.FederatedRepository;
+import io.terrakube.api.rs.federated.Federated;
+import io.terrakube.api.rs.federated.claim.FederatedClaim;
+import org.junit.jupiter.api.Test;
+import org.springframework.security.oauth2.jwt.Jwt;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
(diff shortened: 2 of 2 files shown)


alfespa17 (approved):

alfespa17:
thank you for your help @jvdrag

PR #3380, opened 2026-08-07
fix(executor): enable native S3 state locking on the generated backend

jvdrag:
### Problem

The generated `aws_backend_override.tf` sets no locking at all, so nothing stops two writers from working on the same state object at the same time.

Terrakube serializes jobs per workspace, which covers job against job. It does not cover a job running while someone applies the same stack from a laptop or from CI, and that case is easy to hit when the repository backend points at the same key the executor writes to. The second writer simply overwrites.

### Fix

Add `use_lockfile = true` to the generated backend when both conditions hold:

- version is 1.10.0 or newer. Native S3 locking through conditional writes landed in Terraform 1.10 and OpenTofu 1.10, and the argument does not exist before that.
- no custom endpoint is configured. S3 compatible services do not all implement the conditional write this depends on, so MinIO and similar setups keep the current behavior.

The existing version parsing is reused, so version constraints like `~>1.13.0` and X-Range wildcards resolve the same way they already do for the endpoints block.

### Test

Three cases added to `AwsTerraformStateImplTest`:

- 1.10.0 gets `use_lockfile = true`. This one fails without the change.
- 1.9.8 does not get it, since the argument is invalid there.
- a custom endpoint does not get it, even on 1.13.0.

```
Tests run: 27, Failures: 0   (with the change)
Tests run: 27, Failures: 1   (without it)
  testGetBackendStateFile_UsesLockfileFromTerraform110:87 expected: <true> but was: <false>
```

### Notes

Only the AWS state backend is touched. Azure and GCP have their own locking behavior, and the local backend goes through the API instead.

### DIFF
--- a/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java
+++ b/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java
@@ -61,6 +61,8 @@ public class AwsTerraformStateImpl implements TerraformState {
 
     private boolean includeBackendKeys;
 
+    private boolean useLockfile;
+
     @NonNull
     TerraformOutputPathService terraformOutputPathService;
 
@@ -113,13 +115,16 @@ public class AwsTerraformStateImpl implements TerraformState {
               log.warn("No including backend information");
             }
 
-            // Native S3 locking (conditional writes) is available from Terraform 1.10
-            // and OpenTofu 1.10. Without it the generated backend takes no lock at all,
-            // so a job and anything else writing the same state can overwrite each other.
-            // Skipped when a custom endpoint is set: S3 compatible services do not all
-            // support the conditional write this relies on.
-            if(endpoint == null && version.compareTo(new ComparableVersion("1.10.0")) >= 0){
-                awsBackendHcl.appendln("    use_lockfile = true");
+            // Native S3 locking, off by default. Turning it on makes concurrent runs on
+            // the same state wait for each other, which is why it is not the default:
+            // plan only jobs are allowed to run in parallel on purpose.
+            // Worth enabling when something outside Terrakube writes the same state.
+            if(useLockfile) {
+                if(version.compareTo(new ComparableVersion("1.10.0")) >= 0){
+                    awsBackendHcl.appendln("    use_lockfile = true");
+                } else {
+                    log.warn("use_lockfile is enabled but ignored, it requires terraform or tofu 1.10 and this job runs {}", version);
+                }
             }
 
             if(endpoint != null){
--- a/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java
+++ b/executor/src/main/java/io/terrakube/executor/plugin/tfstate/aws/AwsTerraformStateImpl.java
@@ -113,6 +113,15 @@ public class AwsTerraformStateImpl implements TerraformState {
               log.warn("No including backend information");
             }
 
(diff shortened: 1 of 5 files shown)


alfespa17:
there is a queue for workspace jobs you can't run more than one job at the time.

jvdrag:
Thanks for looking. The queue covers Terrakube against itself, but there are two cases it does not reach.

**Plan-only templates bypass it.** In `ScheduleJob` (line 155), a plan-only non-CLI template goes through `isActiveApplyOrDestroyRunning`, which only blocks when an earlier job is currently in a `terraformApply`, `terraformDestroy` or `customScripts` step. Two jobs whose current step is `terraformPlan` run at the same time, on the same workspace and the same state key.

**A cancelled job does not stop terraform.** `InactiveJobs` marks a job cancelled 6 hours after its creation date, `running` included. The executor only reads that status in `UpdateJobStatusImpl#isJobCancelled`, which runs after the terraform process returns, so the process is never interrupted. Meanwhile the API sees a terminal job, `findByWorkspaceAndStatusNotInAndIdLessThan` excludes it, and the next job starts. A job that spent five hours in `waitingApproval` and then began applying is still applying when the next apply begins.

There is also the case outside Terrakube. State lives in plain S3 under a stable key, so anything else pointed at that backend, a workstation or a separate pipeline, is invisible to the queue. We point our repo backends at the Terrakube key on purpose, so that CI plans read the same state the executor writes instead of a second copy that drifts from it.

The change costs nothing when there is no contention. S3 has had native locking since Terraform 1.10, so there is no DynamoDB table involved, and the condition already skips custom endpoints and older versions.

Happy to put it behind a property if you would rather not change the default.

alfespa17:
I see your point, but adding `use_lockfile = true` will basically stop people from running a simple `plan` and the mechanism added [here](https://github.com/terrakube-io/terrakube/issues/2850) will be useless 🤔

What if instead of doing this `awsBackendHcl.appendln("    use_lockfile = true");` you add some new property like this 

https://github.com/terrakube-io/terrakube/blob/41386003409caba9ab8c06e300a9a66d7c94241f/executor/src/main/resources/application.properties#L17

You can leave the default as `false` and if someone would like to enable it they will simply have to add some environment variable, in that way people using terrakube have the option to select what they want?

> We point our repo backends at the Terrakube key on purpose

You did this on purpose but in a real world scenario why would you do this 😅? If you are using terrakube you could the `cloud/remote` block instead of using the key directly to connect to the s3 bucket.

jvdrag:
Good catch on #2850, I had not connected the two. Serializing those plans would have been a regression.

Done as you suggested. `io.terrakube.executor.plugin.tfstate.aws.useLockfile`, set through `AwsUseLockfile`, default `false`. Nothing changes for anyone who does not set it.

Two details:

- The version guard stays. `use_lockfile` is not a valid argument before 1.10, so enabling it on an older version logs a warning instead of writing a backend that fails to init.
- I dropped the custom endpoint restriction. It made sense while this was the default, but now that turning it on is an explicit choice, blocking S3 compatible endpoints would just get in the way of people who know their storage supports conditional writes.

Tests cover the four combinations: off by default, on with 1.10, on with an older version, on with a custom endpoint. 28 pass.

> why would you do this

Fair question. Our CI runs `terraform plan` on pull requests, outside Terrakube, and we wanted it reading the same state the executor writes instead of a second copy that drifts. Pointing the backend at the same key was the shortest path. You are right that the `cloud` block is the better answer for that, and I will look at moving to it. The flag still helps anyone whose state bucket is reachable by other tooling, which is the situation the default cannot know about.

PR #3381, opened 2026-08-08
fix(fanout): request cache federated details, scope ui requests

siler:
i am observing 6.5 second latencies when loading workspace pages. our org has ~4k jobs so far, and on each page load it retrieves all of them from the db. this appears to be unavoidable when loading a workspace because of the `OneToMany` annotation, even when the jobs aren't needed. there are a number of problems that contribute to the latency, but here are a couple of relatively minor changes that optimize things quite a bit.

the federated queries are a classic N+1 problem. since it is unlikely that we need to check for updates to this information in the span of a single request, it is now cached per request. this effectively halves the number of db queries that occur on pageload. this does not cache the results of any checks, just the data retrieved from the database.

the second piece is the ui changes to use fieldsets in various pages such that they only fetch the fields from organization that they actually consume. this was causing significant collection fanout on request and contributing to the latency. i added a little bit of type safety before realizing that the typechecker is actually broken as configured (currently 94 typescript errors on main).

with my test of 40 jobs, the data looks like this:

|                                    | federated queries | total statements | collection fetches |
| ---------------------------------- | ----------------- | ---------------- | ------------------ |
| before both (no memo, no fieldset) | 433               | 541              | 100                |
| memo only                          | 3                 | 111              | 100                |
| fieldset only                      | 87                | 97               | 5                  |
| after both               | 3                 | 13               | 5                  |


jakegroves:
**Overall:** approach is sound, numbers back it up also

Some issues when looking over (i know draft):
1. **`ui/src/domain/Modules/List.tsx` fix has no effect** — confirmed no file imports `./Modules/List`; `App.tsx` only routes to `Registry.tsx` → `ModuleList.tsx` (different component). This call site is dead code from what i can tell. worth double-checking the other 4 edited files are actually routed/reachable
2. **Non-atomic get-or-create in `FederatedLookupService.requestCache()`** — get-then-set on the request attribute isn't atomic, so two threads on the same request could each build a `ConcurrentHashMap` and silently drop one
3.  **N+1 pattern not fully fixed** — `DexGroupServiceImpl.isMemberWithLimitedAccessV2`/`isMemberWithProjectAccess` (lines 145, 154) do one DB query per `Organization` during the same kind of Elide fan-out.. just noting as it won't fully fix n+1 pattern
4. worth havign a shared helper .. tying the query string to its `Sparse<T,K>` type. Two different `SparseOrganization` aliases exist with different field sets (`Workspaces/Details.tsx` vs `Settings/General.tsx`) — same name, different shape. a small `sparseFields()` helper would remove future drift risk

siler:
alright, tackled all of them. a couple of notes:

i noticed that `DexGroupServiceImpl` calls `.toString()` on the `aud` `List` unlike other consumers of the JWT. it's unclear if this is intentional, but it made things a little more complicated w/r/t atomicity. here's the issue:

```
issuer=<Terrakube> audience=<Terrakube>     ← DexAuthenticatedUserImpl
issuer=<Terrakube> audience=<[Terrakube]>   ← DexGroupServiceImpl
```

because the audience isn't the same for all instances that might call the memoizer i couldn't use a single slot for the entire request. it kind of seems unintentional because usually people don't rely on methods like `.toString()` for formatting something that is consumed programmatically. so instead i went with a multi-slot solution based on what was memoizing the federated query.

we can get fancier with the typing solution for queries, but the syntax has a lot of options and it gets fairly complicated unless you only want to cover a subset. the "ultimate" solution would be to have a single (curried) function that accepted the axios instance and all of the parameters needed to construct the query string, then it could return a completely typesafe result.

that would be a lot of code and i'm not sure it needs to live in this pr. let me know if you would like to include it. the downside of the current implementation is that since we're not wrapping axios and the remainder of the query string we have to separate the part of the query string we construct and the type. this could potentially be made a little better but i didn't want to introduce two different request patterns (using axios vs. this new axios consumer). i'm also not sure how useful it would be at the moment since the code doesn't actually typecheck.

anywho, let me know what you think.

alfespa17 (approved):

PR #3392, opened 2026-08-10
feat(executor): carry plan-built artifacts through to apply

### LINKED ISSUE #852: No Such File or Directory: Local Lambda Directory
### Bug description 🐞

When executing a terraform apply command in a Terrakube setup, the expected directory home/cnb/.terraform-spring-boot/executor/<organization_id>/<workspace_id>/builds/lambda_function_<sha256>.zip is not created, which results in a failure with the error: Error: reading ZIP file (builds/lambda_function_<sha256>.zip): open builds/lambda_function_<sha256>.zip: no such file or directory.

### Steps to reproduce

1. **Requirements:**
   - Create an organization in Terrakube.
   - Create a workspace in the organization, either as CLI-driven or VCS, but ensure remote execution is enabled.

2. **Setup:**
   - Use the following module to set up a GitLab runner using version 7.5.0 from the provided module [cattle-ops/gitlab-runner/aws](https://registry.terraform.io/modules/cattle-ops/gitlab-runner/aws/latest).
   - Example invocation:
     ```hcl
     module "vpc" {
       source  = "terraform-aws-modules/vpc/aws"
       version = "5.8.1"

       name = "vpc-${var.environment}"
       cidr = "10.0.0.0/16"

       azs                     = [data.aws_availability_zones.available.names[0]]
       private_subnets         = ["10.0.1.0/24"]
       public_subnets          = ["10.0.101.0/24"]
       map_public_ip_on_launch = false

       enable_nat_gateway = true
       single_nat_gateway = true

       tags = {
         Environment = var.environment
       }
     }

     module "gitlab-runner" {
       source  = "cattle-ops/gitlab-runner/aws"
       version = "7.5.0"
       vpc_id    = module.vpc.vpc_id
       subnet_id = element(module.vpc.private_subnets, 0)
       environment = var.environment
       runner_gitlab = {
         url = var.gitlab_url
         registration_token = var.registration_token
       }
     }
     ```

3. **Run Terrakube in Docker Compose:**
   1. Access the executor container:
      ```sh
      docker exec -it <executor_container_id> bash
      ```
   2. Navigate to the path in executer container:
      ```sh
      cd /home/cnb/.terraform-spring-boot/executor/<organization_id>/<workspace_id>/
      ```
   3. Execute `terraform plan` on local directory invocation terraform code:
      ```sh
      terraform plan
      ```
      - Observe that Terraform files are created and the module repository is cloned within the `.terraform` directory on the executor container.
      - A directory `/home/cnb/.terraform-spring-boot/executor/<organization_id>/<workspace_id>/builds/lambda_function_<sha256>.zip` is created before the plan completes.
   4. After `terraform plan` completes, the files in `/home/cnb/.terraform-spring-boot/executor/<organization_id>/<workspace_id>/` are deleted.
   5. Run `terraform apply`:
      - The steps from 3 to 6 are repeated.
      - However, the directory `/home/cnb/.terraform-spring-boot/executor/<organization_id>/<workspace_id>/builds/lambda_function_<sha256>.zip` is not created again, leading to a failure.

### Expected behavior

The `terraform apply` should recreate the directory `/home/cnb/.terraform-spring-boot/executor/<organization_id>/<workspace_id>/builds/lambda_function_<sha256>.zip` as it did during the `terraform plan`, ensuring the process completes without errors.

**Error Message:**
```plaintext
Error: reading ZIP file (builds/lambda_function_<sha256>.zip): open builds/lambda_function_<sha256>.zip: no such file or directory

### Example repository

_No response_

### Anything else?

_No response_

jakegroves:
**Status: experiment, up for discussion.** Posting this as a working spike against #852, not a finished proposal — open to a different design if there's a better way to do this.

## Why

Terrakube runs every TCL step — plan, apply — in a brand-new git clone, on a stateless pool of executor pods with no shared filesystem and no session affinity. Plan and apply aren't guaranteed to land on the same pod, or the same directory on the same pod. Terrakube also applies from a *saved* plan file (`terraform apply terraformLibrary.tfPlan`), not a live plan+apply in one process.

That combination breaks any Terraform construct where a build artifact gets produced by something outside Terraform's own apply-time execution graph, then referenced by a later resource via a bare file path:

- `data "archive_file"` — its `Read` has a side effect of writing a zip to disk, but Terraform only calls `Read` once, during plan (since its inputs are statically known), and freezes the result into the saved plan. Apply never re-invokes it. The zip plan wrote only ever existed in plan's now-discarded clone.
- A before-script step that builds some output file directly — same problem: nothing regenerates it in apply's clone.
- Even `path.module`-derived paths make this worse, not better: `path.module` resolves to plan's own temp clone directory, so a value computed via `path.module` freezes an *absolute* path into the saved plan that's meaningless once apply runs somewhere else.

Originally reported as #852: a Lambda zip built by a before-script, referenced via `local_existing_package`, plans fine, apply fails with "no such file or directory".

Affected resource types:

- `aws_lambda_function`
- `aws_lambda_layer_version`
- `aws_s3_object` / `aws_s3_bucket_object`
- `aws_elastic_beanstalk_application_version`
- `aws_glue_job`
- `aws_cloudfront_function`
- `aws_codebuild_project`
- `aws_appsync_resolver` / `aws_appsync_function`
- `google_cloudfunctions_function` / `google_cloudfunctions2_function`
- `google_storage_bucket_object`
- `google_app_engine_standard_app_version` / `google_app_engine_flexible_app_version`
- `azurerm_linux_function_app` / `azurerm_windows_function_app` / `azurerm_function_app`
- `azurerm_linux_web_app` / `azurerm_windows_web_app` / `azurerm_app_service`
- `azurerm_storage_blob`
- `azurerm_template_deployment`
- `helm_release`
- `kubernetes_manifest` / `kubectl_manifest`
- `data "archive_file"` (upstream of most of the above)

Worth noting: it's *not* needed for modules that deliberately engineer around this with a real apply-time provisioner instead of a data source — e.g. `terraform-aws-modules/lambda`'s `create_package = true` rebuilds its zip inside a `null_resource` `local-exec` provisioner with an explicit `depends_on`, which genuinely re-executes in apply's own clone. Verified live: a job using that module's `create_package` path completes successfully with zero use of this feature. But most real-world Terraform isn't engineered that way — this covers the common case.

## What

- **Opt-in via Template TCL.** Nothing happens automatically — a workspace only gets this behavior once its Template's plan step is edited to add an `artifacts:` list. Existing workspaces with no `artifacts:` declared see zero behavior change; no packaging step runs, nothing extra gets stored
  ```yaml
  flow:
    - type: "terraformPlan"
      step: 100
      name: "Plan"
      artifacts:
        - "build/**"        # packaged after a successful plan with changes
        - "${LAMBDA_DIR}/*.zip"   # ${VAR} patterns are substituted; dropped
                                  # entirely (not blank-matched) if unset
    - type: "terraformApply"
      step: 200
      name: "Apply"

Everything under build/** (and any *.zip in $LAMBDA_DIR, if set) gets tarballed after step 100 and extracted into step 200's clone before terraform apply runs. Remove the artifacts: list to go back to the old behavior for that workspace
- New `artifacts:` field on TCL plan steps — a list of glob patterns (with `${VAR}` substitution) describing which files a plan step is expected to produce
- After a successful multi-step plan with changes, the executor packages every file matching those patterns into a tarball and stores it alongside the plan file
- Before the apply step runs, the executor downloads and extracts that tarball into apply's fresh clone, before `terraform apply` starts
- Checksum-verified end to end: a declared artifact that's missing or doesn't match its stored checksum fails the apply step loudly, rather than silently proceeding without it
- Implemented across all four state backends — AWS S3, Azure Blob, GCP Cloud Storage, and local disk — each storing artifacts under their own container/prefix (`plan-artifacts/`, separate from `tfstate/`) so a storage lifecycle policy can expire build artifacts independently of plan/state history.
- New `job.terraform_plan_artifacts` / `job.terraform_plan_artifacts_checksum` columns.
- Companion PR to terrakube-io/terrakube-spring-boot-starter#87 adds the matching `JobAttributes` fields the executor needs to read/write these through the generated client.

## How

- `Flow.artifacts` (TCL model) → `ExecutorContext.artifactPatterns` (API) → `TerraformJob.artifactPatterns` (executor).
- `ArtifactPatternResolver` substitutes `${VAR}` references in each glob; a pattern referencing an unset variable is dropped entirely, rather than substituting to an empty string (which would silently widen the glob to match everything).
- `ArtifactGlobMatcher` + `ArtifactPackagingService` walk the workspace after a successful plan and package matches into `plan-artifacts.tar.gz` via `TarGzArchiver`.
- Packaging is gated on `!isDestroy && exitCode == 2 && multiStepJob` — only plans with real changes, in multi-step jobs, get packaged; plan-only jobs and no-op plans skip the work entirely.
- `ArtifactVerifier` checksums the tarball on the way up and the way down.
- Each state backend (`AwsTerraformStateImpl`, `AzureTerraformStateImpl`, `GcpTerraformStateImpl`, `LocalTerraformStateImpl`) implements `saveArtifacts`/`downloadArtifacts` against its own container/prefix.
- 631 api + 190 executor tests green (new + existing), plus live verification on a real minikube deployment against real AWS, including an A/B comparison confirming the original bug reproduces with the feature stashed out.

## Related

- Fixes #852
- Depends on terrakube-io/terrakube-spring-boot-starter#87

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorContext.java
@@ -16,6 +16,8 @@ import java.util.List;
 public class ExecutorContext {
     private List<Command> commandList;
     private List<Command> onFailureList;
+    private List<String> artifactPatterns;
+    private boolean multiStepJob;
     private String type;
     private String organizationId;
     private String workspaceId;
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
@@ -143,6 +143,8 @@ public class ExecutorService {
 
         executorContext.setCommandList(flow.getCommands());
         executorContext.setOnFailureList(flow.getOnFailure());
+        executorContext.setArtifactPatterns(flow.getArtifacts());
+        executorContext.setMultiStepJob(job.getStep() != null && job.getStep().size() > 1);
         executorContext.setType(flow.getType());
         executorContext.setIgnoreError(flow.isIgnoreError());
         executorContext.setTerraformVersion(job.getWorkspace().getTerraformVersion());
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/model/Flow.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/model/Flow.java
@@ -19,6 +19,7 @@ public class Flow {
     private int step;
     List<Command> commands;
     List<Command> onFailure;
+    List<String> artifacts;
 
     List<ScheduleTemplate> templates;
 
--- a/api/src/main/java/io/terrakube/api/rs/job/Job.java
+++ b/api/src/main/java/io/terrakube/api/rs/job/Job.java
@@ -71,6 +71,12 @@ public class Job extends GenericAuditFields {
     @Column(name = "terraform_plan")
     private String terraformPlan;
 
+    @Column(name = "terraform_plan_artifacts")
+    private String terraformPlanArtifacts;
+
+    @Column(name = "terraform_plan_artifacts_checksum")
+    private String terraformPlanArtifactsChecksum;
+
     @CreatePermission(expression = "user is a super service")
     @UpdatePermission(expression = "user is a super service")
     @Column(name = "approval_team")
--- a/api/src/main/resources/db/changelog/changelog.xml
+++ b/api/src/main/resources/db/changelog/changelog.xml
@@ -115,4 +115,5 @@
     <include file="/db/changelog/local/changelog-2.33.0-job-command-comment-id.xml"/>
     <include file="/db/changelog/local/changelog-2.33.0-job-soft-delete.xml"/>
     <include file="/db/changelog/local/changelog-2.33.0-job-status-index.xml"/>
+    <include file="/db/changelog/local/changelog-2.33.0-plan-artifacts.xml"/>
 </databaseChangeLog>
--- a/api/src/main/resources/db/changelog/local/changelog-2.33.0-plan-artifacts.xml
+++ b/api/src/main/resources/db/changelog/local/changelog-2.33.0-plan-artifacts.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">
+    <changeSet id="1" author="terrakube">
+        <addColumn tableName="job">
+            <column name="terraform_plan_artifacts" type="varchar(1024)"/>
+        </addColumn>
+        <addColumn tableName="job">
+            <column name="terraform_plan_artifacts_checksum" type="varchar(64)"/>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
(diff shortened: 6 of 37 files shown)


jakegroves:
build will fail until decide if this is a good approach moving forward and would need the https://github.com/terrakube-io/terrakube-spring-boot-starter/pull/87 released... but draft PR for idea for now

alfespa17:
This could work but you will need to have an specific template for each use case 🤔, for example one for aws_lambda_function or another for azurerm_template_deployment depending of which files you want to save, unless you put all the posible conditions in a single template 🤔

jakegroves:
> This could work but you will need to have an specific template for each use case 🤔, for example one for aws_lambda_function or another for azurerm_template_deployment depending of which files you want to save, unless you put all the posible conditions in a single template 🤔

i was thinking if you set it as a env var, then the workspace could override it as an env var... possibly

alfespa17:
> i was thinking if you set it as a env var, then the workspace could override it as an env var... possibly

This could be a good idea, we do something similar with the optional ephemeral executor parameters instead of creating new properties for the workspace table

PR #3401, opened 2026-08-14
feat(notification): add Slack/Teams/webhook notifications for job status changes

jakegroves:
### What

Notification system so you can get pinged when a run changes status — configurable per organization (applies to every workspace) and/or per workspace, both additive.

- Slack, Microsoft Teams, and generic webhook channels
- Filter by run status and now by template (e.g. notify on "Plan and Apply" but not bare "Plan")
- Per-config message style: Detailed (full card) or Simple (one-line ping)
- Retryable (429/5xx) vs. terminal (other 4xx) failure classification, with a manual retry action for terminal deliveries
- destinationUrl validated against private/loopback/link-local ranges before every send, with an opt-out for self-hosted setups that intentionally target something internal

### Why these choices

- **Additive only, no override/suppression.** Org-wide and workspace configs both fire independently. An earlier draft had an "override for this workspace" concept and it just made state harder to reason about for no real benefit.
- **Transactional outbox instead of sending inline.** Most real status transitions happen via a plain JPA save in the Quartz scheduler, not through Elide, so there's no request/response cycle to hang a synchronous send off anyway — and decoupling enqueue from delivery means a slow or down destination can't block a job's status update.
- **Plain Incoming Webhooks, not an OAuth bot install.** Simpler for a user to set up (paste a URL vs. an app-install flow), at the cost of no message threading/editing-in-place. Worth noting Slack's own TFE integration has the same limitation.
- **Read access to an org-wide config is broader than write access.** Anyone managing a workspace it applies to can see it, even if only org-wide admins can edit it — otherwise a workspace-level manager has no way to know an org-wide notification is about to fire for their runs.
- **Message style is an explicit per-config toggle, not inferred from job status.** First pass tried auto-picking detailed vs. compact based on status; in practice that just meant losing information people wanted, so it's user-controlled now.
- **Template filter narrows, never widens.** An empty selection always means "every template." No way to accidentally configure a notification that silently never fires.

### How it was tested

- Local webhook receiver for the generic webhook channel
- A real Slack workspace for the Slack channel — this is what surfaced the delivery bugs fixed in the second commit (Slack rejecting payloads missing top-level `text`, and reporting some errors as HTTP 200 with a plain-text body instead of a 4xx)
- Full backend and frontend test suites pass

### Related

Slack screenshots:
image
image
image

Companion Terraform provider terrakube-io/terraform-provider-terrakube#160 in progress to manage these configurations as code — will link that PR here once it's up

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/notification/ClaimedOutbox.java
+++ b/api/src/main/java/io/terrakube/api/plugin/notification/ClaimedOutbox.java
@@ -1,0 +1,8 @@
+package io.terrakube.api.plugin.notification;
+
+import java.util.Date;
+
+import io.terrakube.api.rs.notification.NotificationConfiguration;
+
+record ClaimedOutbox(NotificationConfiguration configuration, String payload, int attemptCount, Date lastAttemptAt) {
+}
--- a/api/src/main/java/io/terrakube/api/plugin/notification/JobNotificationTrigger.java
+++ b/api/src/main/java/io/terrakube/api/plugin/notification/JobNotificationTrigger.java
@@ -1,0 +1,128 @@
+package io.terrakube.api.plugin.notification;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+import io.terrakube.api.plugin.notification.payload.NotificationContext;
+import io.terrakube.api.plugin.notification.payload.NotificationPayloadRenderer;
+import io.terrakube.api.rs.job.Job;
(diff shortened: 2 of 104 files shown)


alfespa17:
Hello @jakegroves you should block the access to this property `NotificationConfiguration.signingSecret` after the elide entity is created, if not anyone in the org could start sending messages.

You could use something like [this](https://github.com/terrakube-io/terrakube/blob/main/api/src/main/java/io/terrakube/api/rs/checks/variable/UserReadSecret.java) `@ReadPermission(expression = "user read signingSecret")`

alfespa17:
Can you add the `io.terrakube.notification.ssrf.blockPrivateNetworks=true` to the application-demo.properties too please.

jakegroves:
> Can you add the `io.terrakube.notification.ssrf.blockPrivateNetworks=true` to the application-demo.properties too please.

done both, good spots, i also updated PR to link the provider to the pr

alfespa17 (approved):

PR #3402, opened 2026-08-14
fix: re-queue last applied configuration for CLI/API workspace Run now

KetchupBomb:
### Problem
For CLI/API-driven workspaces (no VCS connection, `branch = "remote-content"`), clicking
**Run now** in the UI fails. On v2.31.2 it throws a `NullPointerException`; the run cannot
start.

Root cause is how the last configuration-version pointer is tracked. In
`RemoteTfeService.createRun`, `workspace.source` was overwritten with the uploaded
configuration-version's tarball URL on **every** run — including speculative (plan-only)
runs and runs later discarded before approval. So `workspace.source` tracked the last
configuration *created*, not the last configuration *applied*.

The UI "Run now" job carries no `overrideSource`, so the executor falls back to
`workspace.source`. As a result "Run now" could:
- re-queue the **wrong** configuration (a later speculative plan's), or
- fail outright when `workspace.source` was still the initial `"empty"` placeholder
  (never-applied workspace) — the NPE above.

### Fix
Track the "last applied configuration" at the moment an **apply is dispatched** (i.e. the
plan was approved and apply/destroy is being attempted), rather than at run creation:

- **`RemoteTfeService.createRun`** — no longer eagerly writes `workspace.source`. The
  current run is unaffected (it already uses `job.overrideSource`).
- **`ExecutorService`** — new `persistAppliedConfigurationSource(...)`, invoked when a
  `terraformApply` flow is dispatched. It records the configuration-version tarball onto
  `workspace.source` **only** for CLI/API workspaces (`vcs == null && ssh == null &&
  branch == "remote-content"`), and is a no-op if the source is null/blank/unchanged.
  Both the manual-approve and auto-apply paths funnel through this single dispatch point,
  so apply and destroy are covered, success or failure.
- **`ExecutorService` guard** — the existing remote-content guard now also treats the
  literal `"empty"` placeholder as "no configuration uploaded", failing with a clear
  message instead of an NPE (defense-in-depth for direct API calls / stale clients).
- **UI (`CreateJob` / `WorkspaceDetails`)** — "Run now" is disabled with an explanatory
  tooltip when a CLI/API workspace has no applied configuration yet
  (`source === "empty" && branch === "remote-content"`). VCS workspaces are unaffected and
  remain always-enabled, since they always have a git source to run.

### VCS / SSH safety
VCS and SSH workspaces are double-guarded: `persistAppliedConfigurationSource` returns
early when `vcs`/`ssh` is set, and such workspaces never have `branch == "remote-content"`
(the remote-content override lives on the Job, not the Workspace). Their `workspace.source`
(a git repository URL) is never overwritten. The UI change likewise never triggers for them.

### Behavior after this change
- Speculative `terraform plan` runs no longer change the re-run pointer.
- A run that is created but discarded before approval no longer changes it.
- Once a plan is approved and apply/destroy is dispatched, that configuration-version
  becomes the workspace's source, so a later UI **Run now** re-queues that same
  configuration (plan/apply or destroy is chosen fresh at click time).
- A never-applied CLI/API workspace shows a disabled **Run now** with a tooltip instead of
  erroring.

### Tests
- `ExecutorServiceTest` (+7): apply persists the source; speculative plan does not; VCS and
  SSH workspaces are skipped; non-`remote-content` branch is skipped; idempotent when
  unchanged; blank source ignored.
- `Create.test.tsx` (new): "Run now" enabled with no reason (VCS case); disabled with a
  reason (unconfigured CLI/API); disabled without plan permission.

### Validation
- API (JDK 25, Maven): `ExecutorServiceTest` 10/10; regression run across
  `ExecutorServiceTest, RemoteTfeServiceTest, ScheduleJobTest, JobManageHookTest` — 52/52,
  BUILD SUCCESS.
- UI: `Create.test.tsx` 3/3; `tsc --noEmit -p tsconfig.test.json` clean.

### Notes
- No DB migration: `workspace.source` is reused as the "last applied configuration" pointer
  for CLI/API workspaces (consistent with how these workspaces already use `source`).
- Context on why a "vcs"-named resource is used for a CLI workflow: the provider's
  `terrakube_workspace_cli` resource does not expose a `folder` attribute, so workspaces
  needing a subdirectory are created via `terrakube_workspace_vcs` with
  `branch = "remote-content"` and no `vcs_id`. To Terrakube these are indistinguishable from
  `workspace_cli` workspaces, and this fix covers them identically. (A `folder` attribute on
  `terrakube_workspace_cli` would be a reasonable follow-up in the provider repo.)

### Scope / non-goals
- Does not add a `folder` attribute to any provider resource (separate repo).
- Does not change VCS, webhook, or scheduled-run behavior.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import io.terrakube.api.plugin.scheduler.job.tcl.executor.ephemeral.EphemeralExecutorService;
 import io.terrakube.api.plugin.scheduler.job.tcl.executor.persistent.PersistentExecutorService;
 import io.terrakube.api.plugin.scheduler.job.tcl.model.Flow;
+import io.terrakube.api.plugin.scheduler.job.tcl.model.FlowType;
 import io.terrakube.api.plugin.token.dynamic.DynamicCredentialsService;
 import io.terrakube.api.plugin.vcs.TokenService;
 import io.terrakube.api.repository.GlobalVarRepository;
@@ -28,6 +29,7 @@ import io.terrakube.api.repository.ReferenceRepository;
 import io.terrakube.api.repository.SshRepository;
 import io.terrakube.api.repository.VariableRepository;
 import io.terrakube.api.repository.VcsRepository;
+import io.terrakube.api.repository.WorkspaceRepository;
 import io.terrakube.api.rs.collection.Collection;
 import io.terrakube.api.rs.collection.Reference;
 import io.terrakube.api.rs.collection.item.Item;
@@ -38,6 +40,7 @@ import io.terrakube.api.rs.job.address.AddressType;
 import io.terrakube.api.rs.ssh.Ssh;
 import io.terrakube.api.rs.vcs.Vcs;
 import io.terrakube.api.rs.vcs.VcsConnectionType;
+import io.terrakube.api.rs.workspace.Workspace;
 import io.terrakube.api.rs.workspace.parameters.Category;
 import io.terrakube.api.rs.workspace.parameters.Variable;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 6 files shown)


KetchupBomb:
@alfespa17 , we chatted in the past. Our org is looking to adopt TK. We are heavy CLI/API-driven users. We'll be contributing to various aspects if you're open to receive suggestions.

alfespa17 (approved):

alfespa17:
Make sense 🤔

alfespa17:
> @alfespa17 , we chatted in the past. Our org is looking to adopt TK. We are heavy CLI/API-driven users. We'll be contributing to various aspects if you're open to receive suggestions.

All help is welcome 😁

PR #3411, opened 2026-08-17
feat(ui): add in-app API docs page

jakegroves:
## Why

There was no self-service way to discover or explore the Terrakube API. Elide exposes a live OpenAPI document at `/doc`, but only as raw JSON — not something a developer can browse. The only alternative was a manually-maintained snapshot (`openapi-spec/v2_27_0.yml`) that a maintainer re-exports and commits by hand on each release, with no guarantee it matches what's actually deployed.

That's friction for anyone trying to build against the API — internal engineers, customers automating against Terrakube, or partners integrating with it. Today the only path in is reading a static YAML file in the repo or hand-crafting requests against `/doc`'s raw JSON. This adds a proper, interactive entry point: browse the live API, see real request/response shapes, and execute calls directly from the browser using your own session.

## Why Scalar

Evaluated three renderers against our spec (~30 resource tags, full JSON:API CRUD per resource):

| | Minified | Gzipped | Interactive |
|---|---|---|---|
| swagger-ui-react | 1.30 MB | 362 KB | Yes |
| redoc | 999 KB | 293 KB | No (read-only) |
| **Scalar** | 3.21 MB | 959 KB | Yes |

Scalar is the largest of the three — its React wrapper pulls in a full Vue runtime under the hood — but it was chosen anyway because it's the only option that pairs **interactive "Try it" request execution** with a large-spec-friendly layout (sidebar nav + search, similar to Redoc's tag grouping, vs. Swagger UI's flat expandable list). Given our spec's size, browsability mattered more than the extra ~600KB, and the page is lazy-loaded on its own route (`/api-docs`) and gzip-compressed in production, so it never touches the main app bundle or any other page's load time.

## What it does

- New `/api-docs` page, linked from the Help menu (top-right dropdown).
- Full-bleed layout outside the normal app shell — Scalar renders its own sidebar/nav, so nesting it inside our `AppSidebar` produced a redundant double sidebar. A small "← Back to Terrakube" link replaces the app chrome on this page.
- Fetches the spec live from the API's `/doc` endpoint (requires auth, like everything else) via the existing authenticated axios client, rather than relying on a stale committed copy.
- Auto-attaches the signed-in user's bearer token to every "Try it" request via Scalar's `onBeforeRequest` hook, so requests work immediately without pasting in a token manually.
- Overrides `baseServerURL` to the API's actual origin — the spec's `servers` entry is a relative path (Elide doesn't know its own public hostname when generating it), so without this, "Try it" requests would resolve against the docs page's own origin instead of the API's.

Requests fired from this page run as the signed-in user with their real permissions — it's not a sandbox, and it's not an elevated/service credential.

## How to use it

1. Sign in to Terrakube as usual.
2. Click the **?** help menu (top right) → **API Docs**.
3. Browse endpoints via the sidebar or search.
4. Click into an operation and use **Test Request** to execute it live against the API — your session's token and permissions apply automatically.

## Screenshots
image
image
image

## Testing

- Unit tests (TDD) for `ApiDocsPage` (spec fetch, `content` passed to Scalar, `baseServerURL`, bearer token attached/omitted) and `HelpMenu` (in-app navigation vs. the existing external links).
- Full `ui` test suite, typecheck, and lint pass (one pre-existing flaky/unrelated test in `EditNotificationConfiguration` excluded — confirmed unaffected, passes in isolation).

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -6,6 +6,7 @@
   "dependencies": {
     "@ant-design/icons": "^6.3.2",
     "@monaco-editor/react": "^4.7.0",
+    "@scalar/api-reference-react": "^0.9.63",
     "ansi-to-react": "^6.2.6",
     "antd": "^6.5.4",
     "axios": "^1.19.0",
--- a/ui/src/components/HelpMenu/HelpMenu.css
+++ b/ui/src/components/HelpMenu/HelpMenu.css
@@ -53,6 +53,7 @@
   cursor: pointer;
   display: flex !important;
   align-items: center !important;
+  gap: 12px !important;
   color: #1a1a1a !important;
   font-size: 14px !important;
   line-height: 1.4 !important;
@@ -62,6 +63,11 @@
 
 .help-menu-item:hover {
   background-color: #f5f5f5 !important;
+}
+
+.help-menu-item-icon {
+  color: #656a76 !important;
+  font-size: 16px !important;
 }
 
 /* Dark mode styles */
@@ -82,4 +88,8 @@
 
 [data-theme="dark"] .help-menu-item:hover {
   background-color: #21262d !important;
+}
+
+[data-theme="dark"] .help-menu-item-icon {
+  color: #8b949e !important;
 }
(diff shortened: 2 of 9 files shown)


alfespa17:
This looks pretty cool 😎

alfespa17 (approved):

alfespa17:
One thing to keep in mind is that /docs only include the elide endpoints, it does not include other endpoints that are manage "manually" for example the "personal acces token endpoint" there is a way to customize the docs in elide but I never tested

PR #3430, opened 2026-08-21
fix(vcs): mint GitHub App tokens on demand instead of on a timer

jplorier:
## Problem

GitHub App installation tokens are kept warm by a Quartz job per cached row, firing every 55 minutes (`createTask(3300, …)`) whether or not anyone needs a token, while `getGitHubAppToken` hands out whatever is in the row.

**That fails silently.** An installation token is valid for one hour, so if the timer misses a beat — schedule lost on restart, job throws, row written without an expiry — the cached token expires and every consumer keeps being handed it. GitHub answers `401`, which reaches the user as

```
org.eclipse.jgit.api.errors.TransportException: https://github.com/<owner>/<repo>.git: not authorized
```

naming neither the token nor the installation. Nothing in the flow reconsiders the row, so it never recovers on its own. This is observable in the wild: a row whose `token` was last written 11 days earlier was still being served, and clones had been failing since the day after.

**And the work is mostly wasted.** A token has no value except at the moment a caller needs one. A workspace used twice a day costs 26 refreshes a day.

Two smaller problems in the same path:

* The installation id is cached and never revisited. Ids are not stable — reinstalling an app, or installing it on another account, issues new ones — so an id GitHub stops accepting leaves that owner permanently broken.
* When the first installation lookup fails, the empty `GitHubAppToken` that was built is saved anyway and a refresh scheduled for it. That row poisons the cache: later reads find it, return a `null` token, and produce the same silent unauthenticated clone.

## Change

**Mint on demand, drop the timer.** A cached token is served while it is valid and minted when it is not. Every consumer — job dispatch, imports, module refresh, webhooks, discovery — already goes through `getGitHubAppToken`, so one expiry check there covers all of them. This is both the cheaper path (two mints instead of 26, for that twice-a-day workspace) and the one that cannot fail silently: staleness is detected at the moment it matters, and the error surfaces on the request that needed the token.

**Resolve the installation on every mint** via `/repos/{owner}/{repo}/installation` instead of trusting the cached id. One call, on a path that now runs at most once per token lifetime, and it also proves the repository is inside that installation's scope — which is the thing the caller is about to rely on. An account holds at most one installation of a given app, so this identifies the owner's installation too.

**Keep failures loud.** An unresolvable installation, or a mint returning no token, raises instead of caching a token-less row or serving the expired one still present.

**Upgrade safety.** Existing deployments have refresh triggers persisted in their Quartz tables pointing at a class that no longer exists; firing one would fail to instantiate it. `GitHubAppTokenScheduleCleanup` unschedules them on startup, and a failure there only warns — it must not stop the application.

Removed: `ScheduleGitHubAppToken`, `ScheduleGitHubAppTokenService`, the `refreshAccessToken` path they drove, and the orphaned-row cleanup that existed only to service them.

No schema change: this builds on the `expires_at` column from #3336.

## Tests

`GitHubTokenServiceTest`, using the existing local `HttpServer` harness:

| test | covers |
|---|---|
| `validCachedToken_isServedWithoutAnyHttpCall` | minting stays confined to when a token is needed |
| `expiredCachedToken_mintsAndUpdatesInPlace` | refresh updates the row rather than duplicating it |
| `missingExpiry_treatedAsExpired` | rows written before an expiry existed self-heal instead of being served forever |
| `noCachedRow_mintsAndCachesOne` | first mint populates owner, app id and installation |
| `staleInstallationId_isReplacedOnMint` | an id GitHub no longer accepts is replaced, not reused |
| `noInstallationServesRepository_failsAndKeepsRow` | failure surfaces; the cached row is not blanked |
| `firstMintWithNoInstallation_cachesNothing` | no poisoned row |
| `repositoryMissing_isRejected` | resolution requires owner and repository |
| `getInstallationToken_missingExpiresAt_doesNotThrow` | discovery path unaffected |

`GitHubTokenServiceTest` 9/9, and the wider selection passes: `VcsRepositoryDiscoveryGitHubTests`, `GitHubWebhookServiceTest`, `VcsTests`, `ScheduleVcsTest`, `VcsRepositoryAccessServiceTest`, `VcsRepositoryDiscoveryFacadeTest`, `VcsRepositoryControllerTest` and the other provider suites, plus a clean `test-compile` of the module.

## Note

An earlier revision of this branch re-resolved the installation only as a *retry* after a stale id failed, and fell back to serving the cached row when a refresh failed. Both were wrong in the same direction — the second reproduces exactly the silent-expired-token bug being fixed — and have been replaced by the design above.


alfespa17:
I will leave this as draft until you test it and don't forge to put some screenshot to validate it is working correctly

Issue #3382, opened 2026-08-08
/terraform/index.json has no stale-cache fallback, so one failed upstream fetch wedges every Terraform job


jack60612:
### Summary

`TofuJsonController` serves a stale cached copy when the upstream releases fetch fails ([`9bf308cc`](https://github.com/AzBuilder/terrakube/commit/9bf308cc3695d66fdb0d3a290b2427d5688f9622), "return cached tofu releases on GitHub API failure to prevent executor NPE"). `TerraformJsonController` never got the equivalent. On 2.32.2 a single failed fetch of the Terraform index therefore returns HTTP 500 whose body is the configured URL, and every executor pinned to Terraform then dies:

```
TerraformDownloader : Error fetching terraform releases 500 Internal Server Error from GET http://<api>/terraform/index.json
SimpleAsyncUncaughtExceptionHandler : Unexpected exception occurred invoking async method: ExecutorJobImpl.createJob
java.lang.NullPointerException: Cannot invoke "io.terrakube.terraform.TerraformResponse.getVersions()" because "this.terraformReleases" is null
```

The job stays `running` with an almost empty log until it is cancelled by hand — exactly the failure mode #3260 fixed for OpenTofu. Workspaces on OpenTofu were unaffected throughout, which is what made the asymmetry obvious.

### What triggered it for us

The API's own fetch of `https://releases.hashicorp.com/terraform/index.json` started returning **HTTP 494** from CloudFront:

```
DownloadReleasesService : No GitHub token provided - using unauthenticated request (subject to rate limits)
DownloadReleasesService : Failed to download releases from https://releases.hashicorp.com/terraform/index.json: HTTP 494
TerraformJsonController : Unknown status code [494]
```

`curl` from a pod in the same namespace, sending the identical `User-Agent: releases-downloader` and `Accept: application/json` that `DownloadReleasesService` sets, gets **200** for the same URL. No proxy is configured, the JVM has no proxy system properties, and the same client fetches `api.github.com` for the OpenTofu index without trouble. So something about the Reactor Netty request is being refused where a plain client is not. I could not pin down the offending difference from the source alone — `DownloadReleasesService` only sets `followRedirect(true)`, `proxyWithSystemProperties()`, that User-Agent, that Accept, and a conditional bearer token.

[...]

Issue #3376, opened 2026-08-05
Improve reliability of Terraform/OpenTofu version resolution and binary caching


jakegroves:
### Feature description 💡

The executor currently re-fetches the Terraform/OpenTofu release manifest (`releases.hashicorp.com/terraform/index.json`, or the tofu equivalent proxied through `AzBuilderApiUrl`) on every single terraform/tofu command invocation within a job (`init`, `plan`, `apply`, `destroy`, etc.), via `terraform-spring-boot-starter`'s `TerraformDownloader`. There's no retry/backoff on these calls, no fallback mirror, and the actual binary zip cache (`~/.terraform-spring-boot/...`) only helps on long-lived pods — on ephemeral executors, every job is a cold start regardless.

This creates two reliability risks worth investigating for a project positioned as a Terraform Cloud alternative:

1. **External dependency hammering** — every job run hits HashiCorp's/GitHub's API directly, with no caching or backoff, which doesn't scale well and isn't resilient to upstream rate limits or outages.
2. **Coupling to our own control-plane API** — the tofu release index is proxied through `AzBuilderApiUrl`, so if that API is down, it also blocks version resolution for jobs that do manage to start, not just job scheduling.

Proposed directions to improve resiliency:

- Add a Quartz job (following the existing `ModuleRefreshJob`/`ProviderRefreshService` pattern in `io.terrakube.api.plugin.scheduler`) that periodically resolves the Terraform/OpenTofu release manifests once, centrally, and persists the result to a new DB table (e.g. `EngineVersion`: engine, version, release date, latest flag) — mirroring how `Module`/`ModuleVersion` already work. Version/range resolution (including `latest`) then reads from this table instead of hitting HashiCorp/GitHub per job, and the API always hands executors an already-resolved, concrete version — so executors (ephemeral or not) never need to query a release manifest themselves, only fetch/cache the binary for a version they're told to use
[...]

PR #3434, opened 2026-08-21
fix(api): stop DownloadReleasesService accumulating User-Agent headers

klinux:
## Problem

`DownloadReleasesService` is a singleton `@Service` that keeps the injected `WebClient.Builder` in a field and calls `defaultHeaders(...)` on it for **every** request:

```java
@Service                                        // singleton
public class DownloadReleasesService {
    private WebClient.Builder webClientBuilder;  // injected once, mutable

    public void downloadReleasesToFile(...) {
        WebClient webClient = webClientBuilder
                .clientConnector(...)
                .defaultHeaders(h -> {
                    h.add("User-Agent", "releases-downloader");   // <-- appends, never replaces
```

`DefaultWebClientBuilder.defaultHeaders()` applies the consumer **immediately** to the builder's own `HttpHeaders`. Because the builder is shared and mutated in place, each call appends *one more* `User-Agent` value. The header grows for the lifetime of the pod. (`setAccept()` uses `set()` and is unaffected — only `add()` leaks.)

Eventually the remote rejects the request. Measured against `api.github.com` from inside an affected cluster:

| `User-Agent` size | Response |
|---|---|
| ≤ 6800 bytes | 200 |
| 7200–7600 bytes | 500 |
| ≥ 8000 bytes | 400 |

## Impact

Once over the threshold, `/tofu/index.json` returns 500 and executors get `tofuReleases == null` in `TerraformDownloader`:

```
ERROR TerraformDownloader : Error fetching tofu releases 500 Internal Server Error
                            from GET http://terrakube-api-service:8080/tofu/index.json
ERROR SimpleAsyncUncaughtExceptionHandler : Unexpected exception occurred invoking
                            async method: ExecutorJobImpl.createJob(TerraformJob)
java.lang.NullPointerException: Cannot invoke "java.util.List.size()"
                            because "this.tofuReleases" is null
```

This is the upstream trigger behind the failure mode described in #3382 and #3376. The stale-cache fallbacks added since help the API survive a failed fetch, but the fetch itself keeps failing until the pod restarts, because the header never shrinks.

Two things make it worse in practice:

- The same service backs both the terraform and tofu endpoints, so the header grows at twice the rate.
- `TerraformDownloader`'s constructor fetches **both** release lists unconditionally, so workspaces using Terraform break on a failing tofu fetch too.

On releases before 2.33.0-beta.7 the NPE escapes the `@Async` `createJob()` uncaught, `SimpleAsyncUncaughtExceptionHandler` merely logs it, and the job stays in `running` forever while the k8s Job reports `Completed`. We hit exactly this on 2.31.3 in production, on both ephemeral and long-lived executors.

## Fix

`clone()` the builder before mutating it, so every call starts from the injected prototype. One line, no behaviour change beyond not leaking state between calls.

## Test

Adds `DownloadReleasesServiceTest`, which performs five downloads against a local `HttpServer` and asserts exactly one `User-Agent` value per request. No new dependencies.

Against the previous code it fails on the second request, which is the bug reproduced directly:

```
org.opentest4j.AssertionFailedError: request 2 sent 2 User-Agent values:
    [releases-downloader, releases-downloader] ==> expected: <1> but was: <2>
```

With the fix:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/json/DownloadReleasesService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/json/DownloadReleasesService.java
@@ -26,7 +26,13 @@ public class DownloadReleasesService {
     }
 
     public void downloadReleasesToFile(String releasesUrl, File releasesFile, String githubToken) {
+        // clone() is essential: webClientBuilder is a field on this singleton bean, and
+        // DefaultWebClientBuilder.defaultHeaders() applies the consumer immediately to the
+        // builder's own HttpHeaders. Mutating the shared builder would make h.add("User-Agent", ..)
+        // below append one more value on every call, growing the header until the remote
+        // rejects the request (api.github.com starts returning 400/500 past ~7KB of headers).
         WebClient webClient = webClientBuilder
+                .clone()
                 .clientConnector(new ReactorClientHttpConnector(
                         HttpClient.create()
                                 .followRedirect(true)
--- a/api/src/test/java/io/terrakube/api/plugin/json/DownloadReleasesServiceTest.java
+++ b/api/src/test/java/io/terrakube/api/plugin/json/DownloadReleasesServiceTest.java
@@ -1,0 +1,73 @@
+package io.terrakube.api.plugin.json;
+
+import com.sun.net.httpserver.HttpServer;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.web.reactive.function.client.WebClient;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class DownloadReleasesServiceTest {
+
+    private HttpServer server;
+    private final List<List<String>> receivedUserAgents = new ArrayList<>();
+    private Path tempDir;
+
+    @BeforeEach
+    void startServer() throws IOException {
+        tempDir = Files.createTempDirectory("releases-test");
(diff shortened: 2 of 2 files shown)


alfespa17 (approved):

alfespa17:
I think you found a bigger problem, I will create an additional PR to fix the rest of the places where the web client builder is used in the same way. Thank you for your help @klinux

klinux:
Happy to help! Since you're going to sweep the rest, here's what I found surveying the other `WebClient.Builder` usages on `main` — the summary is that one of them is worse than the one this PR fixed.

**Same accumulation bug, bigger impact — `GitLabWebhookService` (~line 286)**

It adds *two* `.filter(...)` calls per invocation, and `filter()` accumulates exactly like `headers.add()`:

```java
// DefaultWebClientBuilder
public WebClient.Builder filter(ExchangeFilterFunction filter) {
    initFilters().add(filter);                          // accumulates
}
public WebClient.Builder defaultHeader(String header, String... values) {
    initHeaders().put(header, Arrays.asList(values));   // overwrites
}
```

So after N webhook calls every request runs through 2N logging filters. That's worse than the header case: it's O(n) work per request rather than just bytes on the wire, and each filter iterates and logs every request/response header.

There's a second effect worth noting — the builder is shared, so those filters also apply to the `WebClient` built in `getGitlabProjectId()` (~line 478), which never declared them. That method logs `Authorization` headers it didn't opt into (debug level only, but still).

**Shared-state mutation without growth**

`AzDevOpsTokenService` (~129), `GitLabTokenService` (~51), `PersistentExecutorService` (~73) only use `baseUrl` / `defaultHeader` / `clientConnector`, which overwrite rather than append — so nothing grows unbounded. They do still mutate the injected builder, so each call leaves its `baseUrl` and headers behind for the next one. Harmless today because every caller sets what it needs, but it's fragile: the moment a caller omits one, it silently inherits the previous call's value. `clone()` fixes these uniformly.

**Already correct**

`PublicRegistryProxyService` builds its `WebClient` once in the constructor, so it's fine as is.

For what it's worth, the regression test in this PR is written against the service rather than the builder, so it won't catch these. A test asserting the injected builder is unchanged after a call would generalize better if you want one for the sweep.

PR #3436, opened 2026-08-22
feat: add support for Terragrunt

alfespa17:
Not sure if this feature makes sense so I am leaving this as draft I was just experimenting a little bit

imagen

imagen

imagen

imagen

imagen

This requires https://github.com/terrakube-io/terraform-spring-boot/pull/96

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/json/TerragruntJsonController.java
+++ b/api/src/main/java/io/terrakube/api/plugin/json/TerragruntJsonController.java
@@ -1,0 +1,89 @@
+package io.terrakube.api.plugin.json;
+
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.FileUtils;
+import org.springframework.data.redis.core.RedisTemplate;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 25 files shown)


jakegroves:
i like it especially as it compeltes terraform/tofu as its just a wrapper

alfespa17:
> i like it especially as it compeltes terraform/tofu as its just a wrapper

@jakegroves what I don't understand is what could be the purpose to add it here.

I have never use terragrunt😅,  I was literally testing it and adding the code in the terraform library. 

From what I understand you can use it to trigger several environments at the same time with "run-all" and it basically generates some terraform code that is executed behind the scene by terraform or tofu and it also generates automatically the back-end configuration (Terrakube generates its own backend config).

In my test I was just testing with 1 environment not multiple ones and just for one "workspaces" I guess you can use vanilla terraform for that.

I was thinking that maybe terragrunt can trigger several workspaces at the same time if those workspaces are using "local execution" and terragrunt generate the Terrakube backend configuration.

Maybe use it for something like running `stacks` inside terrakube but I am not really sure 😅

Something like

```terraform
# <root>/terragrunt.hcl
locals {
  # Normalize the relative folder path into a valid workspace name (e.g. "prod-us-east-1-vpc")
  workspace_name = replace(path_relative_to_include(), "/", "-")
}

generate "cloud" {
  path      = "cloud.tf"
  if_exists = "overwrite_terragrunt"
  contents  = <<EOF
terraform {
  cloud {
    organization = "my-org-name"
    hostname     = "my-terrakube.local-instance.com" 

    workspaces {
      name = "${local.workspace_name}"
    }
  }
}
EOF
}
```

> I didnt test the above code  XD it was just an idea

PR #3439, opened 2026-08-24
fix: make webhook delivery and scheduler processing scale to large shared webhooks

### LINKED ISSUE #3438: Webhook delivery and downstream scheduler processing fail to scale on repos with many shared workspaces
### Bug description 🐞

Repos with a GitHub webhook shared across many workspaces (~30 in the reported case) hit several related failures once workspace count or PR size grows:

1. **Webhook delivery times out.** The webhook handler does the entire per-workspace fan-out (fetch PR file changes, evaluate trigger patterns, kick off jobs) synchronously inside the HTTP request GitHub is waiting on. Above a certain matching-workspace count, the request can't finish before GitHub's delivery timeout, and GitHub reports the delivery as failed ("We couldn't deliver this payload: timed out") — the workspaces that should have triggered are silently skipped.
2. **Large/monorepo PRs can miss files.** `GitHubWebhookService`/`AzDevOpsWebhookService` only fetch the first page of the PR's changed-files API. A PR with more changed files than fit on one page silently loses the remaining files, so workspaces whose trigger patterns only match files past the first page never fire. (GitLab is unaffected — it already paginates correctly.)
3. **Redundant per-workspace API calls.** Independent of the two bugs above, every workspace in the fan-out loop re-fetches the same PR file-changes data from the VCS provider individually — a PR matching 30 workspaces makes 30x the necessary API calls, which also makes bug #1 worse.
4. **Scheduler holds DB connections for the full job duration.** Once a job is dispatched, `ScheduleJob` runs the entire terraform execution inside one long-lived transaction, holding a JDBC connection for the whole run. A large shared webhook that dispatches many jobs at once can hold enough connections simultaneously to exhaust the pool, turning a single PR into an availability problem for the rest of the API.
5. **Dead/broken GitHub call in the trigger path.** `GitHubWebhookService` also calls a commit-status endpoint (`/pulls/{n}/statuses`) that isn't a valid GitHub API endpoint and 404s on every invocation, adding failed calls into the same request path as #1.

### Steps to reproduce

1. Configure a shared GitHub webhook on a repo covering a large number of workspaces (~30+).
2. Open a PR that changes files matching the trigger patterns for many of those workspaces at once (bug #1), and/or changes enough files that the PR-files API response spans more than one page (bug #2).
3. Merge/push so the webhook fires.
4. Check the delivery in GitHub → Settings → Webhooks → Recent Deliveries — observe timeout/failure.
5. Separately, watch API logs/DB connection pool usage while several dispatched jobs run concurrently from the same fan-out (bug #4) — observe connections held for the full job duration rather than released promptly.

### Expected behavior

- Webhook delivery should be acknowledged quickly regardless of how many workspaces it matches, and the matching workspaces' runs should start shortly after — matching-workspace count and PR size shouldn't be able to make GitHub give up on a delivery or cause missed triggers.
- All changed files in a PR should be considered for trigger matching, regardless of how many pages the provider's API returns them across.
- A large fan-out shouldn't hold a DB connection per in-flight job for the run's full duration, and shouldn't be able to starve the rest of the API of connections.
- No calls to invalid/nonexistent provider endpoints should be made as part of normal webhook processing.

### Example repository

N/A — reproduction depends on having a shared webhook across a large number of real workspaces; not easily reduced to a minimal public repo

### Anything else?

These are all instances of the same underlying issue — webhook/scheduler processing scales with workspace count and PR size but isn't bounded or made async anywhere in the chain — rather than unrelated bugs, which is why they're filed together. This is a scaling ceiling, not an edge case: it gets worse as more workspaces share a webhook, and there's no automatic retry from Terrakube's side once GitHub gives up on a delivery

jakegroves:
## Why

Repos with a GitHub webhook shared across many workspaces (~30 in the reported case) hit a chain of related scaling failures, filed together as one bug report since they're all instances of the same root problem: webhook/scheduler processing scales with workspace count and PR size, but nothing in the chain is bounded or async.

- Webhook delivery timed out because the entire per-workspace fan-out ran synchronously inside the HTTP request GitHub was waiting on — a scaling ceiling, not something fixable by optimizing the existing code.
- Fixing that required looking closely at what the fan-out does per workspace, which surfaced a real correctness bug (PR file-changes pagination missing files past page one) and a real efficiency bug (every workspace independently re-fetching identical PR data) that were live regardless of the timeout.
- The scheduler has the identical structural problem one layer downstream: each dispatched job holds a DB connection for its entire run, so a large fan-out can hold enough connections simultaneously to exhaust the pool. Fixing webhook delivery without this would just move the scaling ceiling, not remove it.
- Narrowing that transaction boundary is a real behavior change — code that used to run safely inside one long transaction (any lazy Hibernate association was fine to touch anywhere) no longer does. Shipping it with unclosed lazy-loading gaps would trade one production incident for another, so those had to be found and closed as part of this work, not after.

## What

- **Async webhook delivery.** Webhook ingest now validates and persists the delivery (new `RepoWebhookDelivery` entity, `PENDING` status) in a fast transaction and returns immediately; a poller job claims deliveries and does the workspace fan-out out-of-band, with retry/attempt tracking and retention cleanup. Mirrors the existing `notification_outbox` pattern already in the codebase.
- **PR file-changes pagination fixed.** GitHub and Azure DevOps only fetched the first page of PR-changed-files, so large/monorepo PRs could silently miss files past page one and fail to trigger workspaces that should have matched. GitLab was already correct. Also fixed rename handling (`previous_filename`/`oldPath`) so renamed files match on both old and new paths.
- **Redundant per-workspace fetch removed.** PR file-changes are now fetched once per delivery and reused across all matching workspaces, instead of every workspace re-fetching the same data independently.
- **Scheduler scale-hardening.** `ScheduleJob`'s transaction boundary is narrowed so a job's DB connection isn't held for the full terraform run; workspace fan-out runs on a dedicated bounded executor instead of unbounded parallelism; VCS HTTP calls go through a pooled `RestTemplate` with explicit connect/read timeouts; the Quartz thread pool is sized and validated at startup; Micrometer/Prometheus metrics were added for delivery queue depth/age, fan-out, Quartz, and HikariCP so these failure modes are visible going forward.
- **Dead GitHub endpoint removed.** `GitHubWebhookService` called a commit-status endpoint (`/pulls/{n}/statuses`) that isn't a valid GitHub API endpoint and 404'd on every call (with a broken format string besides) — deleted rather than fixed, since it wasn't doing anything useful.
- **Two `LazyInitializationException` gaps closed.** Narrowing the transaction boundary above exposed lazy Hibernate accesses that used to be implicitly safe: `Collection.item` (caught via a production log) and `Job.address` (found by auditing the executor for the same access pattern, since it's read to build terraform `-target`/`-replace` flags). Both fixed by loading through an explicit repository query/`JOIN FETCH` instead of the lazy relationship. `ScheduleJob.execute()` also gained a safety-net catch so any similar gap fails the job cleanly and retries instead of crashing the worker.

## How

- **Outbox/poller:** `RepoWebhookDelivery` table + repository (`claimForDelivery`, `findDueForDispatch`, `reclaimStuckProcessingRows`, `failStuckProcessingRowsAtMaxAttempts`) backing `RepoWebhookDeliveryPollerJob` and a `RepoWebhookDeliveryRetentionJob`; `RepoWebhookService.processV2Webhook` split into `acceptV2Webhook` (fast, `@Transactional`, enqueues) and `processClaimedDelivery` (fan-out).
- **Pagination:** GitHub follows `Link: rel="next"` with `per_page=100`; Azure DevOps loops `$top`/`$skip` using `nextTop`/`nextSkip` from the response body (verified against Microsoft's REST docs), capped at `MAX_PR_FILE_PAGES`.
- **Redundant fetch:** `processClaimedDelivery` fetches PR file changes once via `repoWebhook.getVcs()` before the workspace loop; per-workspace fetch in `processWorkspaceWebhook` becomes a fallback, only used if the pre-fetched result is empty (Azure DevOps excluded, per an existing code comment about per-workspace credentials).
- **Transaction boundary:** `ScheduleJob.execute()` loads the job via `findById()` outside any wrapping transaction rather than running the whole execution inside one long transaction. The Redis per-job execution lock (`EXECUTION_LOCK_PREFIX`) still guards single-worker processing and is unaffected, since Spring Data JPA commits each repository call's own short transaction regardless of the removed wrapping transaction.
- **Bounded fan-out:** workspace fan-out runs via `CompletableFuture.runAsync(...)` on a dedicated `WorkspaceFanoutExecutorConfig` executor instead of the caller's thread/unbounded parallelism.
- **Pooled HTTP client:** VCS calls go through a shared `PoolingHttpClientConnectionManager`-backed `RestTemplate` (`WebhookHttpClientConfig`) with explicit timeouts; `WebhookServiceBase.makeApiRequest` catches `ResourceAccessException` so a slow/hanging call fails fast instead of hanging a thread.
- **Lazy-loading fixes:** `ReferenceRepository.findByWorkspaceWithCollectionItems` (explicit `JOIN FETCH` on `Collection.item`) and `AddressRepository.findByJob` added to replace the two lazy accesses that broke; `ExecutorService` updated to read through both instead of the entity's lazy relationships.

## Testing

- `mvn -pl api -am test` — 889 tests, `BUILD SUCCESS`. New integration tests run against a real Postgres instance (deliberately not `@Transactional`) specifically for the lazy-loading fixes, since that bug class only reproduces without Spring's test-transaction rollback masking it: `RepoWebhookAcceptV2WebhookIntegrationTest`, `ReferenceRepositoryTest`, `AddressRepositoryTest`.
- tested manually against a private monorepo with 35 workspaces

Closes #3438

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -122,6 +122,10 @@
             <artifactId>spring-boot-starter-actuator</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-registry-prometheus</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-security</artifactId>
         </dependency>
--- a/api/src/main/java/io/terrakube/api/plugin/datasource/DataSourceAutoConfiguration.java
+++ b/api/src/main/java/io/terrakube/api/plugin/datasource/DataSourceAutoConfiguration.java
@@ -20,13 +20,22 @@ import javax.sql.DataSource;
 public class DataSourceAutoConfiguration {
 
     @Bean
-    public DataSource getDataSource(DataSourceConfigurationProperties dataSourceConfigurationProperties) {
+    public DataSource getDataSource(DataSourceConfigurationProperties dataSourceConfigurationProperties,
+            io.micrometer.core.instrument.MeterRegistry meterRegistry) {
         HikariConfig config = new HikariConfig();
         config.setMaximumPoolSize(dataSourceConfigurationProperties.getPoolSize());
         config.setMinimumIdle(dataSourceConfigurationProperties.getPoolMinIdle());
         config.setConnectionTimeout(dataSourceConfigurationProperties.getPoolConnectionTimeout());
         config.setIdleTimeout(dataSourceConfigurationProperties.getPoolIdleTimeout());
         config.setMaxLifetime(dataSourceConfigurationProperties.getPoolMaxLifetime());
+        // Exposes hikaricp.connections.active/idle/pending/max/min as Prometheus gauges. This
+        // MicrometerMetricsTrackerFactory ships inside HikariCP itself (com.zaxxer.hikari.metrics.
+        // micrometer), not Micrometer - already on the classpath via the existing HikariCP
+        // dependency, no new pom.xml entry needed for it specifically. Set on the config, not the
+        // constructed HikariDataSource, since the pool starts eagerly inside the
+        // HikariDataSource(config) constructor below.
+        config.setMetricsTrackerFactory(
+                new com.zaxxer.hikari.metrics.micrometer.MicrometerMetricsTrackerFactory(meterRegistry));
 
         switch (dataSourceConfigurationProperties.getType()) {
             case SQL_AZURE:
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ExecutorAvailabilityListener.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ExecutorAvailabilityListener.java
@@ -1,8 +1,11 @@
 package io.terrakube.api.plugin.scheduler;
 
+import java.util.concurrent.atomic.AtomicLong;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
 import io.terrakube.api.repository.JobRepository;
 import jakarta.annotation.PostConstruct;
-import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.quartz.SchedulerException;
 import org.springframework.dao.DataAccessException;
@@ -14,7 +17,6 @@ import org.springframework.stereotype.Component;
 
 // Reacts to the executor module's "capacity available" doorbell by immediately waking the
 // current front of the FIFO dispatch queue, instead of waiting for its next 30s Quartz retry.
-@AllArgsConstructor
 @Component
 @Slf4j
 public class ExecutorAvailabilityListener implements MessageListener {
@@ -23,9 +25,23 @@ public class ExecutorAvailabilityListener implements MessageListener {
     // shared module for this constant.
     public static final String CHANNEL = "terrakube:executor-available";
 
-    RedisMessageListenerContainer redisMessageListenerContainer;
-    JobRepository jobRepository;
-    ScheduleJobService scheduleJobService;
+    private final RedisMessageListenerContainer redisMessageListenerContainer;
+    private final JobRepository jobRepository;
+    private final ScheduleJobService scheduleJobService;
+    // Initialized to "now" at startup, not zero, so the gauge below doesn't report a huge bogus
+    // age before the very first real signal arrives after a fresh deploy.
+    private final AtomicLong lastSignalEpochMillis = new AtomicLong(System.currentTimeMillis());
+
+    public ExecutorAvailabilityListener(RedisMessageListenerContainer redisMessageListenerContainer,
+            JobRepository jobRepository, ScheduleJobService scheduleJobService, MeterRegistry meterRegistry) {
+        this.redisMessageListenerContainer = redisMessageListenerContainer;
+        this.jobRepository = jobRepository;
+        this.scheduleJobService = scheduleJobService;
+        Gauge.builder("executor.availability.age.seconds", lastSignalEpochMillis,
+                        signal -> (System.currentTimeMillis() - signal.get()) / 1000.0)
+                .description("Seconds since the executor module last signalled it has capacity available")
+                .register(meterRegistry);
+    }
 
     @PostConstruct
     public void subscribe() {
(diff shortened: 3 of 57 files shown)


alfespa17:
A couple of comments, can you include these values in application-demo.properties too

```properties
##################
#WEBHOOK DELIVERY#
##################
# Dispatch pool is bounded so a burst of shared-webhook deliveries (e.g. a force-push touching a
# repo with many workspaces) can't spawn unbounded threads competing for the app's DB connection
# pool or the VCS provider's API rate limit. The poller is the durability backstop for the
# immediate-dispatch path - crash recovery, executor-queue-full rejections, and retries whose
# backoff has elapsed - see RepoWebhookDeliveryPollerJob.
io.terrakube.webhook.dispatch.executor.corePoolSize=10
io.terrakube.webhook.dispatch.executor.maxPoolSize=20
io.terrakube.webhook.dispatch.executor.queueCapacity=200
io.terrakube.webhook.poller.batchSize=200
io.terrakube.webhook.poller.stuckProcessingThresholdMinutes=5
io.terrakube.webhook.delivery.retentionDays=30
# Bounds how many workspaces on one shared webhook are processed concurrently - see
# WorkspaceFanoutExecutorConfig. Deliberately small: each workspace's processing itself makes
# outbound VCS/executor HTTP calls, so this is the real ceiling on how many of those run at once
# for a single delivery, not just a thread-pool size.
io.terrakube.webhook.workspace-fanout.concurrency=4
io.terrakube.webhook.workspace-fanout.queue-capacity=200
# Pooled HTTP client for GitHub/GitLab/Azure DevOps API calls (WebhookHttpClientConfig) - replaces
# a fresh RestTemplate + TCP/TLS handshake per call. connectionRequestTimeoutSeconds is how long a
# caller waits for a pooled connection to free up before failing, separate from the connect/response
# timeouts against the remote host itself.
io.terrakube.webhook.dispatch.http.connectTimeoutSeconds=5
io.terrakube.webhook.dispatch.http.responseTimeoutSeconds=30
io.terrakube.webhook.dispatch.http.connectionRequestTimeoutSeconds=2
```

Can you also allow to override the value adding some environment variable, something like:

```properties
io.terrakube.webhook.dispatch.http.connectTimeoutSeconds=${DispatchConnectTimeoutSeconds:disable}
```

This new class will be included in the Elide endpoints by default, maybe you can put some validation that only admins can see the data or maybe just exclude this endpoints, most information there will be just for debugging

```java
public class RepoWebhookDelivery extends GenericAuditFields {
```

jakegroves:
Addressed all three:

- Added the webhook dispatch/poller/HTTP-client properties to `application-demo.properties` as well
- Made every one of those properties overridable via environment variable (e.g. `WebhookDispatchConnectTimeoutSeconds`), same pattern as the rest of the file
- Excluded `RepoWebhookDelivery` from Elide's JSON-API/GraphQL endpoints (`@Exclude`)

alfespa17 (approved):

PR #3442, opened 2026-08-24
fix: remove Quartz job context when job history pruning deletes jobs

### LINKED ISSUE #3441: KEEP_JOB_HISTORY pruning orphans Quartz job contexts - orphaned triggers throw EntityNotFoundException forever
### Bug description

When `KEEP_JOB_HISTORY` pruning deletes a completed job, the job's Quartz job context is not removed with it. Normally a job's own terminal-status scheduler tick removes its context — but under high completion rates, `deleteOldJobs` (triggered by a *later* job completing) can delete the row **before** the pruned job's own cleanup tick has run.

The result is an orphaned Quartz trigger pointing at a row that no longer exists. Every subsequent fire throws:

```
org.quartz.core.JobRunShell: Job DEFAULT.TerrakubeV2_Job_<id> threw an unhandled Exception:
jakarta.persistence.EntityNotFoundException: Unable to find io.terrakube.api.rs.job.Job with id <id>
    at io.terrakube.api.rs.job.Job$HibernateProxy.getCreatedDate(Unknown Source)
    at io.terrakube.api.plugin.scheduler.ScheduleJob.runExecution(...)
```

The exception is thrown on the very first entity access — **before any code path that could deschedule the trigger** — so the orphan refires forever. It can only be removed by manual surgery on the `qrtz_*` tables.

`KEEP_JOB_HISTORY_SOFT_DELETE=true` does not help: the `Job` entity carries `@SQLRestriction("deleted = false")`, so a soft-deleted row is equally invisible to `getReferenceById` and orphans the trigger the same way.

### Reproduction

1. Set `KEEP_JOB_HISTORY` to a small value (e.g. `100`) on a workspace or org.
2. Drive speculative runs at a high sustained rate (we used ~2 runs/sec for 10 minutes, ~870 runs).
3. Once completions outpace the scheduler's cleanup ticks, pruning starts winning the race. In our run, the triggers of **~15% of all jobs** were orphaned (127 of ~870), each then throwing `EntityNotFoundException` every few minutes indefinitely — still firing 30+ minutes after the load stopped, with no way to stop on their own.

Observed on 2.31.2; the relevant code paths on `main` are unchanged (`ScheduleJob.execute` takes no `EntityNotFoundException` handling; `deleteOldJobs` deletes rows without touching the Quartz context).

### Impact

Bounded but permanent and accumulating: each orphan costs an ERROR log entry, a scheduler worker slot, and a DB round-trip every few minutes, forever. Any installation using retention under real load will accumulate them.

### Proposed fix

PR incoming: (1) `deleteOldJobs` removes the pruned job's Quartz context before deleting/soft-deleting the row; (2) `execute()` catches `EntityNotFoundException` and removes the orphaned context by id — which also retroactively self-heals orphans created before the fix, on their next fire.

adeeelsameer:
Fixes #3441

### Problem

`KEEP_JOB_HISTORY` pruning deletes (or soft-deletes) Job rows without removing their Quartz job contexts. Under high completion rates the prune races the pruned job's own terminal-status cleanup tick and can win — leaving a trigger that fires forever against a row that no longer exists, throwing `jakarta.persistence.EntityNotFoundException` on every fire with no code path that can ever clean it up. Soft delete orphans triggers identically, since `@SQLRestriction("deleted = false")` hides the row from `getReferenceById`. Full mechanism, reproduction, and observed impact (~15% of runs orphaned in one saturated 10-minute burst): see the linked issue.

### Fix (two parts)

1. **Prevention** — `deleteOldJobs` now removes the pruned job's Quartz context (`scheduleJobService.deleteJobContext`) before deleting the row, in **both** the hard- and soft-delete branches. When the context was already removed by the job's own terminal tick (the normal case), this is a harmless no-op; a scheduler hiccup is logged and does not abort pruning.
2. **Self-healing** — `ScheduleJob.execute()` catches `EntityNotFoundException`, logs a warning, and removes the orphaned context **by id** (the entity is gone; `removeJobContext` now takes the raw id for exactly this reason). This also retroactively cleans up contexts orphaned before this fix: each heals itself on its next fire instead of erroring forever.

### Tests

- New: `orphanedTriggerSelfDeschedules` — drives `execute()` with a missing Job row and asserts the Quartz context is deleted and no exception propagates.
- Extended: the three existing `completedJobWithHistory*` tests now also verify `deleteJobContext` is invoked for the pruned job (hard delete, workspace-var, and soft-delete variants).
- Full `ScheduleJobTest` suite: 41/41 passing.


alfespa17:
can you help me fixing the merge issues @adeeelsameer

adeeelsameer:
fixed! @alfespa17

alfespa17 (approved):

PR #3445, opened 2026-08-25
fix: schedule job context even when workspace status update fails

### LINKED ISSUE #3444: Job stuck in pending when the CREATE hook fails before scheduling the Quartz context
### Bug description

`JobManageHook` (CREATE, POSTCOMMIT) runs `updateWorkspaceStatus(job)` before `scheduleJobService.createJobContext(job)`, and the hook only catches `ParseException | SchedulerException`. Any `RuntimeException` thrown by the workspace save aborts the hook before the Quartz context is created. Since the job row is already committed, the job sits in `pending` with no steps and no trigger until the inactive-jobs sweep fails it 6 hours later, and every non plan-only job in the workspace queues behind it.

Observed on 2.32.1 with this race: one API replica completed a job and its tick ran the `KEEP_JOB_HISTORY` cleanup (`deleteOldJobs`), deleting an old job of the workspace, while another replica handled a new job POST for the same workspace. The CREATE hook logged `JobCreateHook <id>` and then threw:

```
jakarta.persistence.EntityNotFoundException: Unable to find io.terrakube.api.rs.job.Job with id 29593
```

`Create Job Context TerrakubeV2_Job_<id>` was never logged and the job stayed pending until manually cancelled. Still present on main.

### Expected behavior

A failure to update the workspace last-status label must not prevent the job from being scheduled.

dnl555:
Fixes #3444

In the CREATE path `updateWorkspaceStatus` runs before `createJobContext`, and the hook only catches `ParseException | SchedulerException`. A `RuntimeException` from the workspace save (e.g. `EntityNotFoundException` racing the `KEEP_JOB_HISTORY` cleanup on another replica) aborts the POSTCOMMIT hook, so the committed job never gets a Quartz context and stays in `pending` forever (the UPDATE path is affected the same way).

This wraps the status update in its own try/catch, same approach already used for `createAddresses` in this hook, so scheduling always runs. Unit test included: CREATE with a throwing workspace save still calls `createJobContext`.

### DIFF
--- a/api/src/main/java/io/terrakube/api/rs/hooks/job/JobManageHook.java
+++ b/api/src/main/java/io/terrakube/api/rs/hooks/job/JobManageHook.java
@@ -95,8 +95,14 @@ public class JobManageHook implements LifeCycleHook<Job> {
 
     private void updateWorkspaceStatus(Job job) {
         log.info("Updating last status for workspace {} to {}", job.getWorkspace().getName(), job.getStatus());
-        job.getWorkspace().setLastJobStatus(job.getStatus());
-        job.getWorkspace().setLastJobDate(new Date(System.currentTimeMillis()));
-        workspaceRepository.save(job.getWorkspace());
+        try {
+            job.getWorkspace().setLastJobStatus(job.getStatus());
+            job.getWorkspace().setLastJobDate(new Date(System.currentTimeMillis()));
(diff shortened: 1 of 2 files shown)


alfespa17:
Why did you delete all the other test cases?

PR #3451, opened 2026-08-25
feat(registry): make module download resilient to concurrent load and S3 latency

### LINKED ISSUE #3450: Concurrent terraform init calls intermittently time out downloading modules from the registry (AWS S3 backend)
### Bug description 🐞

Concurrent `terraform init` calls intermittently time out while resolving or downloading a private registry module. The affected sequence: Terraform requests `GET /terraform/modules/v1/{org}/{module}/{provider}/{version}/download`, the registry resolves the module and returns `X-Terraform-Get` pointing at its ZIP endpoint, then Terraform requests that ZIP endpoint and the registry reads the whole artifact from S3 before responding.

Three amplifiers in the AWS storage backend's request path make this worse under load:

- A cold `getModuleVersionPath` cache entry does Terrakube API calls, VCS lookups, and an S3 `HeadObject` — and misses for the same key aren't coalesced, so a burst of requests each independently repeats the same expensive resolution.
- Every ZIP request proxies a full S3 `GetObject` through the registry pod. On failure, `downloadModule` catches the exception and returns an empty byte array — turning an S3 failure into an ambiguous "successful" empty download instead of a clear error.
- The controller increments the download count synchronously before returning the metadata response, adding non-critical bookkeeping latency to every request on the client-visible critical path.

### Steps to reproduce

Run a burst of concurrent `terraform init` operations (e.g. 30-40+) against workspaces that all resolve the same or different modules from the registry, with AWS S3 as the storage backend
Some requests intermittently return a 504/timeout at the ALB, or in the empty-ZIP case, a run fails with a corrupt/unreadable module archive rather than a clear error.

### Expected behavior

- A cached module lookup makes no remote call, and concurrent cold lookups for the same module version make at most one remote resolution attempt per pod.
- A transient S3, API, or VCS failure fails fast and explicitly (never an empty ZIP).
- The registry doesn't proxy module ZIP bytes through the pod when AWS S3 storage is configured.
- Download-count updates never delay Terraform's metadata response.

### Example repository

_No response_

### Anything else?

For AWS deployments, an S3 Gateway VPC endpoint should already be in place as required infrastructure, but it doesn't make unbounded synchronous S3 calls on the request path safe by itself — the fix needs to bound the S3 client's timeouts/retries and get the registry pod out of the ZIP data path (presigned S3 redirect), not just rely on network-layer infra

jakegroves:
## Why

Closes #3450.

Concurrent `terraform init` calls intermittently time out while resolving or downloading a private registry module from AWS S3 storage. Three amplifiers in the request path made this worse under load:

- A cold `getModuleVersionPath` cache entry does Terrakube API calls, VCS lookups, and an S3 `HeadObject` — and misses for the same key weren't coalesced, so a burst of requests each independently repeated the same expensive resolution.
- Every ZIP request proxied a full S3 `GetObject` through the registry pod, and on failure `downloadModule` silently returned an empty byte array instead of an error.
- The controller incremented the download count synchronously before returning the metadata response, adding non-critical bookkeeping latency to every request's critical path.

## What changed

- **Coalesced cache misses**: `ModuleServiceImpl.getModuleVersionPath` now uses `@Cacheable(sync = true)`, so concurrent requests for the same not-yet-cached module version resolve once per pod instead of once per request.
- **Async download-count**: moved off the metadata response's critical path onto a new bounded `downloadCountExecutor`, with retry and backoff on transient failures (mirrors the executor pattern from #3439).
- **Presigned S3 redirects**: `module.zip` now redirects (`302`) to a short-lived presigned S3 URL instead of proxying bytes through the registry pod, behind `presignedRedirectEnabled` (default off — also the rollback switch back to byte-proxying).
- **No more empty ZIPs on failure**: `AwsStorageServiceImpl` now throws `StorageUnavailableException` on S3 failures instead of swallowing them to `new byte[0]`; a new `StorageExceptionHandler` maps that to `503` + `Retry-After: 5`.
- **Bounded S3 client**: `S3Client`/`S3Presigner` now use an explicit Apache HTTP client with configurable connection-pool size and connect/socket/API-call timeouts, plus `RetryMode.STANDARD`, instead of SDK defaults.

## Scope

AWS S3 storage backend only. Azure/GCP/Local get the cache-coalescing and async download-count fixes (storage-agnostic), but not presigned redirects or bounded client timeouts — those backends still proxy bytes through the pod as before. Metrics/observability and LocalStack integration tests are intentionally out of scope for this PR.

## Impact

Measured locally (MinIO + a real AWS S3 bucket) by forcing the exact burst scenario from "Why" — many concurrent requests hitting the same cold module-version cache entry, and heavy concurrent `module.zip` traffic:

| | Before this PR | After this PR |
|---|---|---|
| 20 concurrent requests for the same **uncached** module version | 20 independent Terrakube API/VCS/S3 resolutions | **1** resolution — confirmed via registry logs (single `HeadObject` + git-clone + S3-upload cycle); all 20 callers return together at ~6.2s instead of each paying the full cost |
| Same key, warm | N/A (would already be fast if it got there) | 5–14ms per request |
| `module.zip` at 1000 concurrent requests (MinIO) | Registry proxies every byte; failure mode is a silent empty ZIP | 1000/1000 succeeded via presigned redirect, p95 117ms, zero errors |
| `module.zip` at 1000 concurrent requests (real AWS S3, `eu-west-1`) | — | 998/1000 (99.8%) succeeded end-to-end; **0 errors on the registry side** (1000/1000 redirects issued) — the 2 failures were client-side connection blips on the follow-up leg to S3 |
| S3 outage mid-request | Empty ZIP returned as if successful (`200`) | `503` + `Retry-After: 5` in 169ms (bounded client timeout, not a hang); clean recovery confirmed immediately once storage came back, no lingering bad state |

Cache footprint: `getModuleVersionPath`/`getAvailableVersions` are independently bounded Caffeine caches (`maximumSize(1000)` each, `expireAfterWrite(10min)`) holding short strings/URLs — negligible heap impact (low hundreds of KB at full capacity). `sync = true` coalescing is per-key, so a wide burst across many *distinct* cold keys still uses one Tomcat thread per distinct key for the duration of its resolution — a thread-pool sizing consideration, not a cache-memory one.

## Test plan

- [x] `mvn test -pl registry` — all 39 tests pass (15 new/updated)
- [x] Unit test: concurrent cache misses for the same key invoke the resolver once (`ModuleServiceImplCacheTest`)
- [x] Unit test: cache hit makes no further Terrakube API/S3 calls (`ModuleServiceImplCacheTest`)
- [x] Unit test: download-count update doesn't block the caller and retries transient failures (`ModuleServiceImplAsyncDownloadCountTest`)
- [x] Unit test: `module.zip` redirects when a presigned URL is available, proxies otherwise, and maps storage failure to `503` (`ModuleWebServiceImplTest`)
- [x] Unit test: S3 failures never produce an empty byte array (`AwsStorageServiceImplTest`)
- [x] Manual, local MinIO, `presignedRedirectEnabled=true`:
  - `module.zip` under load: 1000 concurrent requests (100 workers × 10), 1000/1000 `302`→`200`, p95 117ms, zero errors
  - Cache-coalescing on a cold `getModuleVersionPath` entry: 20 concurrent requests against an uncached module version produced exactly one `HeadObject`/git-clone/S3-upload cycle (confirmed via registry logs), not 20 — all 20 callers returned together at ~6.2s
  - Same key warm: 20 concurrent requests, 5–14ms each
  - Injected failure: stopped MinIO mid-flight and hit a fresh (uncached) module version — got `503` + `Retry-After: 5` in 169ms, logged clearly via `StorageExceptionHandler`, no empty-ZIP fallback. Restarted MinIO and confirmed immediate clean recovery
- [x] Manual, real AWS S3 (`eu-west-1`): 1000 concurrent `module.zip` requests via presigned redirect — 998/1000 (99.8%) succeeded end-to-end, p50 139ms / p95 213ms, zero registry-side errors

### DIFF
--- a/registry/pom.xml
+++ b/registry/pom.xml
@@ -167,6 +167,13 @@
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>sts</artifactId>
         </dependency>
+        <!-- Explicit sync HTTP client so S3Client/S3Presigner connection pool, connect, and
+             socket timeouts are configurable (see AwsStorageServiceProperties) instead of relying
+             on whatever the SDK auto-detects on the classpath. -->
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache-client</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.wiremock.integrations</groupId>
             <artifactId>wiremock-spring-boot</artifactId>
--- a/registry/src/main/java/io/terrakube/registry/OpenRegistryApplication.java
+++ b/registry/src/main/java/io/terrakube/registry/OpenRegistryApplication.java
@@ -3,8 +3,10 @@ package io.terrakube.registry;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.scheduling.annotation.EnableAsync;
 
 @EnableCaching
+@EnableAsync
 @SpringBootApplication
 public class OpenRegistryApplication {
 
--- a/registry/src/main/java/io/terrakube/registry/configuration/DownloadCountExecutorConfig.java
+++ b/registry/src/main/java/io/terrakube/registry/configuration/DownloadCountExecutorConfig.java
@@ -1,0 +1,30 @@
+package io.terrakube.registry.configuration;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+@Configuration
+public class DownloadCountExecutorConfig {
+
+    // Bounded, dedicated pool for the download-count bookkeeping call
+    // (ModuleServiceImpl.updateModuleDownloadCount), so it never blocks the Terraform-facing
+    // metadata response (ModuleWebServiceImpl.getModuleVersionPath). Deliberately small: this is
+    // one lightweight, best-effort API call per module download, not a fan-out. A full queue
+    // rejects rather than blocks the caller - a dropped update only affects a non-critical
+    // download counter, never the client-visible response.
+    @Bean("downloadCountExecutor")
+    public ThreadPoolTaskExecutor downloadCountExecutor(
+            @Value("${io.terrakube.registry.download-count.executor.corePoolSize:2}") int corePoolSize,
+            @Value("${io.terrakube.registry.download-count.executor.maxPoolSize:5}") int maxPoolSize,
+            @Value("${io.terrakube.registry.download-count.executor.queueCapacity:200}") int queueCapacity) {
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        executor.setCorePoolSize(corePoolSize);
+        executor.setMaxPoolSize(maxPoolSize);
+        executor.setQueueCapacity(queueCapacity);
+        executor.setThreadNamePrefix("download-count-");
+        executor.initialize();
+        return executor;
+    }
+}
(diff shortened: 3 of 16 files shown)


alfespa17:
I cherry picked your changes and I added GCP and Azure support in the above PR

PR #3454, opened 2026-08-26
fix: misc ui fixes

BaptisteCentreon:
**Description**

Since my previous MR, we started using Terrakube a lot more day to day, and this MR is the result: a batch of small UI fixes on a few pain points.

Nothing big on its own, but together they make the app much nicer to use. It builds on top of the recent UI work (#3339 sidebar, #3410 row links, #3394 session recovery) and extends it where gaps remained.

**Most notable changes**

- **Navigation uses real links everywhere.** #3410 converted the org/workspace rows; this finishes the job. Sidebar, org selector, user menu, tables, lists, run titles, registry buttons, VCS pickers — everything that changes the URL is now a real `<a href>`, so cmd+click / open in new tab work. Post-action redirects untouched.
- **Pages no longer lose or corrupt the organization context.** Workspace create/import now have org-scoped routes (old paths still work) so the sidebar keeps the org. The "create" segment of `/organizations/create` is no longer treated as an org id (it fired API calls failing with 400), and pages without an org context really clear the org state.
- **Switching organization refreshes the content.** Projects and registry only fetched on mount and kept showing the previous org's data.
- **Session expiry sends you back to the login page.** Complements #3394's 401 handling with listeners on the OIDC token-expiry events.

**Other fixes**

- Removed usage of `Steps.Step`, which no longer exists in antd v6 (Publish module, Add VCS, Add template).
- The router was recreated on every `App` render, remounting the whole tree on each token renewal.
- Re-selecting the current org in the selector now navigates instead of doing nothing.
- The org selector now uses antd theme variables instead of hardcoded colors.
- The org name is validated on the edit form like on creation (it is a path segment in the registry protocol).
- Workspace update/delete and the "New workspace" load now surface errors instead of hanging on a spinner.
- Lighter polling, same cadence: no more templates/actions/permissions reload per tick; completed step logs are cached.
- The registry page uses the fluid layout like every other page instead of a 1200px cap.
- State arrays are no longer sorted in place during render (unstable ordering between overview and run list).
- The org permissions hook no longer keeps the previous org's permissions while switching.
- "Sign out"/"Account settings" are a real button and link; destructive Popconfirms have a red confirm button; the tab title follows the page.
- Tests updated accordingly (router wrappers, href assertions).

**Note for reviewers**

The MR is intentionally split into small, self-contained commits (one fix per commit). If you want any specific point split out into its own PR, everything should cherry-pick cleanly.

### DIFF
--- a/ui/src/components/HelpMenu/HelpMenu.tsx
+++ b/ui/src/components/HelpMenu/HelpMenu.tsx
@@ -1,12 +1,11 @@
 import { ApiOutlined, DownOutlined, QuestionCircleOutlined } from "@ant-design/icons";
 import { Dropdown } from "antd";
 import { useState } from "react";
-import { useNavigate } from "react-router-dom";
+import { Link } from "react-router-dom";
 import "./HelpMenu.css";
 
 export const HelpMenu = () => {
   const [isOpen, setIsOpen] = useState(false);
-  const navigate = useNavigate();
 
   const helpItems = [
     {
@@ -26,11 +25,6 @@ export const HelpMenu = () => {
     },
   ];
 
-  const handleApiDocs = () => {
-    setIsOpen(false);
-    navigate("/api-docs");
-  };
-
   return (
     <Dropdown
       trigger={["click"]}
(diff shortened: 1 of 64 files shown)


jakegroves:
I appreciate you building upon the recent changes i've done @BaptisteCentreon the code generally looks sound and brings it closer together
not ran a local copy on your branch yet

would be nice to include this in 2.33 also @alfespa17

BaptisteCentreon:
I've got plenty more where it came from 😅 I have some more factorisation into cleaner components and more optimizations built on top of this but I wanted to avoid an apex pr with a lot of changes.

alfespa17:
> I appreciate you building upon the recent changes i've done @BaptisteCentreon the code generally looks sound and brings it closer together not ran a local copy on your branch yet
> 
> would be nice to include this in 2.33 also @alfespa17

It is a good idea to include it, I will test it later today.

alfespa17 (approved):

BaptisteCentreon:
I have created a follow up pr : https://github.com/terrakube-io/terrakube/pull/3465

PR #3455, opened 2026-08-26
fix: persist job.overrideSource for CLI/API driven Run now jobs

KetchupBomb:
## Problem

Clicking **Run now** on a CLI/API-driven ("remote-content") workspace fails with:

```java
Failed to prepare work dir
java.io.IOException: No configuration tarball URL is set for job <id> (overrideSource is null).
The workspace branch is 'remote-content' but no configuration has been uploaded — upload it
via the terraform CLI or attach a VCS connection before running.
```

...even when the workspace has a successful prior apply and the "Run now" button isn't disabled.

## Root cause

`ExecutorService.execute()` correctly resolves the download source for the job via the `workspace.source` fallback (from #3402), but never writes that resolved value back onto **`job.overrideSource`** — the specific column `SetupWorkspaceImpl.downloadWorkspaceTarGz()` independently re-reads live via the API when the executor actually downloads the tarball, rather than trusting the dispatch payload (the same pattern it already follows for `terraformPlan` and job status elsewhere in that module).

The UI's "Run now" job-creation call (`ui/src/domain/Jobs/Create.tsx`) only ever sets `overrideBranch`, never `overrideSource` — so that column stays `null` for any job created this way, and the executor fails even though the API-side resolution was correct.

<details>
<summary>Why this wasn't caught by #3402's tests</summary>

#3402 added tests asserting `workspace.getSource()` resolves correctly and that the UI button disables when appropriate — both true and both passing. Neither asserted anything about `job.getOverrideSource()`, the column the *executor module* — a separate service — actually reads. The gap sits exactly at that seam, which no test on either side of the module boundary covered.
</details>

## Fix

`ExecutorService.persistJobOverrideSource()` backfills `job.overrideSource` before dispatch whenever it wasn't already set at job creation:

- **No-op for genuine CLI-driven runs** — `RemoteTfeService.createRun()` already sets `overrideSource` correctly at creation, so this never overwrites it.
- **Runs for every remote-content flow type** (plan/apply/destroy) — unlike the apply-only `persistAppliedConfigurationSource`, since *any* flow the executor runs needs a non-null source to download the tarball, not just applies.
- **Excludes VCS/SSH-backed workspaces**, mirroring `persistAppliedConfigurationSource`. The "Run now" branch-name field is free text — a VCS workspace with a mistyped/pasted `"remote-content"` value would otherwise get its **git URL** persisted as `overrideSource`, turning a clear "no configuration uploaded" error into a confusing tarball-download failure instead.

## Test plan

- [x] 6 new unit tests in `ExecutorServiceTest` covering: sets when unset, no-ops when already set (CLI path unaffected), no-ops for non-remote-content branch, no-ops for blank source, no-ops for VCS workspace, no-ops for SSH workspace
- [x] 1 new test in `SetupWorkspaceTest` (executor module) reproducing the original failure end-to-end via `prepareWorkspace()`
- [x] Full `api` module suite: 816 tests, 0 failures, 0 new errors (2 pre-existing/unrelated: a flaky RBAC integration test and an ARM64-emulation Testcontainers limitation — both confirmed present on unpatched `main` too)
- [x] Full `executor` module suite: 187/187 passing
- [x] Patch verified to apply cleanly (`git apply --check`) against current `main`

## Related

Follow-up to #3402.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ExecutorService.java
@@ -26,6 +26,7 @@ import io.terrakube.api.plugin.token.dynamic.DynamicCredentialsService;
 import io.terrakube.api.plugin.vcs.TokenService;
 import io.terrakube.api.repository.AddressRepository;
 import io.terrakube.api.repository.GlobalVarRepository;
+import io.terrakube.api.repository.JobRepository;
 import io.terrakube.api.repository.ReferenceRepository;
 import io.terrakube.api.repository.SshRepository;
 import io.terrakube.api.repository.VariableRepository;
@@ -78,6 +79,8 @@ public class ExecutorService {
     TokenService tokenService;
     @Autowired
     WorkspaceRepository workspaceRepository;
+    @Autowired
+    JobRepository jobRepository;
     @Autowired
     private VariableRepository variableRepository;
     @Autowired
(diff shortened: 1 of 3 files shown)


alfespa17:
I did a quick test using devcontainer and the "Run now" button is now working for CLI driven workflow.

PR #3456, opened 2026-08-27
feat: make the module refresh delays configurable

rtrompier:
A new module version can take up to 15 minutes to appear in the registry. You cannot tune either half of that delay. The Quartz job that reads the Git tags of a module runs every 300 seconds, hardcoded in three places. The registry then caches the version list in process for 10 minutes, also hardcoded.

Both become settings. The defaults are the values in use today:

- `ModuleRefreshInterval` (api) is the interval between two Git tag reads of one module. The default is 300 seconds.
- `ModuleVersionsCacheTtlSeconds` (registry) is the time that the registry keeps a version list. The default is 600 seconds.

If you set both to 30 seconds, the worst case decreases to about one minute. A deployment that keeps the defaults sees no change.

Quartz keeps triggers in the database. As a result, a change to the property is not enough. `ModuleRefreshJob` compares the trigger that fired it against the configured value. That trigger is already in memory, so the comparison costs no I/O. On a mismatch, the job reschedules the trigger through a new `ScheduleServiceBase.rescheduleTask`, which swaps the trigger in one call. The existing `createTask` deletes the job, then creates it again, and the module loses its periodic refresh for good if that fails in between.

The job also becomes `@DisallowConcurrentExecution`. A low interval lets two runs of one module overlap. Both runs then insert the same versions against the unique constraint on `(module_id, version)`.


alfespa17 (approved):
I think this looks good, just keep in mind that using a small value for the refresh can creates problems.

rtrompier:
You are right, and that is why this PR keeps the defaults. A low interval multiplies the git ls-remote calls against the VCS. The PR also adds @DisallowConcurrentExecution to the job for that reason.

Two follow-up changes remove the need for a low interval:
1. A POST `/module/v1/{moduleId}/refresh` endpoint, guarded by the same "team manage module" permission. A release pipeline calls it right after it pushes the tag. The interval can then remain at 300 seconds, and the new version appears immediately.
2. An optional Redis cache (but necessary IMO if you have multiple instances) that holds the module version list for every registry instance. The API can then delete the entry as soon as it imports a version, rather than waiting for the TTL to expire. CAFFEINE stays the default, so an existing deployment sees no change.

The two changes are independent. If you don't want a Redis dependency in the registry module, I can send only the first.

PR #3465, opened 2026-08-28
refactor(ui): components factorisation & settings pages cleanup

BaptisteCentreon:
**Description**

This MR cleans up all the organization and user settings pages. Same layout everywhere, real navigation, and better create/edit forms. The rest of the app will come in later MRs.

**Most notable changes**

- **The Org settings that open as full pages now have a real URL.** Teams, templates, actions, VCS, federated credentials and notifications used local state, they are now routes and the buttons that lead there are real links.
- **All settings pages use the same layout.** Same widths, section headers with a doc link, save/cancel at the bottom right. The team form shows roles as a grid with a color per role.
- **Org and user settings forms were reworked.** Titled sections instead of one long column, two fields per row when they are short, and the help texts moved into tooltips next to the labels.
- **A lot of duplicated pieces were factored into shared components** under `src/components`, grouped by folder.

**Other changes**

- Every org settings page has a help button linking to its documentation.
- The VCS setup guide shows numbered steps with copyable values, and the credential fields sit inside the step they belong to.
- Template cards sit on a grid with uniform heights and a visible hover state.
- Monaco is bundled with the app instead of being downloaded from a CDN on first use, which was very slow on restricted networks.
- The org settings modals moved to their own components instead of inline JSX.
- Delete confirmations use a shared modal, and deleting an organization asks you to type its name.
- Deprecated antd v6 props migrated, hardcoded colors moved to theme variables.


BaptisteCentreon:
Same as last time, cherry-pickable commits if you want to split something, cheers 🙂

BaptisteCentreon:
- Added an error boundary and clean error pages for API 5xx and JS errors.
- Settings pages share one Loading component, no more half rendered pages.
- Fix Navigation buttons using real links

alfespa17 (approved):

PR #3478, opened 2026-08-31
feat(executor): per-workspace terraform data directory cache (TF_DATA_DIR)

TigranKhudav:
## Summary

Every job runs in a throw-away clone, so the modules and providers that `terraform`/`tofu init` downloads are fetched again on every run. The provider plugin cache (`TF_PLUGIN_CACHE_DIR`, documented in *Provider Cache*) only covers providers — nothing caches modules.

This adds an opt-in, per-workspace **Terraform data directory cache** to the executor:

- new property `io.terrakube.executor.terraform.dataDirCacheRoot` (env `TerraformDataDirCacheRoot`), empty by default;
- when set, `loadTempEnvironmentVariables` creates `<root>/<organizationId>/<workspaceId>` and runs every terraform/tofu command of the job with `TF_DATA_DIR` pointing at it, so the next job of the same workspace reuses the downloaded modules **and** installed providers;
- a `TF_DATA_DIR` defined as a workspace environment variable keeps precedence; an unset root, or one that cannot be created, keeps the stock behaviour for that job (logged as a warning).

Nothing changes for deployments that do not set the property.

## Why `TF_DATA_DIR`

Terraform/OpenTofu have no module cache, but `TF_DATA_DIR` relocates `.terraform` (modules, providers, backend config) to any path. Because the executor runs `init` without `-upgrade`, a warm data directory is reused as long as versions still match the constraints. Measured with a workspace using two public-registry modules and the `keycloak/keycloak` provider: `init` went from ~14 s (all downloads) to < 1 s with a warm data dir and a committed lock file, and the executor's outbound traffic to GitHub / the registry for those runs dropped to zero.

Since Terrakube already serialises jobs per workspace, no two jobs write the same data directory; with several executor replicas the volume must be RWX.

## Test plan

- [x] `mvn -pl executor test -Dtest=TerraformExecutorServiceImplTest` — three new tests: data dir is created and exported under `<root>/<org>/<workspace>`, a user-provided `TF_DATA_DIR` is kept, nothing is set without a cache root.
- [ ] Deployed with the matching helm chart change (`executor.cache.enabled=true`): second run of the same workspace shows `Initializing modules...` with no `Downloading …` lines.

Companion PRs: helm chart `executor.cache` block, docs *Module Cache* page.

### DIFF
--- a/executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
+++ b/executor/src/main/java/io/terrakube/executor/service/terraform/TerraformExecutorServiceImpl.java
@@ -1064,12 +1064,43 @@ public class TerraformExecutorServiceImpl implements TerraformExecutor {
             terraformJob.getEnvironmentVariables().put("GOOGLE_APPLICATION_CREDENTIALS", workspaceRootDirectory.getAbsolutePath() + "/terrakube_config_dynamic_credentials.json");
         }
 
+        ensurePluginCacheDir(terraformJob);
         applyDataDirCache(terraformJob);
 
         return terraformJob.getEnvironmentVariables();
     }
 
     static final String TF_DATA_DIR = "TF_DATA_DIR";
+    static final String TF_PLUGIN_CACHE_DIR = "TF_PLUGIN_CACHE_DIR";
+
+    /**
+     * Terraform/OpenTofu refuse a plugin cache directory that does not exist ("the directory must
+     * already exist") and a freshly mounted cache volume is empty, so create it instead of letting
+     * every init report "The specified plugin cache dir ... cannot be opened". The variable can come
+     * from the workspace (job environment) or from the executor's own environment (helm chart).
+     */
+    void ensurePluginCacheDir(TerraformJob terraformJob) {
+        String pluginCacheDir = null;
+        if (terraformJob.getEnvironmentVariables() != null) {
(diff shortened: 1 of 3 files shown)


TigranKhudav:
Companion PRs: helm chart terrakube-io/terrakube-helm-chart#293 (executor.cache block), docs terrakube-io/docs#47 (Module Cache page).

TigranKhudav:
Follow-up commit: the executor now also creates a missing `TF_PLUGIN_CACHE_DIR` (workspace env or executor env) before the first terraform command — Terraform/OpenTofu do not create it and a fresh cache volume is empty, so `init` otherwise warns `The specified plugin cache dir ... cannot be opened` on every job. Covered by `loadTempEnvironmentVariablesCreatesMissingPluginCacheDir`; the 4 cache tests pass on JDK 25.

BenjaminDecreusefond:
Hey ! 

Thank you for your submission ! 

Unless you've tested this PR end to end and can confirm it's working correctly in this case I'm just wrong but, I've been through this questioning myself of being able to share module across remote plans and I was thinking that TF_DATA_DIR could do the trick but I'm not sure. 

A module is not like a provider, some module are allowed to write stuff in your .terraform (it has many implications that could break your states). The fact that you can't treat module the same way you treat provider make me think that processing like this by design is not necessarily a good idea.

This has been running in the tofu community for a while now https://github.com/opentofu/opentofu/issues/1199, they are supposed to implement a solution within tofu that allows to achieve what you wish.

https://github.com/opentofu/opentofu/pull/3252

regards !

PR #3482, opened 2026-08-31
fix(registry): avoid NPE when module_version.git_tag is null (#1)

BenjaminDecreusefond:
getModuleVersionPath mapped the version stream to getGitTag() before calling findFirst(), so Optional.of(null) threw an NPE whenever a matching module_version row had a null git_tag. That column is nullable and CanonicalModuleVersionMigration intentionally leaves it null for modules it skips (invalid/unaliased provider or version collisions), so any terraform/tofu init against one of those modules returned a 500.

Reorder findFirst()/map() so the null-tolerant map runs on the resolved row, and fall back to the semver version string (matching the fallback already used by CanonicalModuleVersionMigration and ModuleRefreshJob) instead of null.

Adds a regression test covering the null git_tag case.


alfespa17:
Fixed in #3485

PR #3489, opened 2026-09-01
fix: serialize dates with second precision instead of truncating to the minute

### LINKED ISSUE #2787: Console calculates time from top of the minute, not by second

7487:
## Problem

The console computes relative times ("Triggered from UI 42 seconds ago", "Last run X ago") from timestamps the API truncates to the top of the minute, so a job triggered late in the minute immediately shows 40-50+ seconds of elapsed time.

## Root cause

Elide's default ISO8601 date serde uses the pattern `yyyy-MM-dd'T'HH:mm'Z'` (see `serdesBuilder()` in Elide's `ElideAutoConfiguration`), so every `java.util.Date` attribute served through the serde loses its seconds. Reproduced against current `main` with the new test in this PR: GraphQL returns `lastJobDate` as `2026-01-15T10:20Z` for a value stored as `2026-01-15T10:20:42Z`.

Two serialization paths are affected differently on current `main`:

- **GraphQL** (`GraphQLScalars` → `CoerceUtil.lookup(Date.class)` → serde): still truncates to the minute. This is the path the workspace list reads `lastJobDate` from ("Last run X ago").
- **JSON:API**: was truncating when #2787 was filed, but the Spring Boot 4 / Jackson 3 upgrade (#3470) changed this path to Jackson's default `Date` serialization, which already keeps milliseconds. It no longer truncates today.

This also explains why #2818/#2819 didn't pan out: the truncation never happened in the database or in Spring's Jackson config, it happens in Elide's serde.

## Fix

Register a `SerdesBuilderCustomizer` (Elide's intended extension point for this) that overrides the date pattern to `yyyy-MM-dd'T'HH:mm:ss'Z'`.

Note on inputs: the serde also parses date strings (GraphQL date arguments, JSON:API date attributes/filters), which now expect seconds as well. Nothing in the UI or executor sends date literals to the API, and the previous minute-only pattern rejected full ISO timestamps anyway, so accepting the standard form is strictly more useful.

## Tests

- `graphQlDatesKeepSecondPrecision`: stores a `lastJobDate` with non-zero seconds and asserts GraphQL (same organization → workspace query shape the UI uses) returns it unchanged. Fails on `main` with `2026-01-15T10:20Z`.
- `jsonApiDatesKeepSecondPrecision`: pins job `createdDate`/`updatedDate` in JSON:API responses to at least second precision so the incidental Jackson 3 behavior can't silently regress to minutes.

Both pass with `mvn -pl api test`.

Fixes #2787

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/elide/ElideSerdeConfiguration.java
+++ b/api/src/main/java/io/terrakube/api/plugin/elide/ElideSerdeConfiguration.java
@@ -1,0 +1,23 @@
+package io.terrakube.api.plugin.elide;
+
+import com.yahoo.elide.SerdesBuilderCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import java.util.TimeZone;
+
+@Configuration
+public class ElideSerdeConfiguration {
+
+    /**
+     * Elide's default date serde uses minute precision (yyyy-MM-dd'T'HH:mm'Z'), which
+     * truncates the seconds of fields like createdDate and lastJobDate in GraphQL
+     * responses. The UI then computes relative times from the top of the minute and
+     * shows things like "42 seconds ago" for a job that was just triggered (issue
+     * #2787). Serialize with second precision instead.
+     */
+    @Bean
+    public SerdesBuilderCustomizer secondPrecisionDateSerdes() {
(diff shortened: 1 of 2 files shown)


alfespa17:
thank you for the idea, I have create another [PR](https://github.com/terrakube-io/terrakube/pull/3500) with a more robust implementation

PR #3514, opened 2026-09-03
feat(workspaces): add GraphQL cursor pagination

Macbet:
## What

- Add the `workspacePage` GraphQL query with opaque cursor pagination.
- Move workspace search, status/tag/project filters, and sorting to the server.
- Preserve datastore-level workspace authorization, including federated teams.
- Update the organization workspaces UI to consume server pages.

## Verification

- `container`: targeted UI Jest tests — 22 passed.
- `container`: targeted ESLint.
- `container`: `DexGroupServiceTests` — 12 passed; `FederatedLookupServiceTests` — 10 passed.
- `WorkspaceGraphQlPaginationTest` passed before the rebase; its source was unchanged while resolving conflicts.


alfespa17:
https://elide.io/pages/guide/v7/11-graphql.html

You don't have to create a custom controller to manage that kind of logic

Yahoo Elide already supports pagination, sorting, RSQL filtering, and security checks natively.

Macbet:
@alfespa17 thx for sharing, update code ✅

alfespa17:
### Review: Paginated Workspace List (`perf/paginated-workspace-list`)

Thank you @Macbet for working on this! Transitioning the workspace list from full client-side fetching to server-side pagination with native Elide GraphQL is a major performance improvement. For organizations with hundreds or thousands of workspaces, this dramatically cuts network payload size, browser memory overhead, and database serialization costs.

Below is a detailed analysis of the changes, covering critical security edge cases, database query characteristics, frontend UX regressions, and recommended improvements.

---

### 1. 🌟 Key Highlights & Architectural Strengths

* **Compliance with Terrakube Architecture Rule 1:**
  Reverting custom Spring GraphQL controllers and criteria queries in favor of Elide's native `/graphql/api/v1` endpoint is the right architectural decision. Spring GraphQL is strictly reserved for WebSocket subscriptions.
* **Database-Level RBAC Pushdown (`FilterExpressionCheck`):**
  Replacing in-memory `OperationCheck` annotations (`TeamViewWorkspace`, `TeamLimitedViewWorkspace`, `TeamProjectLimitedViewWorkspace`) with `WorkspaceReadFilter` (extending `FilterExpressionCheck<Workspace>`) enables Elide to push authorization predicates directly into the SQL `WHERE` clause. This allows SQL `LIMIT`, `OFFSET`, and `COUNT` to execute accurately at the database level.
* **Independent Project Discovery:**
  Using `projectService.listProjects(id)` to populate the project filter dropdown prevents the dropdown from being limited to only the projects present on the current page of workspaces.
* **Debounced Search & Page Reset:**
  Properly debouncing the search input (300ms) and resetting `page` to `1` on any filter/sort change prevents inconsistent page offsets.

---

### 2. 🚨 Critical & High-Priority Issues

#### A. Empty Groups Causing `IllegalStateException` in `WorkspaceReadFilter`
In `WorkspaceReadFilter.java`:
```java
@Override
public FilterExpression getFilterExpression(Type<?> entityClass, RequestScope requestScope) {
    Object[] groups = groupService.getEffectiveGroups(requestScope.getUser()).toArray();

    FilterExpression organizationAccess = new AndFilterExpression(
            new InPredicate(path(entityClass, requestScope, "organization.team.name"), groups),
            ...
```

* **Problem:** If a user or service token has no groups (e.g., an empty `groups: []` claim or no groups assigned), `groups` is an empty array (`new Object[0]`).
* When Elide's `CaseAwareJPQLGenerator` tries to build the HQL for `new InPredicate(..., new Object[0])`, it calls `Preconditions.checkState(!params.isEmpty())` internally. With an empty parameter list this throws an `IllegalStateException` **before any SQL is emitted** — the request fails with a 500 rather than a clean 403/empty result.
* Furthermore, if the user has no groups, they cannot match any team, project, or workspace access rule. Executing 3 entity joins for a user with zero groups is wasteful.
* **Fix:** Short-circuit with a `FalsePredicate` immediately when `groups.isEmpty()`. `FalsePredicate` is already available in `elide-core` and uses `Operator.FALSE` with an empty parameter list, so it generates no SQL parameters at all:
  ```java
  Set<String> groups = groupService.getEffectiveGroups(requestScope.getUser());
  if (groups == null || groups.isEmpty()) {
      return new FalsePredicate(path(entityClass, requestScope, "id"));
  }
  Object[] groupArray = groups.toArray();
  ```

---

#### B. Missing Null Guards in `WorkspaceReadFilter.applyPredicateToObject`
In `WorkspaceReadFilter.java`:
```java
@Override
public boolean applyPredicateToObject(
        Workspace workspace, FilterPredicate predicate, RequestScope requestScope) {
    Set<String> groups = groupService.getEffectiveGroups(requestScope.getUser());
    boolean organizationAccess = workspace.getOrganization().getTeam().stream().anyMatch(team ->
            groups.contains(team.getName())
                    && (workspace.getProject() == null || rbacService.canManageWorkspace(team)));
    boolean projectAccess = workspace.getProject() != null
            && workspace.getProject().getProjectAccess().stream()
                    .anyMatch(access -> groups.contains(access.getName()));
    boolean workspaceAccess = workspace.getAccess().stream()
            .anyMatch(access -> groups.contains(access.getName()) && rbacService.canManageWorkspace(access));
    return organizationAccess || projectAccess || workspaceAccess;
}
```

* **Context:** `applyPredicateToObject` is called for single-resource access checks (e.g. `GET /api/v1/workspace/{id}` or relationship mutations). Under normal JPA/Hibernate usage the practical NPE risk is low:
  - `workspace.getOrganization()` has `@ManyToOne @NotNull` — Elide always loads it for a persisted workspace.
  - `workspace.getProject().getProjectAccess()` is **already null-guarded** by the `workspace.getProject() != null &&` short-circuit on the preceding line.
  - `workspace.getAccess()` and `getTeam()` are `@OneToMany` collections — Hibernate initializes these as empty `PersistentBag` instances, not `null`.
* **However**, if the workspace object is constructed manually (e.g. in unit tests or by a future refactor) without going through the JPA lifecycle, `.stream()` on an uninitialised `null` collection will throw `NullPointerException`. Defensive null guards eliminate this fragility and make the code robust against lazy-loading edge cases.
* **Fix:** Add null guards and an early-exit for the empty-groups case:
  ```java
  @Override
  public boolean applyPredicateToObject(
          Workspace workspace, FilterPredicate predicate, RequestScope requestScope) {
      Set<String> groups = groupService.getEffectiveGroups(requestScope.getUser());
      if (groups == null || groups.isEmpty()) {
          return false;
      }

      boolean organizationAccess = workspace.getOrganization() != null
              && workspace.getOrganization().getTeam() != null
              && workspace.getOrganization().getTeam().stream().anyMatch(team ->
                      groups.contains(team.getName())
                              && (workspace.getProject() == null || rbacService.canManageWorkspace(team)));

      boolean projectAccess = workspace.getProject() != null
              && workspace.getProject().getProjectAccess() != null
              && workspace.getProject().getProjectAccess().stream()
                      .anyMatch(access -> groups.contains(access.getName()));

      boolean workspaceAccess = workspace.getAccess() != null
              && workspace.getAccess().stream()
                      .anyMatch(access -> groups.contains(access.getName()) && rbacService.canManageWorkspace(access));

      return organizationAccess || projectAccess || workspaceAccess;
  }
  ```

---

#### C. Token Group Array Types in `DexGroupServiceImpl.getEffectiveGroups`
In `DexGroupServiceImpl.java`:
```java
Object tokenGroups = tokenAttributes.get("groups");
if (tokenGroups instanceof String group && !group.isBlank()) {
    groups.add(group);
} else if (tokenGroups instanceof Collection<?> values) {
    values.stream()
            .filter(Objects::nonNull)
            .map(Object::toString)
            .filter(group -> !group.isBlank())
            .forEach(groups::add);
}
```

* **Problem:** Spring Security's `NimbusJwtDecoder` deserializes JWT claim arrays via Jackson, which always produces `ArrayList<Object>` — matched correctly by the `instanceof Collection<?>` branch. However, the older `toStringArray()` helper (still present but no longer called) handled `Object[]` / `String[]` explicitly. If a mock token in tests or a non-standard OIDC provider populates `groups` as a Java array rather than a `Collection`, `getEffectiveGroups` silently returns an empty set, causing the workspace filter to short-circuit with no visible error.
* **Severity:** Low — not triggered by any standard Spring Security JWT path. Defensive fix recommended to protect test fidelity and future flexibility.
* **Fix:** Support `Object[]` alongside `Collection<?>`:
  ```java
  } else if (tokenGroups instanceof Object[] values) {
      Arrays.stream(values)
              .filter(Objects::nonNull)
              .map(Object::toString)
              .filter(group -> !group.isBlank())
              .forEach(groups::add);
  }
  ```

---

### 3. ⚠️ UI & UX Observations

#### A. "Group by Project" Anomaly with Server-Side Pagination
* **Behavior:**
  * `groupByProject` is enabled by default in user preferences (`localStorage.getItem("groupByProject") !== "false"`).
  * In `OrganizationDetailsPage.tsx`, `groups` is computed over `workspaces` (which only contains the current page's 20 items):
    ```tsx
    const groups = useMemo(() => {
      const map = new Map<string, { key: string; label: string; items: WorkspaceListItem[] }>();
      for (const workspace of workspaces) { ... }
    ```
  * Workspaces are ordered by the server-side sort (e.g. `lastJobDate` or `name`), not grouped by project on the server.
* **Resulting UX issue:**
  1. If Project A has 30 workspaces, 12 might appear on Page 1, 10 on Page 2, and 8 on Page 3.
  2. The group header count `{group.items.length} workspaces` will say "12 workspaces" on Page 1 and "10 workspaces" on Page 2, misleading users about the actual number of workspaces in that project.
  3. `WorkspaceTable` additionally applies `GROUP_PREVIEW_SIZE = 5` with a "Show more" toggle within the 20-item page, creating a double-pagination experience (preview truncation inside page truncation).
* **Suggested Fix:**
  * When `groupByProject` is active, either:
    - Automatically sort by project on the server (`sort: "project.name,name"`), or
    - Group by project at the organization level, or
    - If grouping remains per-page, clarify the label (e.g., "{group.items.length} on this page") and consider setting `GROUP_PREVIEW_SIZE >= pageSize` so items within a single page are not truncated twice.

---

#### B. Status Sort Order Regression
* **Behavior:**
  * Previously, `sortWorkspaces(workspaces, "status")` in `workspaceSort.ts` sorted by custom operational urgency (`STATUS_ORDER`):
    `Running` → `Queue` → `WaitingApproval` → `Failed` → ... → `Completed` → `NeverExecuted`.
  * In `workspaceService.ts`, status sorting is now delegated to the database:
    ```ts
    status: "lastJobStatus,id",
    ```
  * In PostgreSQL/JPA, sorting by `lastJobStatus` orders alphabetically (or by enum ordinal). Alphabetical order places `approved` and `cancelled` first, while critical statuses like `running` and `waitingApproval` are pushed to the very bottom.
* **Suggested Note:**
  * Since Elide does not support arbitrary custom sort mappings in RSQL without database-level schema changes (like an ordinal column or CASE expression), this trade-off should be documented, or the status sort option adjusted.

---

#### C. 8 Count Subqueries Executing on Every Poll and Search Keystroke
* **Behavior:**
  In `workspaceService.ts`, `listWorkspacePage` sends:
  ```graphql
  query WorkspacePage(...) {
    organization(ids: $organizationIds) {
      edges {
        node {
          workspace(...) { ... }
          all: workspace(first: "1", filter: $allFilter) { pageInfo { totalRecords } }
          waitingApproval: workspace(first: "1", filter: $waitingApprovalFilter) { pageInfo { totalRecords } }
          failed: workspace(first: "1", filter: $failedFilter) { pageInfo { totalRecords } }
          pending: workspace(first: "1", filter: $pendingFilter) { pageInfo { totalRecords } }
          queue: workspace(first: "1", filter: $queueFilter) { pageInfo { totalRecords } }
          running: workspace(first: "1", filter: $runningFilter) { pageInfo { totalRecords } }
          completed: workspace(first: "1", filter: $completedFilter) { pageInfo { totalRecords } }
          neverExecuted: workspace(first: "1", filter: $neverExecutedFilter) { pageInfo { totalRecords } }
        }
      }
    }
  }
  ```
  Each alias translates into an independent SQL `COUNT` query with joins on `organization`, `team`, `project`, and `access`.
  With the 10-second polling interval in `OrganizationDetailsPage.tsx`, every open browser tab runs **9 database queries every 10 seconds**.
* **Suggested Optimization:**
  * During background polling (via `usePolling`), only fetch the active page of workspaces, rather than re-evaluating all 8 status counters on every tick.
  * Re-fetch status counts only when filters, tags, or search terms change.

---

### 4. 🧹 Hygiene & Cleanup

#### A. Unrelated Commit Bundled into the Branch
* Commit `22d26c28a` (`fix(context): evict completed empty reads`) modifies `ContextReadService.java`:
  ```java
  } else {
      inFlight.remove(jobId, future);
  }
  ```
* This change relates to executor context caching rather than workspace pagination. It should be separated into its own PR to keep git history clean.

#### B. Dead Code in `workspaceFilter.ts` and `workspaceSort.ts`
* `filterWorkspaces()` and `sortWorkspaces()` are no longer invoked by any UI component. Only their corresponding unit tests (`workspaceFilter.spec.ts` and `workspaceSort.spec.ts`) reference them.
* Consider marking them deprecated or removing them to avoid confusion for future contributors.

---

### 5. Summary of Recommended Action Items

1. **[Backend]** In `WorkspaceReadFilter.getFilterExpression()`, return `new FalsePredicate(...)` when `groups.isEmpty()`.
2. **[Backend]** In `WorkspaceReadFilter.applyPredicateToObject()`, add null checks for `workspace.getOrganization()`, `getTeam()`, `getProject()`, `getProjectAccess()`, and `getAccess()`.
3. **[Backend]** In `DexGroupServiceImpl.getEffectiveGroups()`, handle `tokenGroups instanceof Object[]`.
4. **[Backend]** Rebase or cherry-pick commit `22d26c28a` (`ContextReadService`) into a dedicated PR.
5. **[Frontend]** Review the UX of `groupByProject` when paginated (avoid double-truncation with `GROUP_PREVIEW_SIZE`).
6. **[Frontend]** Optimize polling in `OrganizationDetailsPage.tsx` so 8 count subqueries are not executed every 10 seconds.

Macbet:
@alfespa17 thx for review, fixed

alfespa17:
I was testing, can you double check

I should see 4 workspaces.

imagen

But once I click "simple" org I dont see anything

imagen

Macbet:
@alfespa17 fixed ✅

alfespa17:
I got the same 😅, I am moving the PR to draft for now 

imagen

Macbet:
@alfespa17 found the cause of the empty list, thanks for the screenshot.

In your second screenshot the project filter shows **demo-project** selected. That project has no workspaces in the demo data, so the server-side filter correctly returns nothing. The filter value lives in `sessionStorage` and used to survive switching organizations; before this PR the project dropdown was built from the loaded workspaces, so an empty project could never be picked and the stale value was invisible. Now the page drops a project filter that doesn't belong to the current organization (covered by `OrganizationDetailsPage.test.tsx`).

Also in this push:

- Merged `main`. The new policy compliance filter from the OPA work is now server-side too (`policyComplianceStatus` RSQL filter + aliased count queries), same pattern as the status counts.
- Fixed a permission gap in `WorkspaceReadFilter`: Elide renders `TruePredicate` as `(1 = 1)` regardless of its path, so the custom-role `manageWorkspace` check was a no-op in SQL. Replaced with an explicit `InPredicate`. Added `customRoleTeamWithoutManageWorkspaceCannotReadProjectWorkspaces`, which fails without the fix.

Verification: `WorkspaceGraphQlPaginationTest` (6) and `WorkspaceReadFilterTest` pass on JDK 25; UI jest 201 passed, eslint/tsc clean.

Could you re-test with the project filter set to "All projects"? Moved the PR out of draft so CI runs.

alfespa17:
Sure, I will re test later 🙂, I didn't notice I had the filter 😅

alfespa17:
Now it is working 🙂

imagen

 
imagen

alfespa17:
I was running some analysis and I found the following:

## Critical Security & RBAC Issues

### 🔴 Defect 1: Workspace-Level `read` and `plan` Roles Are Locked Out
In `WorkspaceReadFilter.java`:
```java
FilterExpression workspaceAccess = new AndFilterExpression(
        new InPredicate(path(entityClass, requestScope, "access.name"), groups),
        canManageWorkspace(entityClass, requestScope, "access"));
```
And in `canManageWorkspace`:
```java
return new OrFilterExpression(
        new InInsensitivePredicate(role, "admin", "write"),
        new AndFilterExpression(
                customRole,
                new InPredicate(path(entityClass, requestScope, relation + ".manageWorkspace"), true)));
```
**Impact:**
- In Terrakube's 3-tier RBAC system (`Team` -> `ProjectAccess` -> `Access`), a team granted `role = "read"` or `role = "plan"` directly on a workspace via `Access` will fail `canManageWorkspace`.
- Consequently, users granted read-only or planner access to a workspace **cannot read or list the workspace**. They will receive an empty list or 404.
- **Severe CI/CD & Federated Credential Impact**: Automation workflows using **Federated Credentials** (e.g. GitHub Actions, GitLab CI) are standardly assigned least-privilege teams with `role = "plan"` (for pull request speculative plan validation) or `role = "read"` (for monitoring workspace status). Under this PR, all such federated CI/CD pipelines targeting workspaces via workspace-level `Access` are completely locked out.
- In `ProjectAccess`, any assigned team can view workspaces (`new InPredicate(path(..., "project.projectAccess.name"), groups)`). Workspace-level `Access` should similarly allow viewing for any assigned team, or check `canViewWorkspace`.

**Recommended Fix:**
Allow any team with an `Access` record to read the workspace (mirroring `ProjectAccess`), or allow roles `"read"`, `"plan"`, `"write"`, `"admin"`:
```java
FilterExpression workspaceAccess =
        new InPredicate(path(entityClass, requestScope, "access.name"), groups);
```

---

### 🔴 Defect 2: Internal Service Accounts (`TerrakubeInternal`) Denied by `WorkspaceReadFilter`
In `DexGroupServiceImpl.java`:
- `isServiceMember` grants automatic group membership to internal tokens:
  ```java
  boolean isMember = principal.getTokenAttributes().get("iss").equals("TerrakubeInternal");
  ```
- However, `getEffectiveGroups(User user)` only parses token claims and federated lookups:
  ```java
  public Set<String> getEffectiveGroups(User user) {
      JwtAuthenticationToken principal = (JwtAuthenticationToken) user.getPrincipal();
      Map<String, Object> tokenAttributes = principal.getTokenAttributes();
      Set<String> groups = new LinkedHashSet<>();
      ...
  ```
**Impact:**
- Internal service calls (e.g. executor, internal background hooks) using `TerrakubeInternal` tokens do not carry explicit `"groups"` claims.
- `getEffectiveGroups` returns an empty set `[]`.
- In `WorkspaceReadFilter.java`:
  ```java
  if (groups.length == 0) {
      return new FalsePredicate(path(entityClass, requestScope, "id"));
  }
  ```
- Internal service accounts are denied datastore-level read access unless they match `user is a superuser`.

**Scope & Clarification on Federated Credentials:**
- **Federated Credentials are NOT affected by Defect 2**: `getEffectiveGroups()` explicitly resolves federated tokens through `federatedLookupService.findAllAuthorized(tokenAttributes)` and maps them to `Federated::getName` (verified in `DexGroupServiceTests#federatedTokenWithoutGroupsClaimIsMember`). Thus, federated credentials successfully populate their group names and avoid `FalsePredicate`.
- **`TerrakubeInternal` IS uniquely broken**: Because internal service accounts rely on a hardcoded issuer check (`iss == "TerrakubeInternal"`) rather than database lookups or group claims, omitting internal token handling from `getEffectiveGroups()` leaves internal service tokens completely without groups.

**Recommended Fix:**
In `DexGroupServiceImpl.getEffectiveGroups`, or in `WorkspaceReadFilter`: if `authenticatedUser.isServiceAccountInternal(requestScope.getUser())`, return an allow-all predicate or include all organization teams.

---

### 🟠 Defect 3: Inconsistency in Group Claim Parsing Across `DexGroupServiceImpl`
`getEffectiveGroups` was improved to handle `Object[]`, single `String`, and `Collection<?>`:
```java
Object tokenGroups = tokenAttributes.get("groups");
if (tokenGroups instanceof Object[] values) {
    tokenGroups = Arrays.asList(values);
}
if (tokenGroups instanceof String group && !group.isBlank()) {
    groups.add(group);
} else if (tokenGroups instanceof Collection<?> values) { ... }
```
However, `isMember(User user, String group)` and `isServiceMember(User user, String group)` still only check:
```java
if (tokenGroups instanceof Collection<?> values)
```
**Impact:** If a token provider supplies the `groups` claim as an array or a comma-separated string, `getEffectiveGroups` will find the group, but subsequent `isMember` checks (e.g. in operation checks or hooks) will return `false`.

**Recommended Fix:** Refactor `isMember` and `isServiceMember` to delegate to `getEffectiveGroups(user).contains(group)`.

---

## 3. Performance & Scalability Concerns

### ⚠️ Issue 4: N+1 Subquery Explosion (14 Database Queries per Request)
In `workspaceService.ts`:
```typescript
const countFilters: Record<string, string | undefined> = includeStatusCounts
  ? {
      statusAll: workspaceFilter(request, { status: "All" }),
      ...Object.fromEntries(
        STATUS_COUNT_KEYS.map((key) => [`status_${key}`, workspaceFilter(request, { status: key })])
      ),
      policyAll: workspaceFilter(request, { policyStatus: "All" }),
      ...Object.fromEntries(
        POLICY_COUNT_KEYS.map((key) => [`policy_${key}`, workspaceFilter(request, { policyStatus: key })])
      ),
    }
  : {};
```
**Impact:**
- Every page load with counts executes:
  - 1 main query for the workspace page
  - 8 status count subqueries (`statusAll`, `waitingApproval`, `failed`, `pending`, `queue`, `running`, `completed`, `NeverExecuted`)
  - 5 policy count subqueries (`policyAll`, `COMPLIANT`, `NON_COMPLIANT`, `EXEMPTED`, `UNKNOWN`)
- **Total: 14 queries per navigation action**.
- In Elide GraphQL, each alias triggers an independent `SELECT COUNT(DISTINCT ...)` query joining `workspace`, `organization`, `team`, `project`, and `access`.
- On page turns (e.g. navigating from Page 1 to Page 2), `includeStatusCounts` is `true` by default in `fetchPage()`, causing all 13 count queries to re-execute even though the totals have not changed.

**Recommended Fix:**
1. In `OrganizationDetailsPage.tsx`, pass `includeStatusCounts = false` when changing pages via `handlePageChange` (only reload counts when filters, project, or search query changes).
2. Consider consolidating status/policy counts into a dedicated aggregation query or caching facet totals.

---

### ⚠️ Issue 5: Missing Database Indexes on `workspace` Table
The paginated queries and facet counts filter and sort by:
- `organization_id`
- `last_job_status`
- `last_job_date`
- `project_id`
- `policy_compliance_status`

Currently, `workspace` has foreign key constraints but lacks composite performance indexes for these filtered/sorted columns.

**Recommended Fix:**
Add a Liquibase migration (`changelog-2.35.0-workspace-pagination-indexes.xml`):
```xml
<changeSet id="2-35-0-workspace-indexes" author="terrakube">
    <createIndex tableName="workspace" indexName="idx_workspace_org_status">
        <column name="organization_id"/>
        <column name="deleted"/>
        <column name="last_job_status"/>
    </createIndex>
    <createIndex tableName="workspace" indexName="idx_workspace_org_date">
        <column name="organization_id"/>
        <column name="deleted"/>
        <column name="last_job_date"/>
    </createIndex>
    <createIndex tableName="workspace" indexName="idx_workspace_org_project">
        <column name="organization_id"/>
        <column name="deleted"/>
        <column name="project_id"/>
    </createIndex>
</changeSet>
```

---

## 4. UI / UX & Functional Issues

### ⚠️ Issue 6: Project Grouping Fragmentation Across Pages
In `OrganizationDetailsPage.tsx`:
```typescript
const showGrouped = listViewMode === "compact" && filterState.groupByProject && filterState.projectId === null;
```
Workspaces are fetched page-by-page (sorted by name or last run). The grouping logic in `groups` groups only the 20 items returned on the active page.

**Impact:**
- Workspaces belonging to the same project are split across different pages.
- `WorkspaceTable.tsx` displays `{group.items.length} workspaces on this page`.
- Users cannot see all workspaces for a given project under a single collapsible group.

**Recommended Fix:**
- When `groupByProject` is enabled with server-side pagination, either:
  1. Add server-side sorting by `project.name, name, id` so items belonging to the same project appear sequentially.
  2. Clearly communicate in the UI header that grouping applies per page, or disable pagination / expand page size when grouping is active.

---

### ⚠️ Issue 7: Silent Error Swallowing After Initial Load
In `OrganizationDetailsPage.tsx`:
```typescript
if (response.isError || !response.data) {
  if (!loadedOnce.current) {
    setError({
      title: response.error?.status || "Failed to load workspaces",
      message: response.error?.message,
    });
  }
  setLoading(false);
  return;
}
```
**Impact:**
- If an error occurs on a subsequent search, filter switch, or page navigation, `!loadedOnce.current` is `false`.
- The error is silently swallowed. `setError` is never called.
- The UI retains stale data and gives the user no feedback that the search or filter failed.

**Recommended Fix:**
Use an Ant Design notification or toast (`message.error(response.error?.message)`) for subsequent errors, or set a transient error state.

---

### ⚠️ Issue 8: Job Status Sorting Behavior
In `workspaceSort.ts`:
- Client-side sorting previously sorted by execution urgency (`Running` -> `Queue` -> `WaitingApproval` -> `Failed` -> ...).
- Server-side sorting maps `"status"` to `"lastJobStatus,id"`:
  ```typescript
  status: "lastJobStatus,id",
  ```
- Because `last_job_status` is stored as a varchar/string in PostgreSQL, the sort order is alphabetical (`cancelled`, `completed`, `failed`, `pending`, `queue`, `running`, `waitingApproval`).
- While this is noted with the UI label `"Job status (grouped)"`, users accustomed to seeing active/failed runs at the top may be confused.

---

## 5. Summary Checklist for PR Approval

| Area | Status | Description |
| :--- | :---: | :--- |
| **Elide Architecture** | ✅ Passed | Complies with Rule 1 (no custom GraphQL controllers; uses native Elide endpoint). |
| **Automated Tests** | ✅ Passed | Integration tests pass in API (`WorkspaceGraphQlPaginationTest`) and UI (`yarn test`). |
| **Lint & Formatting** | ✅ Passed | Strict ESLint and Prettier rules pass without warnings. |
| **RBAC `Access` Scoping** | ❌ Action Required | Workspace-level `role="read"` and `role="plan"` users are blocked from viewing workspaces. |
| **Internal Tokens** | ❌ Action Required | `TerrakubeInternal` accounts lacking group claims get denied by `FalsePredicate`. |
| **Query Performance** | ⚠️ Optimization Needed | 14 GraphQL subqueries per page load; re-executes on page turns; missing DB indexes. |
| **Project Grouping** | ⚠️ UX Consideration | Client-side grouping over paginated records fragments project buckets across pages. |
| **Error Handling** | ⚠️ Fix Needed | Post-load errors are silently swallowed in `OrganizationDetailsPage.tsx`. |

---
**Verdict:** **Request Changes** pending resolution of the workspace-level RBAC check and service account token handling.

Macbet:
@alfespa17 thanks for the second pass, pushed 2febab19.

- **Defect 1:** workspace-level `Access` now grants read for any role (read/plan included), mirroring `ProjectAccess`. Note the `main` operation check (`TeamLimitedViewWorkspace`) had the same manage requirement, so this also fixes the pre-existing behaviour. Covered by `workspaceLevelReadRoleCanListTheWorkspace`.
- **Defect 2:** not reproducible. `TerrakubeInternal` tokens pass `user is a superuser` (`isSuperUser` → `isServiceMember` returns true for the internal issuer), so the read filter never applies to them. Added `internalServiceTokenWithoutGroupsListsWorkspaces` with a system token carrying no `groups` claim to pin that.
- **Defect 3:** `isMember` / `isServiceMember` now delegate to `getEffectiveGroups`, so every claim shape and federated names are handled in one place.
- **Issue 4:** facet counts are only refetched when filters change; page turns and page-size changes reuse the last counts, polling stays count-free.
- **Issue 5:** added `changelog-2.35.0-workspace-pagination-indexes.xml` with `(organization_id, deleted, X)` indexes for status, last job date, project and policy status.
- **Issue 6:** server-side `project.name` sort is not possible here. Elide rejects it with `Combination of sorting over relationship and filtering over toMany relationships unsupported`, and the security filter always joins to-many relations (`organization.team`, `access`, `project.projectAccess`). Grouping stays per page, the header already says "on this page".
- **Issue 7:** failures after the initial load now surface as a toast instead of being swallowed; polling failures stay quiet.
- **Issue 8:** left as is, the "Job status (grouped)" label documents the order.

Also made `loadsTheFullUiRequest` independent of suite order: `AccessTests` leaves an extra workspace in the `simple` org, which broke the exact-count assertion when both suites ran together.

Verification: `AccessTests` (21), `WorkspaceGraphQlPaginationTest` (8), `DexGroupServiceTests` (14), `WorkspaceReadFilterTest` (1) pass on JDK 25; UI jest 173 passed, tsc/eslint/prettier clean.

alfespa17:
thank you @Macbet , this is looking good to merge now.

Just one small thing

api/src/main/resources/db/changelog/local/changelog-2.35.0-workspace-pagination-indexes.xml

The migration should be 2.34.0 not 2.35.0, once you change that I will merge it

Macbet:
@alfespa17 fixed 🏁

alfespa17 (approved):

jakegroves:
this is currently cut into https://github.com/terrakube-io/terrakube/releases/tag/2.34.0-beta.3 

postgres rejects `varchar = smallint`
h2 works ... not sure what database this was tested on?

i pushing through a fix: https://github.com/terrakube-io/terrakube/pull/3586 and cutting a `2.34.0-beta.4`

PR #3518, opened 2026-09-04
feat(api): trigger dependent workspaces when a producer applies

klinux:
**Draft on purpose** — opening this to ask whether the direction makes sense for the platform before investing further. The code works and is tested, but the design decisions below are the actual subject of the discussion, and I'd rather change them now than after building on top.

@alfespa17 — worth your read when you have a moment.

## The problem

Once an organization has more than a handful of workspaces, they stop being independent. A VPN workspace reads the VPC's outputs through `terraform_remote_state`; a cluster workspace needs the network; an application needs the database.

Terraform knows how to *read* across that boundary. What's missing is the trigger: today the consumer only re-plans when a human remembers to run it. So the graph drifts silently, and the drift surfaces weeks later as a surprise in an unrelated plan — usually to whoever is on call, not to whoever caused it.

This is, in my experience, the single most cited reason teams reach for Terragrunt on top of an orchestrator. Terrakube already replaces Terragrunt's original purpose (it manages the backend, so nobody writes `backend` blocks) and the module registry covers the DRY story. Dependencies are the remaining gap.

## What this does

Adds a directed edge between workspaces. When a producer's job reaches `completed`, every workspace declaring a dependency on it gets a run.

```
workspace vpc  ──applies──►  triggers run on  ──►  workspace vpn-ipsec
```

## What this deliberately does *not* do

**It does not block or order runs.** Refusing to apply a consumer while its producer is running or stale is a much bigger change: it needs cycle detection at admission time, a timeout story, a decision about what a failed producer does to its consumers, and careful interaction with the reconciliation sweep that landed in 2.33.

I left all of that out so this is reviewable on its own. If the direction is right, ordering is the natural follow-up — but it deserves its own design discussion, not a rider on this one.

## Design notes

**Reuses existing machinery.** `WorkspaceDependency` is a plain edge table with both FKs cascading on delete. The cascade hooks into `ScheduleJob` at the `completed` case — the same place `jobNotificationTrigger.notifyStatusChanged` already fires from. The scheduler's state machine is untouched.

**Cycles are bounded, not rejected.** A cyclic graph would cascade forever, so each triggered job records how many dependency hops produced it, and the chain stops at `io.terrakube.dependency.maxCascadeDepth` (default 5). I chose a depth limit over validating acyclicity at write time because it's simpler and degrades safely — but rejecting cycles up front is arguably more correct, and I'd rather hear your preference than guess. It's a small change either way.

**Off is one property away.** `io.terrakube.dependency.enabled=false` disables it entirely.

**Template resolution** falls back to the consumer's `defaultTemplate`, with an optional per-edge override, so the common case needs no configuration.

## Open questions

1. **Cycle handling** — depth limit (as implemented) or reject at write time?
2. **Which statuses cascade?** Currently only `completed`. Should `noChanges` also trigger dependents? It means the producer's state is unchanged, so arguably not — but a consumer that never ran would stay stale.
3. **Should the triggered run be a plan or an apply?** It follows the consumer's default template today, so an org can decide per workspace. Making dependent runs always plan-only would be safer but less useful.
4. **Provider support** — configuring this as code needs the resource exposed in `terrakube-io/terraform-provider-terrakube`. Happy to follow up there if this direction is accepted.

## Testing

`mvn -pl api test` on the touched classes: **50 passing** — 7 new plus the 43 in `ScheduleJobTest`, which needed the new constructor argument.

The new tests cover the trigger itself and the cases that would bite in production: the depth limit stopping a cycle, a consumer reachable through two edges getting a single run, deleted workspaces being skipped, and a consumer with no template failing quietly rather than throwing.

Not yet verified against a running instance — I wanted the design conversation first.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/ScheduleJob.java
@@ -11,6 +11,7 @@ import io.terrakube.api.plugin.scheduler.reconciliation.JobReconciliationService
 import io.terrakube.api.plugin.scheduler.reconciliation.ReconciliationProperties;
 import io.terrakube.api.plugin.scheduler.reconciliation.ReconciliationResult;
 import io.terrakube.api.plugin.notification.JobNotificationTrigger;
+import io.terrakube.api.plugin.scheduler.dependency.WorkspaceDependencyService;
 import io.terrakube.api.plugin.softdelete.SoftDeleteService;
 import io.terrakube.api.plugin.variable.IncompleteVariableException;
 import io.terrakube.api.plugin.variable.InvalidVariableCategoryException;
@@ -116,6 +117,10 @@ public class ScheduleJob implements org.quartz.Job {
     // this so status-change notifications actually fire for real runs, not just for a job
     // updated via a direct API PATCH.
     JobNotificationTrigger jobNotificationTrigger;
+
+    // Triggers runs on workspaces that consume this one's output. Only fires on a
+    // successful terminal status, so a failed producer never cascades.
+    WorkspaceDependencyService workspaceDependencyService;
 
     // Shared routine that reconciles a job with no remaining executable step to its real terminal
     // status (see reconcileOutOfSteps). ReconciliationProperties gates the guarded FIFO-admission
@@ -272,6 +277,7 @@ public class ScheduleJob implements org.quartz.Job {
                     updateJobStepsWithStatus(job.getId(), JobStatus.notExecuted);
                     updateJobStatusOnVcs(job, JobStatus.completed);
                     postPrCommentIfNeeded(job);
+                    workspaceDependencyService.triggerDependents(job);
                     deleteOldJobs(job);
                     break;
                 case cancelled:
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/dependency/WorkspaceDependencyService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/dependency/WorkspaceDependencyService.java
@@ -1,0 +1,143 @@
+package io.terrakube.api.plugin.scheduler.dependency;
+
+import io.terrakube.api.plugin.scheduler.ScheduleJobService;
+import io.terrakube.api.repository.JobRepository;
+import io.terrakube.api.repository.WorkspaceDependencyRepository;
+import io.terrakube.api.rs.job.Job;
+import io.terrakube.api.rs.job.JobStatus;
+import io.terrakube.api.rs.workspace.Workspace;
+import io.terrakube.api.rs.workspace.dependency.WorkspaceDependency;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+import java.sql.Timestamp;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * Triggers runs on workspaces that consume output from a workspace that just applied.
+ *
+ * The typical case is a workspace reading another one's state through
+ * terraform_remote_state - a VPN that needs the VPC's outputs, for example. Today that
+ * consumer only re-plans when someone remembers to run it, so the graph drifts silently
+ * until a plan surprises whoever is on call.
+ *
+ * Scope is deliberately narrow: this only *triggers* dependent runs, it does not block or
+ * order them. Ordering (refusing to apply a consumer while its producer is running or
+ * stale) is a larger change and is intentionally left out of this iteration - see the pull
+ * request description.
+ */
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class WorkspaceDependencyService {
+
+    private static final String TRIGGERED_VIA = "Dependency";
+
+    private final WorkspaceDependencyRepository workspaceDependencyRepository;
+    private final JobRepository jobRepository;
+    private final ScheduleJobService scheduleJobService;
+
+    /**
+     * Guards against a dependency cycle turning into an endless chain of runs. A run
+     * triggered by a dependency does not itself trigger further runs beyond this depth.
+     */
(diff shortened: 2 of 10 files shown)


alfespa17:
We could check how other platform solved things like this 🤔, how does other tacos manage this things? Any idea or doc could be a good place to start

alfespa17:
Some references that I found

- https://developer.hashicorp.com/terraform/cloud-docs/workspaces/settings/run-triggers
- https://docs.spacelift.io/concepts/stack/stack-dependencies

klinux:
Good call — I went looking before seeing your links, and landed on the same two. Adding Scalr to round out the picture, plus what the comparison actually settles.

## How the other TACOS do it

| Platform | Trigger | Model |
|---|---|---|
| [HCP Terraform / TFE](https://developer.hashicorp.com/terraform/cloud-docs/workspaces/settings/run-triggers) | successful apply on source | plain edge, **max 20 sources** per workspace; the run detail links back to the source workspace and the run that fired it |
| [Scalr](https://docs.scalr.io/docs/workspace-dependencies) | successful run upstream | same shape; if the downstream has auto-apply it applies, otherwise it waits for approval |
| [Spacelift](https://docs.spacelift.io/concepts/stack/stack-dependencies) | successful apply on producer | goes further: passes **outputs directly** into the consumer's inputs, no `terraform_remote_state` |

All three fire on **successful apply of the producer**, which is what this PR implements. Worth noting Terrakube already speaks the TFC API — the `cloud {}` block works — so Run Triggers is arguably the closest analogue rather than a new concept.

## Two levels, and this PR is level 1

The comparison makes the design space clear:

- **Level 1 (TFC, Scalr)** — trigger only. The consumer still reads via `terraform_remote_state`. This is what's implemented here.
- **Level 2 (Spacelift)** — trigger plus output passing. Considerably more powerful and considerably more involved: sensitive outputs crossing a workspace boundary need explicit opt-in, and Spacelift gates it per stack *and* per worker pool.

I'd suggest level 1 first, and only consider level 2 if the demand shows up — but that's your call on where the platform should sit.

## It answers three of my open questions

**Plan or apply for the triggered run?** Scalr settles it nicely: the decision belongs to the *consumer*, not the edge — it follows the downstream's own auto-apply setting. This PR uses the consumer's `defaultTemplate`, which is the same principle.

**Should `noChanges` cascade?** Spacelift has an explicit `trigger_always` flag for "fire even when the referenced value didn't change". So it's configuration, not a global decision. I'd add it per edge rather than pick a default.

**A cap on fan-in.** TFC stops at 20 source workspaces. I hadn't capped anything; cheap protection against a pathological graph, and I'll add it.

There's also one thing I was missing: TFC records provenance on the triggered run — which workspace and which run caused it. I only store the cascade depth. Recording the originating job is small and makes "why did this run?" answerable, so I'll add that too.

## What the references don't settle

**Cycle handling.** None of the three documents what happens with `a -> b -> a`. This PR bounds it with a depth limit rather than rejecting cycles at write time — simple and degrades safely, but validating acyclicity on write is arguably more correct. Still happy to go either way.

## Note on CI

`build-jammy` is failing on `ProjectAccessTests` and `AccessTests`, not on the new tests — I'm looking into it now. Current suspicion is that `WorkspaceDependency` having two `@ManyToOne` relations to `Workspace` confuses Elide's parent resolution under `@Include(rootLevel = false)`. Will push a fix; flagging so nobody spends time on it.

klinux:
Two things: CI is fixed, and a design argument for where this should head.

## CI

Green now — the failures were mine, not flakiness.

`PhysicalNamingStrategyStandardImpl` means Hibernate takes the field name literally, so my migration created `template_reference` while the entity asked for `TEMPLATEREFERENCE`, and every query touching the relation returned 500. `Job.java` already annotates the same field explicitly; this now matches.

The class-level permissions were wrong too: `TeamViewWorkspace` and `TeamManageWorkspace` are `OperationCheck<Workspace>`, so applying them to a `WorkspaceDependency` can't work. Every other child of `Workspace` — `Variable`, `WorkspaceTag`, `Schedule` — declares none and inherits from the parent when reached through the relation. Now it follows that.

`ProjectAccessTests` (12), `AccessTests` (21), `ScheduleJobTest` (43) and the new tests (7) all pass — 83 total.

## Where this should head: the two levels aren't alternatives

Re-reading the Spacelift docs against the TFC ones, there's a point worth making explicit: **Spacelift also triggers on the producer's apply**. Output passing is *additive* to the trigger, not a different approach.

```
level 1:  edge + trigger                          ← this PR
level 2:  edge + trigger + output→input mapping    ← Spacelift
```

So level 1 isn't throwaway work if level 2 is the destination — it's the foundation. What changes is the **modelling**: if output references are the goal, the edge should anticipate a child table for them (`output_name` → `input_name`, like `spacelift_stack_dependency_reference`), rather than being reshaped later.

## The argument for level 2 that I think matters most

Not convenience — **least privilege**.

Today the consumer reads via `terraform_remote_state`, which requires access to the producer's **entire state**: every resource, every attribute, including anything sensitive in there. For a VPN workspace to read one CIDR from a VPC workspace, it needs to be able to read everything the VPC workspace manages.

Output passing narrows that to the named output. That reframes the feature: it isn't only orchestration convenience, it also shrinks the access surface between workspaces. It's probably why Spacelift gates sensitive outputs behind an explicit opt-in at both the stack and worker pool level — they clearly treat the boundary as a security one.

Worth noting Terrakube already has some of the raw material: `TerraformOutputController`, the `StateOutputs`/`OutputData` models, and outputs already persisted per job in storage. Capturing wouldn't start from zero; what's missing is the transport into the consumer's run.

## What I'd suggest

Keep this PR at level 1, but decide the modelling knowing level 2 is where it may go. Concretely: if you'd want output references eventually, I'd shape the edge now so a references table hangs off it cleanly — that costs nothing today and avoids a migration that rewrites the relationship later.

Happy to go either way, and equally happy to park this if the platform doesn't want the surface area. Just let me know which of the two you'd rather have, and whether cycles should be bounded by depth (as implemented) or rejected at write time.

alfespa17:
I am closing this, maybe we can discuss first a little bit more before doing any kind of implementation

alfespa17:
@klinux please check https://github.com/terrakube-io/terrakube/issues/3520

PR #3529, opened 2026-09-07
fix(storage): configure S3 checksum behavior on the client only

### LINKED ISSUE #3528: 2.34.0-beta.1: api/executor/registry crash-loop with self-hosted S3 endpoint
### Bug description 🐞

I want to be completely transparent, and I fully understand if this gets closed without consideration, the below was found whilst working on my homelab setup with AI assistance. I asked for it to triple-check the source(s) and my own configuration to rule out misconfig as the root of the problem, but it is a sycophantic statistical model at the end of the day. I couldn't see anything in the contribution guidelines etc. about AI-generated input, which is why I'm raising this.

Upgrading from `2.33.1` to `2.34.0-beta.1` breaks `terrakube-api`, `terrakube-executor`, and `terrakube-registry` at startup whenever a custom S3-compatible endpoint (MinIO, in our case — the same pattern as upstream's own example `docker-compose.yml`) is configured via `io.terrakube.storage.aws.endpoint`. All three crash-loop with:

```
java.lang.IllegalStateException: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
```

**Root cause:** [PR #3473](https://github.com/terrakube-io/terrakube/pull/3473) ("optional S3 endpoint region, chunked encoding and checksum validation parameters"), merged 2026-08-31 and first shipped in `2.34.0-beta.1`, added `checksumValidationEnabled` to `AwsStorageTypeProperties`. In the custom-endpoint branch of each component's storage auto-configuration (`StorageTypeAutoConfiguration` in `api`, the equivalent in `executor`'s `TerraformStateAutoConfiguration`, and `registry`'s own copy), it now calls both of these on the same S3 client build:

```java
S3Configuration serviceConfiguration = s3ServiceConfiguration(awsStorageTypeProperties); // -> .checksumValidationEnabled(props.isChecksumValidationEnabled())
s3client = S3Client.builder()
        ...
        .serviceConfiguration(serviceConfiguration)
        .requestChecksumCalculation(RequestChecksumCalculation.WHEN_REQUIRED)
        ...
```

The AWS SDK v2's own Javadoc on `S3Configuration.Builder#checksumValidationEnabled` documents that this exact combination throws:

> *This option has been replaced with `S3ClientBuilder#requestChecksumCalculation(...)`... **If both this and one of those options are set, an exception will be thrown.***

`checksumValidationEnabled` is a primitive `boolean` on `AwsStorageTypeProperties` (not `Boolean`), so `props.isChecksumValidationEnabled()` always returns `true` or `false`, never `null` — there is no property value that skips the `.checksumValidationEnabled(...)` builder call. This makes the failure **unconditional** for anyone using the custom-endpoint branch, regardless of what `io.terrakube.storage.aws.checksum-validation-enabled` (or its Coolify/env equivalent) is set to. It does not affect the plain-AWS-S3 / role-authentication branches, since those never call `s3ServiceConfiguration`.

**Impact:** every self-hosted deployment using MinIO/Ceph/R2/Wasabi/etc. as the S3 backend — which is upstream's own documented `docker-compose.yml` pattern — hits this immediately on startup, on all three affected components independently. We confirmed `2.33.1` (the commit immediately before PR #3473) does not have this code path at all and starts cleanly with an identical MinIO configuration.

### Steps to reproduce

1. Deploy Terrakube `2.34.0-beta.1` (any of the three affected components) via Docker Compose with a self-hosted S3-compatible backend — e.g. the bundled MinIO service from upstream's own [example `docker-compose.yml`](https://docs.terrakube.io/getting-started/docker-compose) — so that `io.terrakube.storage.aws.endpoint` (`AwsEndpoint` in the compose env) is set to a non-empty custom endpoint.
2. Start `terrakube-api`, `terrakube-executor`, and `terrakube-registry`.
3. Observe each container fail application-context startup and restart-loop with the `IllegalStateException` above.

### Expected behavior

Should start cleanly and behave as `2.33.1` did with an identical configuration — no `IllegalStateException`, no crash-loop. Either drop the new `checksumValidationEnabled`-driven `S3Configuration` builder call in the custom-endpoint branch, or make it mutually exclusive with the hardcoded `requestChecksumCalculation(WHEN_REQUIRED)` call (e.g. only set the client-level option when the deprecated property was not explicitly configured, per the AWS SDK's own migration guidance for this field).

### Example repository

No separate reproduction repository; this reproduces with upstream's own [documented example `docker-compose.yml`](https://docs.terrakube.io/getting-started/docker-compose) (which uses MinIO as the S3 backend) simply by bumping `TK_VERSION`/image tags to `2.34.0-beta.1`. Affected source, as of `2.34.0-beta.1`: `api/src/main/java/io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.java` (`terraformOutput` bean, `AWS` case), plus the structurally equivalent auto-configuration classes in `executor` and `registry`; the new property lives in `api/src/main/java/io/terrakube/api/plugin/storage/aws/AwsStorageTypeProperties.java`.

### Anything else?

Environment: Terrakube `2.34.0-beta.1`, deployed via Docker Compose, MinIO backend, custom `AwsEndpoint`. Reproducible independently on all three components.

Logs — `terrakube-registry`:

```
2026-09-06T18:49:30.486Z  INFO 1 --- [           main] i.t.r.p.s.c.StorageAutoConfiguration     : Creating AWS SDK with custom endpoint and custom credentials
2026-09-06T18:49:30.537Z  WARN 1 --- [           main] ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'moduleWebServiceImpl': Unsatisfied dependency expressed through field 'moduleService': Error creating bean with name 'moduleServiceImpl' defined in file [/workspace/BOOT-INF/classes/io/terrakube/registry/service/module/ModuleServiceImpl.class]: Unsatisfied dependency expressed through constructor parameter 1: Error creating bean with name 'terraformOutput' defined in class path resource [io/terrakube/registry/plugin/storage/configuration/StorageAutoConfiguration.class]: Failed to instantiate [io.terrakube.registry.plugin.storage.StorageService]: Factory method 'terraformOutput' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.

Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'moduleServiceImpl' defined in file [/workspace/BOOT-INF/classes/io/terrakube/registry/service/module/ModuleServiceImpl.class]: Unsatisfied dependency expressed through constructor parameter 1: Error creating bean with name 'terraformOutput' defined in class path resource [io/terrakube/registry/plugin/storage/configuration/StorageAutoConfiguration.class]: Failed to instantiate [io.terrakube.registry.plugin.storage.StorageService]: Factory method 'terraformOutput' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'terraformOutput' defined in class path resource [io/terrakube/registry/plugin/storage/configuration/StorageAutoConfiguration.class]: Failed to instantiate [io.terrakube.registry.plugin.storage.StorageService]: Factory method 'terraformOutput' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [io.terrakube.registry.plugin.storage.StorageService]: Factory method 'terraformOutput' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
Caused by: java.lang.IllegalStateException: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
```

Logs — `terrakube-api`:

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'moduleWebServiceImpl': ... Error creating bean with name 'moduleServiceImpl' ... Error creating bean with name 'terraformOutput' defined in class path resource [io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.class]: Failed to instantiate [io.terrakube.api.plugin.storage.StorageTypeService]: Factory method 'terraformOutput' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'terraformOutput' defined in class path resource [io/terrakube/api/plugin/storage/configuration/StorageTypeAutoConfiguration.class]: Failed to instantiate [io.terrakube.api.plugin.storage.StorageTypeService]: Factory method 'terraformOutput' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
Caused by: java.lang.IllegalStateException: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
```

Logs — `terrakube-executor`:

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'terraformExecutorServiceImpl' ...
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'setupWorkspaceImpl' defined in file [/workspace/BOOT-INF/classes/io/terrakube/executor/service/workspace/SetupWorkspaceImpl.class] ...
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'terraformState' defined in class path resource [io/terrakube/executor/plugin/tfstate/configuration/TerraformStateAutoConfiguration.class]: Failed to instantiate [io.terrakube.executor.plugin.tfstate.TerraformState]: Factory method 'terraformState' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [io.terrakube.executor.plugin.tfstate.TerraformState]: Factory method 'terraformState' threw exception with message: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
Caused by: java.lang.IllegalStateException: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
```

**Workaround:** pin to `2.33.1` until fixed.

7487:
Fixes #3528

Since #3473 the custom-endpoint branch of the S3 client setup in `api`, `executor` and `registry` sets `checksumValidationEnabled(...)` on `S3Configuration` while the same builder chain also sets `requestChecksumCalculation(WHEN_REQUIRED)` on the client. AWS SDK v2 (`DefaultS3BaseClientBuilder.finalizeServiceConfiguration`) rejects that combination when the client is built:

```
java.lang.IllegalStateException: Checksum behavior has been configured on both S3Configuration and the client/global level. Please limit checksum behavior configuration to one location.
```

The property is a primitive `boolean`, so the `S3Configuration` setter is always called and the failure is unconditional for anyone using `io.terrakube.storage.aws.endpoint` (MinIO, Ceph, R2, ...), whatever `checksumValidationEnabled` is set to. Reproduced on SDK 2.54.7 with both `true` and `false`.

### Change

- Drop `checksumValidationEnabled` from `S3Configuration` in all three components.
- Map the property to the client-level `responseChecksumValidation` instead (`true` -> `WHEN_SUPPORTED`, the SDK default; `false` -> `WHEN_REQUIRED`). This is the same mapping the SDK applies internally for the legacy flag. Request checksums stay `WHEN_REQUIRED`, as they were before #3473, so the default configuration behaves exactly like 2.33.1.
- `api`: extract the custom-endpoint client build into `StorageTypeAutoConfiguration.s3CompatibleClient(...)` (same static-helper pattern as `storageClientOverride` / `s3ServiceConfiguration`) so the test can build the real client.

### Tests

`StorageClientResilienceTest`:
- `s3ServiceConfigurationLeavesChecksumBehaviorToClient`: `S3Configuration.checksumValidationEnabled` stays unset.
- `s3CompatibleClientBuildsWithChecksumValidationEnabled` / `...Disabled`: build the custom-endpoint `S3Client` and assert the resolved request/response checksum settings.

All three new tests fail against `main` (2 x `IllegalStateException`, 1 assertion) and pass with this change. `executor` and `registry` compile; their change is the same two lines inline.

### DIFF
--- a/api/src/test/java/io/terrakube/api/plugin/storage/configuration/StorageClientResilienceTest.java
+++ b/api/src/test/java/io/terrakube/api/plugin/storage/configuration/StorageClientResilienceTest.java
@@ -120,6 +120,7 @@ class StorageClientResilienceTest {
         props.setEndpoint("http://localhost:9000");
         props.setAccessKey("minio");
         props.setSecretKey("minio123");
+        props.setChecksumValidationEnabled(true);
         return props;
     }
 }
--- a/api/src/test/java/io/terrakube/api/plugin/storage/configuration/StorageClientResilienceTest.java
+++ b/api/src/test/java/io/terrakube/api/plugin/storage/configuration/StorageClientResilienceTest.java
@@ -2,14 +2,18 @@ package io.terrakube.api.plugin.storage.configuration;
 
 import io.terrakube.api.plugin.storage.aws.AwsStorageTypeProperties;
 import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.core.checksums.RequestChecksumCalculation;
+import software.amazon.awssdk.core.checksums.ResponseChecksumValidation;
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
 import software.amazon.awssdk.core.retry.RetryPolicy;
+import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.S3Configuration;
 
 import java.time.Duration;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class StorageClientResilienceTest {
@@ -47,7 +51,6 @@ class StorageClientResilienceTest {
         assertEquals("auto", props.getEndpointRegion());
         assertTrue(config.pathStyleAccessEnabled());
         assertTrue(config.chunkedEncodingEnabled());
-        assertTrue(config.checksumValidationEnabled());
     }
 
     @Test
@@ -55,13 +58,11 @@ class StorageClientResilienceTest {
         AwsStorageTypeProperties props = new AwsStorageTypeProperties();
         props.setEndpointRegion("us-east-1");
         props.setChunkedEncodingEnabled(false);
-        props.setChecksumValidationEnabled(false);
 
         S3Configuration config = StorageTypeAutoConfiguration.s3ServiceConfiguration(props);
 
         assertEquals("us-east-1", props.getEndpointRegion());
         assertFalse(config.chunkedEncodingEnabled());
-        assertFalse(config.checksumValidationEnabled());
         assertTrue(config.pathStyleAccessEnabled());
     }
 
@@ -73,5 +74,52 @@ class StorageClientResilienceTest {
         S3Configuration config = StorageTypeAutoConfiguration.s3ServiceConfiguration(props);
 
         assertFalse(config.pathStyleAccessEnabled());
+    }
+
+    @Test
+    void s3ServiceConfigurationLeavesChecksumBehaviorToClient() {
+        AwsStorageTypeProperties props = new AwsStorageTypeProperties();
+        props.setChecksumValidationEnabled(false);
+
+        S3Configuration config = StorageTypeAutoConfiguration.s3ServiceConfiguration(props);
+
+        // The SDK refuses to build a client when checksum behavior is set on both S3Configuration
+        // and the client (#3528), so it must stay unset here.
+        assertNull(config.toBuilder().checksumValidationEnabled());
+    }
+
+    @Test
+    void s3CompatibleClientBuildsWithChecksumValidationEnabled() {
+        AwsStorageTypeProperties props = s3CompatibleProps();
+
+        try (S3Client client = StorageTypeAutoConfiguration.s3CompatibleClient(props,
+                StorageTypeAutoConfiguration.storageClientOverride(props))) {
+            assertEquals(RequestChecksumCalculation.WHEN_REQUIRED,
+                    client.serviceClientConfiguration().requestChecksumCalculation());
+            assertEquals(ResponseChecksumValidation.WHEN_SUPPORTED,
+                    client.serviceClientConfiguration().responseChecksumValidation());
+        }
+    }
+
+    @Test
+    void s3CompatibleClientBuildsWithChecksumValidationDisabled() {
+        AwsStorageTypeProperties props = s3CompatibleProps();
+        props.setChecksumValidationEnabled(false);
+
+        try (S3Client client = StorageTypeAutoConfiguration.s3CompatibleClient(props,
+                StorageTypeAutoConfiguration.storageClientOverride(props))) {
+            assertEquals(RequestChecksumCalculation.WHEN_REQUIRED,
+                    client.serviceClientConfiguration().requestChecksumCalculation());
+            assertEquals(ResponseChecksumValidation.WHEN_REQUIRED,
+                    client.serviceClientConfiguration().responseChecksumValidation());
+        }
+    }
+
+    private static AwsStorageTypeProperties s3CompatibleProps() {
+        AwsStorageTypeProperties props = new AwsStorageTypeProperties();
+        props.setEndpoint("http://localhost:9000");
+        props.setAccessKey("minio");
+        props.setSecretKey("minio123");
+        return props;
     }
 }
--- a/executor/src/main/java/io/terrakube/executor/plugin/tfstate/configuration/TerraformStateAutoConfiguration.java
+++ b/executor/src/main/java/io/terrakube/executor/plugin/tfstate/configuration/TerraformStateAutoConfiguration.java
@@ -99,9 +99,7 @@ public class TerraformStateAutoConfiguration {
                                 .endpointOverride(URI.create(awsTerraformStateProperties.getEndpoint()))
                                 .serviceConfiguration(serviceConfiguration)
                                 .requestChecksumCalculation(RequestChecksumCalculation.WHEN_REQUIRED)
-                                .responseChecksumValidation(awsTerraformStateProperties.isChecksumValidationEnabled()
-                                        ? ResponseChecksumValidation.WHEN_SUPPORTED
-                                        : ResponseChecksumValidation.WHEN_REQUIRED)
+                                .responseChecksumValidation(responseChecksumValidation(awsTerraformStateProperties.isChecksumValidationEnabled()))
                                 .build();
                     } else {
                         log.info("Creating AWS SDK with custom credentials");
@@ -165,6 +163,12 @@ public class TerraformStateAutoConfiguration {
                     .terraformStatePathService(terraformStatePathService)
                     .build();
         return terraformState;
+    }
+
+    static ResponseChecksumValidation responseChecksumValidation(boolean checksumValidationEnabled) {
+        return checksumValidationEnabled
+                ? ResponseChecksumValidation.WHEN_SUPPORTED
+                : ResponseChecksumValidation.WHEN_REQUIRED;
     }
 
     private AwsBasicCredentials getAwsBasicCredentials(AwsTerraformStateProperties awsTerraformStateProperties) {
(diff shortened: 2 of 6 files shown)


api/src/test/java/io/terrakube/api/plugin/storage/configuration/StorageClientResilienceTest.java
```diff
@@ -74,4 +75,51 @@ void s3ServiceConfigurationHonorsDisabledPathStyleAccess() {
...
+
+    @Test
+    void s3CompatibleClientBuildsWithChecksumValidationDisabled() {
+        AwsStorageTypeProperties props = s3CompatibleProps();
+        props.setChecksumValidationEnabled(false);
+
+        try (S3Client client = StorageTypeAutoConfiguration.s3CompatibleClient(props,
+                StorageTypeAutoConfiguration.storageClientOverride(props))) {
+            assertEquals(RequestChecksumCalculation.WHEN_REQUIRED,
+                    client.serviceClientConfiguration().requestChecksumCalculation());
```
utafrali, line 112: `s3CompatibleProps()` never calls `props.setChecksumValidationEnabled(true)`, so `s3CompatibleClientBuildsWithChecksumValidationEnabled` silently depends on the property's default value being `true`. If that default ever changes, the test will still compile and run but will be testing the wrong branch of the ternary. Adding `props.setChecksumValidationEnabled(true)` here makes the test's intent self-documenting and removes the hidden coupling to the default.

executor/src/main/java/io/terrakube/executor/plugin/tfstate/configuration/TerraformStateAutoConfiguration.java
```diff
@@ -97,6 +99,9 @@ public TerraformState terraformState(TerrakubeClient terrakubeClient, TerraformS
                                 .endpointOverride(URI.create(awsTerraformStateProperties.getEndpoint()))
                                 .serviceConfiguration(serviceConfiguration)
```
utafrali, line 100: The boolean-to-`ResponseChecksumValidation` mapping (`true` -> `WHEN_SUPPORTED`, `false` -> `WHEN_REQUIRED`) is now duplicated inline in `executor` and `registry` rather than shared with `api`'s `s3CompatibleClient`. The `api` module extracted this into a testable static helper precisely so the mapping could be verified in isolation. If the SDK adds a third checksum mode or the mapping logic needs to change, all three copies need updating. Consider extracting the mapping to a shared static method (even package-private) in each module, mirroring what `api` did.

utafrali (approved):
Clean, well-targeted fix for a real unconditional crash. The root cause is correctly diagnosed, the SDK mapping is accurate, and the three new tests are designed to fail against `main` and pass with the fix. The two minor issues above are worth addressing but do not block merging.

api/src/test/java/io/terrakube/api/plugin/storage/configuration/StorageClientResilienceTest.java
7487: Good catch. `s3CompatibleProps()` now calls `props.setChecksumValidationEnabled(true)` explicitly in 4c1f64a, so the enabled-branch test no longer relies on the property default.

executor/src/main/java/io/terrakube/executor/plugin/tfstate/configuration/TerraformStateAutoConfiguration.java
7487: Done in 4c1f64a: `executor` and `registry` each get a package-private `static ResponseChecksumValidation responseChecksumValidation(boolean)` and the builders call it instead of the inline ternary. Added a one-test class per module (`TerraformStateAutoConfigurationTest`, `StorageAutoConfigurationTest`) asserting `true -> WHEN_SUPPORTED`, `false -> WHEN_REQUIRED`. Ran the touched tests in all three modules with JDK 25: 10/10 green.

alfespa17 (approved):

PR #3543, opened 2026-09-09
fix: read job log stream with a single offset in RemoteTfeService

### LINKED ISSUE #3540: RemoteTfeService passes a duplicate stream key to XREAD, breaking remote-backend logs on strict Redis-protocol servers
### Summary

`RemoteTfeService.getPlanLogs()` and `getApplyLogs()` each pass the same Redis stream key twice to a single `XREAD`. Redis and Valkey tolerate this, so most deployments never notice. Servers that validate the `STREAMS` clause reject it outright, and because the resulting exception is swallowed at `log.debug`, the endpoint the Terraform CLI polls as a run's `log-read-url` then returns an empty body with nothing logged anywhere.

### The code

https://github.com/terrakube-io/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java#L1613-L1615

```java
List<MapRecord<String, String, String>> messagesPlan = redisTemplate.opsForStream()
        .read(StreamOffset.fromStart(String.valueOf(job.get().getId())),
                StreamOffset.latest(String.valueOf(job.get().getId())));
```

`getApplyLogs()` has the same call at L1653-L1655.

`read(StreamOffset...)` treats each vararg as its own stream, so this sends `XREAD STREAMS <jobId> <jobId> 0-0 $`.

### Why the second offset is wrong regardless of backend

`StreamOffset.latest()` is `$`, meaning "entries added after this call". In a non-blocking `XREAD` that is always empty. The `fromStart` offset already returns the entire history, which is what the surrounding code wants -- it accumulates everything into a `TextStringBuilder` and then slices by the `offset`/`limit` query parameters. So the second offset contributes nothing on any backend; it only duplicates the key.

### Behaviour by backend

Verified with Testcontainers against real servers, using the project's own Spring Data Redis / Lettuce stack (no mocks):

| Backend | `read(fromStart(k), latest(k))` | `read(fromStart(k))` |
|---|---|---|
| Redis 6.2, 7.0, 7.2, 7.4, 8.x | returns full history | returns full history |
| Valkey 7.2, 8, 9 | returns full history | returns full history |
| DragonflyDB | throws `RedisCommandExecutionException: ERR Same stream specified multiple time` | returns full history |

DragonflyDB dedupes the `STREAMS` keys via a map insert and errors on any repeated key, independent of the offsets given ([`stream_family.cc`](https://github.com/dragonflydb/dragonfly/blob/main/src/server/stream_family.cc)). Redis has no such validation, which is why the defect stayed latent.

Since the chart lets operators point at an external server via `api.properties.redisHostname`, a strict implementation is a reachable configuration.

### Why it is invisible when it does break

```java
} catch (Exception ex) {
    log.debug(ex.getMessage());
}
```

DEBUG is not enabled in any shipped configuration, so `logs` keeps its initial `"".getBytes()` value and every poll returns an empty body with no diagnostic. From the operator's side `terraform plan` simply prints its preamble and never streams anything.

Worth noting this catch is not on a routine path: reading a stream that does not exist returns empty rather than erroring on both Redis and DragonflyDB, so raising the level does not make normal polling noisy.

### Proposed fix

Drop the second `StreamOffset` in both methods, and raise the swallowed `log.debug` to `log.error` so a future failure here is diagnosable.

### Notes for reproducing

Terrakube deletes the stream at job completion (`LogsServiceRedis.deleteLogs`), so `XLEN <jobId>` is 0 once a run finishes -- reproduce against a job while it is still running.

### Version

Verified on `main` at e7d51f0b.

putz612:
Fixes #3540

`getPlanLogs()` and `getApplyLogs()` each passed the same Redis stream key twice to a single `XREAD`:

```java
.read(StreamOffset.fromStart(String.valueOf(job.get().getId())),
        StreamOffset.latest(String.valueOf(job.get().getId())));
```

`StreamOffset.latest()` is `$`, which in a non-blocking `XREAD` means "entries added after this call" and so contributes nothing to a historical read. `fromStart` alone already returns the whole stream, which is what the surrounding code expects — it accumulates everything into a `TextStringBuilder` and then slices by the `offset`/`limit` query parameters. The second offset therefore only duplicated the stream key.

Redis and Valkey accept the repeated key, so this was latent for them. Servers that validate the `STREAMS` clause reject it, and the exception was swallowed at `log.debug`, leaving the CLI's `log-read-url` returning an empty body with nothing logged.

## Changes

- Drop the second `StreamOffset` in both methods.
- Raise the swallowed `log.debug(ex.getMessage())` to `log.error(...)` with the job id, in both catch blocks, so a future failure here is diagnosable rather than silent.

## Verification

Measured against real servers with Testcontainers, using this project's own Spring Data Redis / Lettuce stack:

| Backend | before | after |
|---|---|---|
| Redis 6.2 / 7.0 / 7.2 / 7.4 / 8.x | full history returned | full history returned |
| Valkey 7.2 / 8 / 9 | full history returned | full history returned |
| DragonflyDB | `RedisCommandExecutionException: ERR Same stream specified multiple time` | full history returned |

Driving the real `getPlanLogs` against DragonflyDB: empty string before the change, the plan output after it.

On the logging change specifically — reading a stream that does not exist returns empty rather than erroring on both Redis and DragonflyDB, so this catch is not on the routine polling path and raising the level does not make normal operation noisy.

## Tests

`RemoteTfeServiceTest` gains a regression test per method asserting exactly one `StreamOffset` is passed and that its key is the job id. Asserting on the returned text alone would still pass against the old call under a lenient mock, so these pin the call shape instead.

I also wrote two Testcontainers integration tests covering the table above, including one that drives `getPlanLogs` end to end against DragonflyDB. They are not included here because this module's tests are currently mock-based and they would add container pulls to CI — happy to add them if you would like them.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
@@ -1611,8 +1611,7 @@ public class RemoteTfeService {
                         try {
                             @SuppressWarnings("unchecked")
                             List<MapRecord<String, String, String>> messagesPlan = redisTemplate.opsForStream()
-                                    .read(StreamOffset.fromStart(String.valueOf(job.get().getId())),
-                                            StreamOffset.latest(String.valueOf(job.get().getId())));
+                                    .read(StreamOffset.fromStart(String.valueOf(job.get().getId())));
 
                             for (MapRecord<String, String, String> mapRecord : messagesPlan) {
                                 Map<String, String> streamData = (Map<String, String>) mapRecord.getValue();
@@ -1626,7 +1625,7 @@ public class RemoteTfeService {
                             logs = logsOutputString.substring(offset, endIndex).getBytes(StandardCharsets.UTF_8);
                             log.debug("{}", logs);
                         } catch (Exception ex) {
-                            log.debug(ex.getMessage());
+                            log.error("Could not read plan logs for job {}: {}", job.get().getId(), ex.getMessage());
                         }
                     }
                 }
@@ -1651,8 +1650,7 @@ public class RemoteTfeService {
                         try {
                             @SuppressWarnings("unchecked")
                             List<MapRecord<String, String, String>> messagesApply = redisTemplate.opsForStream().read(
-                                    StreamOffset.fromStart(String.valueOf(job.get().getId())),
-                                    StreamOffset.latest(String.valueOf(job.get().getId())));
+                                    StreamOffset.fromStart(String.valueOf(job.get().getId())));
 
                             for (MapRecord<String, String, String> mapRecord : messagesApply) {
                                 Map<String, String> streamData = (Map<String, String>) mapRecord.getValue();
@@ -1666,7 +1664,7 @@ public class RemoteTfeService {
                             logs = logsOutputString.substring(offset, endIndex).getBytes(StandardCharsets.UTF_8);
                             log.debug("{}", logs);
                         } catch (Exception ex) {
-                            log.debug(ex.getMessage());
+                            log.error("Could not read apply logs for job {}: {}", job.get().getId(), ex.getMessage());
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


alfespa17 (approved):

alfespa17:
thank you @putz612

PR #3547, opened 2026-09-11
feat: workspace dependency graph and run triggers

### LINKED ISSUE #3520: feat: High-Level Architecture Design: Workspace Dependency Graph & Run Triggers in Terrakube
## 1. Executive Summary

In complex cloud infrastructure architectures, infrastructure components are modularized across distinct workspaces (e.g., `vpc-networking` &rarr; `database-clusters` &rarr; `application-workloads`). A successful run in an upstream workspace must automatically trigger downstream dependent workspaces to propagate changes (such as state outputs, VPC IDs, or database connection strings) without manual intervention.

This architecture introduces **Upstream Run Triggers** to Terrakube. By configuring run triggers, workspaces declare dependencies on upstream source workspaces. When a job modifying an upstream workspace finishes with `completed` status, the Terrakube Run Trigger engine evaluates the dependency graph, enforces cycle protection, and reactively enqueues downstream jobs. These downstream jobs are then dispatched through Terrakube's existing distributed locking and two-tier FIFO queue.

```mermaid
graph TD
    subgraph Organization
        WS_NET["Workspace: Networking (VPC)"]
        WS_DB["Workspace: Database (RDS)"]
        WS_APP1["Workspace: API Service"]
        WS_APP2["Workspace: Web Frontend"]

        WS_NET -->|"Run Trigger (on apply completed)"| WS_DB
        WS_NET -->|"Run Trigger (on apply completed)"| WS_APP1
        WS_DB -->|"Run Trigger (on apply completed)"| WS_APP1
        WS_APP1 -->|"Run Trigger (on apply completed)"| WS_APP2
    end

    classDef ws fill:#1e293b,stroke:#3b82f6,stroke-width:2px,color:#fff;
    class WS_NET,WS_DB,WS_APP1,WS_APP2 ws;
```

---

## 2. Key Design Decisions

Following our architectural discovery and design tree resolution, the system adheres to the following principles:

| Decision Area | Selected Design Choice | Rationale |
|---|---|---|
| **Model Architecture** | **Upstream Run Triggers** (`WorkspaceRunTrigger`) | Downstream workspaces declare subscriptions to upstream workspaces, creating a decentralized, composable DAG without requiring rigid monolithic pipeline definitions. |
| **Trigger Qualification** | **State-Changing Runs Only** (`apply` / `destroy` `completed`) | Pure speculative plans do not modify state and must not trigger downstream runs. Only runs that change infrastructure state qualify. |
| **Multi-Upstream Synchronization** | **Independent Reactive Triggering** | Each upstream completion immediately triggers the downstream workspace. Terrakube's existing workspace-level queue safely serializes multiple executions on the same workspace. |
| **Cycle & Loop Safeguards** | **Static DFS Cycle Detection + Runtime Hop Limit** | Static graph validation prevents cycles at configuration time (HTTP 400); runtime causation tracking (`cascadeDepth <= 10`) guarantees termination against unexpected cascades. |
| **Boundary Scope** | **Intra-Organization** | Dependencies can span any workspaces within the same Organization (including across different Projects), maintaining tenant isolation while supporting cross-project architecture. |
| **Execution Template** | **Downstream Default Template (with optional trigger override)** | Executes the downstream workspace's `defaultTemplate` by default, but allows overriding the template in the `WorkspaceRunTrigger` configuration. |
| **Security & Identity** | **`serviceAccount` with Setup Permissions** | Creating/updating a trigger requires `team manage workspace` on destination and `team view workspace` on source; automated executions run under system `serviceAccount` with `JobVia.RUN_TRIGGER`. |
| **Execution Lineage** | **Causation Tracking & DAG Querying** | Downstream jobs store `triggeredByJobId` and `cascadeDepth`, enabling end-to-end traceability and DAG visualization in UI and API. |

---

### 3.1 Entity Model (`io.terrakube.api.rs.workspace.trigger.WorkspaceRunTrigger`)

In accordance with **Terrakube Architecture Rule 1 (Elide API Entity Conventions)**, the trigger relationship is exposed as a first-class Yahoo Elide JSON:API and GraphQL resource.

```mermaid
classDiagram
    class Organization {
        +UUID id
        +String name
    }
    class Workspace {
        +UUID id
        +String name
        +String defaultTemplate
        +List~WorkspaceRunTrigger~ runTriggers
        +List~WorkspaceRunTrigger~ sourceRunTriggers
    }
    class WorkspaceRunTrigger {
        +UUID id
        +Workspace sourceWorkspace
        +Workspace destinationWorkspace
        +Template template
        +Organization organization
        +boolean enabled
        +String branchFilter
    }
    class Job {
        +UUID id
        +Workspace workspace
        +JobStatus status
        +String via
        +UUID triggeredByJobId
        +int cascadeDepth
    }

    Organization "1" --> "*" Workspace
    Workspace "1" --> "*" WorkspaceRunTrigger : "destination (runTriggers)"
    Workspace "1" --> "*" WorkspaceRunTrigger : "source (sourceRunTriggers)"
    Workspace "1" --> "*" Job
    Job ..> Job : "triggeredByJobId (lineage)"
```

#### Entity Definition: `WorkspaceRunTrigger.java`
```java
package io.terrakube.api.rs.workspace.trigger;

import com.yahoo.elide.annotation.*;
import io.terrakube.api.plugin.security.audit.GenericAuditFields;
import io.terrakube.api.rs.Organization;
import io.terrakube.api.rs.hooks.workspace.WorkspaceRunTriggerHook;
import io.terrakube.api.rs.template.Template;
import io.terrakube.api.rs.workspace.Workspace;
import jakarta.persistence.*;
import lombok.Getter;
import lombok.Setter;
import org.hibernate.annotations.JdbcTypeCode;

import java.sql.Types;
import java.util.UUID;

@ReadPermission(expression = "team view workspace")
@CreatePermission(expression = "team manage workspace")
@UpdatePermission(expression = "team manage workspace")
@DeletePermission(expression = "team manage workspace")
@LifeCycleHookBinding(operation = LifeCycleHookBinding.Operation.CREATE, phase = LifeCycleHookBinding.TransactionPhase.PRECOMMIT, hook = WorkspaceRunTriggerHook.class)
@LifeCycleHookBinding(operation = LifeCycleHookBinding.Operation.UPDATE, phase = LifeCycleHookBinding.TransactionPhase.PRECOMMIT, hook = WorkspaceRunTriggerHook.class)
@Include
@Getter
@Setter
@Entity(name = "workspace_run_trigger")
@Table(name = "workspace_run_trigger", uniqueConstraints = {
    @UniqueConstraint(name = "uk_run_trigger_source_dest", columnNames = {"source_workspace_id", "destination_workspace_id"})
})
public class WorkspaceRunTrigger extends GenericAuditFields {

    @Id
    @JdbcTypeCode(Types.VARCHAR)
    @GeneratedValue(strategy = GenerationType.UUID)
    private UUID id;

    @ManyToOne(optional = false, fetch = FetchType.LAZY)
    @JoinColumn(name = "source_workspace_id", nullable = false)
    private Workspace sourceWorkspace;

    @ManyToOne(optional = false, fetch = FetchType.LAZY)
    @JoinColumn(name = "destination_workspace_id", nullable = false)
    private Workspace destinationWorkspace;

    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumn(name = "template_id")
    private Template template;

    @ManyToOne(optional = false, fetch = FetchType.LAZY)
    @JoinColumn(name = "organization_id", nullable = false)
    private Organization organization;

    @Column(name = "enabled", nullable = false)
    private boolean enabled = true;

    @Column(name = "branch_filter")
    private String branchFilter;
}
```

### 3.2 Enhancements to Existing Entities

1. **`Workspace.java`**:
   ```java
   // Inbound triggers: Workspaces that trigger THIS workspace
   @OneToMany(mappedBy = "destinationWorkspace", cascade = CascadeType.ALL, orphanRemoval = true)
   private List<WorkspaceRunTrigger> runTriggers;

   // Outbound triggers: Workspaces that THIS workspace triggers
   @OneToMany(mappedBy = "sourceWorkspace", cascade = CascadeType.ALL, orphanRemoval = true)
   private List<WorkspaceRunTrigger> sourceRunTriggers;
   ```

2. **`Job.java`**:
   ```java
   @Column(name = "triggered_by_job_id")
   @JdbcTypeCode(Types.VARCHAR)
   private UUID triggeredByJobId;

   @Column(name = "cascade_depth", nullable = false)
   private int cascadeDepth = 0;
   ```

3. **`JobVia.java`**:
   ```java
   public enum JobVia {
       UI("UI"),
       CLI("CLI"),
       GITHUB("Github"),
       GITLAB("Gitlab"),
       BITBUCKET("Bitbucket"),
       AZURE_DEVOPS("AzureDevops"),
       SCHEDULE("Schedule"),
       RUN_TRIGGER("RunTrigger"); // New Trigger Source
       ...
   }
   ```

### 3.3 Database Migration (Liquibase Mandate)
In accordance with **Rule 2**, a Liquibase changelog is defined:
`terrakube/api/src/main/resources/db/changelog/local/changelog-2.34.0-workspace-run-trigger.xml`:
- Create table `workspace_run_trigger` with UUID primary key, foreign keys to `workspace` (source and destination) and `organization`.
- Add unique constraint `(source_workspace_id, destination_workspace_id)`.
- Add columns `triggered_by_job_id` (varchar) and `cascade_depth` (integer default 0) to table `job`.
- Add index on `workspace_run_trigger(source_workspace_id, enabled)` for sub-millisecond trigger lookups.

---

## 4. Static Cycle Detection & Graph Validation

To eliminate circular dependencies before they can be persisted, `WorkspaceRunTriggerHook` runs a Directed Acyclic Graph (DAG) cycle check during Elide's `PRECOMMIT` phase.

```mermaid
flowchart TD
    Start["User / CLI creates WorkspaceRunTrigger(A -> B)"] --> CheckSelf{"source == destination?"}
    CheckSelf -- Yes --> ErrorSelf["Throw 400 Bad Request: Self-trigger forbidden"]
    CheckSelf -- No --> CheckOrg{"source.org == dest.org?"}
    CheckOrg -- No --> ErrorOrg["Throw 400 Bad Request: Cross-organization triggers forbidden"]
    CheckOrg -- Yes --> LoadGraph["Load all active triggers for Organization"]
    LoadGraph --> AddCandidate["Simulate adding edge (A -> B) to adjacency list"]
    AddCandidate --> DFS["Run Depth-First Search (DFS) / Cycle Detection"]
    DFS --> CycleDetected{"Cycle Found?"}
    CycleDetected -- Yes --> ErrorCycle["Throw 400 Bad Request: Circular dependency detected: A -> ... -> A"]
    CycleDetected -- No --> Commit["Allow Transaction Commit"]
```

### DFS Cycle Detection Algorithm (`WorkspaceGraphValidationService`)
```java
@Service
public class WorkspaceGraphValidationService {

    private final WorkspaceRunTriggerRepository triggerRepository;

    public void validateAcyclic(UUID orgId, UUID candidateSourceId, UUID candidateDestId) {
        if (candidateSourceId.equals(candidateDestId)) {
            throw new InvalidTriggerException("A workspace cannot trigger itself.");
        }

        // Build adjacency map: sourceWorkspaceId -> Set<destinationWorkspaceId>
        List<WorkspaceRunTrigger> existingTriggers = triggerRepository.findByOrganizationIdAndEnabledTrue(orgId);
        Map<UUID, Set<UUID>> adj = new HashMap<>();
        for (WorkspaceRunTrigger t : existingTriggers) {
            adj.computeIfAbsent(t.getSourceWorkspace().getId(), k -> new HashSet<>())
               .add(t.getDestinationWorkspace().getId());
        }

        // Add the prospective edge
        adj.computeIfAbsent(candidateSourceId, k -> new HashSet<>()).add(candidateDestId);

        // Check if there is a path from candidateDestId back to candidateSourceId
        Set<UUID> visited = new HashSet<>();
        if (hasPathDFS(adj, candidateDestId, candidateSourceId, visited)) {
            throw new CyclicDependencyException(String.format(
                "Circular dependency detected. Adding trigger from '%s' to '%s' creates an execution loop.",
                candidateSourceId, candidateDestId
            ));
        }
    }

    private boolean hasPathDFS(Map<UUID, Set<UUID>> adj, UUID current, UUID target, Set<UUID> visited) {
        if (current.equals(target)) return true;
        if (!visited.add(current)) return false;

        for (UUID neighbor : adj.getOrDefault(current, Collections.emptySet())) {
            if (hasPathDFS(adj, neighbor, target, visited)) {
                return true;
            }
        }
        return false;
    }
}
```

---

## 5. Runtime Execution & Reactive Dispatch Engine

When an upstream job successfully completes an apply or destroy operation, the `WorkspaceRunTriggerService` evaluates downstream triggers and schedules subsequent runs.

### 5.1 End-to-End Sequence Diagram

```mermaid
sequenceDiagram
    autonumber
    actor Dev as Developer / VCS
    participant API as Terrakube API
    participant DB as PostgreSQL
    participant Quartz as Quartz Scheduler
    participant Hook as WorkspaceRunTriggerService
    participant Exec as Terrakube Executor

    Dev->>API: Triggers run on Workspace A
    API->>Exec: Dispatches Job A (apply)
    Exec->>Exec: Executes Terraform apply successfully
    Exec->>API: Status callback: setCompletedStatus(Job A)
    API->>DB: UPDATE job SET status='completed' WHERE id=Job A

    Note over API,Hook: Transaction Commit Hook / Direct Invocation
    API->>Hook: onJobCompleted(Job A)
    Hook->>Hook: Validate: Job A is state-changing (not plan-only)? Yes.
    Hook->>Hook: Validate: cascadeDepth < MAX_CASCADE_DEPTH (10)? Yes.
    Hook->>DB: SELECT * FROM workspace_run_trigger WHERE source_workspace_id = WS_A AND enabled = true
    
    loop For each downstream trigger (e.g. Workspace B)
        Hook->>Hook: Check branch filter (if configured)
        Hook->>Hook: Resolve template (Trigger.template OR Workspace.defaultTemplate)
        Hook->>DB: INSERT INTO job (workspace_id=WS_B, status='pending', via='RunTrigger', triggered_by_job_id=Job_A.id, cascade_depth=Job_A.cascade_depth + 1, created_by='serviceAccount')
        Hook->>Quartz: scheduleJobService.createJobContext(Job B)
    end

    Note over Quartz,Exec: Two-Tier FIFO Queue Takes Over
    Quartz->>API: ScheduleJob.execute(Job B)
    API->>DB: Check FIFO admission & Workspace serialization
    API->>Exec: Dispatch Job B when executor capacity and queue permit
```

### 5.2 Trigger Service Implementation (`WorkspaceRunTriggerService`)

```java
@Slf4j
@Service
@RequiredArgsConstructor
public class WorkspaceRunTriggerService {

    public static final int MAX_CASCADE_DEPTH = 10;

    private final WorkspaceRunTriggerRepository triggerRepository;
    private final JobRepository jobRepository;
    private final TemplateRepository templateRepository;
    private final ScheduleJobService scheduleJobService;
    private final JobNotificationTrigger jobNotificationTrigger;

    public void processDownstreamTriggers(Job completedJob) {
        if (completedJob.getStatus() != JobStatus.completed) {
            return;
        }

        // 1. Filter out speculative plan-only jobs
        if (isPlanOnlyJob(completedJob)) {
            log.debug("Job {} on workspace {} was plan-only. Skipping downstream triggers.", 
                      completedJob.getId(), completedJob.getWorkspace().getName());
            return;
        }

        // 2. Enforce runtime hop limit to prevent runaway cascades
        if (completedJob.getCascadeDepth() >= MAX_CASCADE_DEPTH) {
            log.error("Job {} reached max cascade depth ({}); halting downstream trigger propagation.", 
                      completedJob.getId(), MAX_CASCADE_DEPTH);
            return;
        }

        Workspace sourceWorkspace = completedJob.getWorkspace();
        List<WorkspaceRunTrigger> triggers = triggerRepository
                .findBySourceWorkspaceAndEnabledTrue(sourceWorkspace);

        if (triggers.isEmpty()) {
            return;
        }

        log.info("Processing {} downstream triggers for completed job {} on workspace {}",
                 triggers.size(), completedJob.getId(), sourceWorkspace.getName());

        for (WorkspaceRunTrigger trigger : triggers) {
            try {
                triggerDownstreamWorkspace(trigger, completedJob);
            } catch (Exception e) {
                log.error("Failed to execute run trigger {} for downstream workspace {}: {}",
                          trigger.getId(), trigger.getDestinationWorkspace().getName(), e.getMessage(), e);
            }
        }
    }

    private void triggerDownstreamWorkspace(WorkspaceRunTrigger trigger, Job sourceJob) throws Exception {
        Workspace destWorkspace = trigger.getDestinationWorkspace();

        if (destWorkspace.isLocked()) {
            log.warn("Downstream workspace {} is locked; skipping triggered run from job {}", 
                     destWorkspace.getName(), sourceJob.getId());
            return;
        }

        // Create downstream Job entity
        Job downstreamJob = new Job();
        downstreamJob.setRefresh(true);
        downstreamJob.setRefreshOnly(false);
        downstreamJob.setWorkspace(destWorkspace);
        downstreamJob.setOrganization(destWorkspace.getOrganization());
        downstreamJob.setStatus(JobStatus.pending);
        downstreamJob.setCreatedBy("serviceAccount");
        downstreamJob.setUpdatedBy("serviceAccount");
        downstreamJob.setVia(JobVia.RUN_TRIGGER.getValue());
        downstreamJob.setTriggeredByJobId(sourceJob.getId());
        downstreamJob.setCascadeDepth(sourceJob.getCascadeDepth() + 1);

        // Resolve template: trigger override -> workspace default -> org default
        String templateId = trigger.getTemplate() != null 
                ? trigger.getTemplate().getId().toString() 
                : destWorkspace.getDefaultTemplate();

        if (templateId != null && !templateId.isBlank()) {
            Template template = templateRepository.getReferenceById(UUID.fromString(templateId));
            downstreamJob.setTcl(template.getTcl());
            downstreamJob.setTemplateReference(templateId);
        }

        Date now = new Date();
        downstreamJob.setCreatedDate(now);
        downstreamJob.setUpdatedDate(now);

        downstreamJob = jobRepository.save(downstreamJob);
        jobNotificationTrigger.notifyStatusChanged(downstreamJob);

        log.info("Created downstream Job {} (depth {}) for workspace {} triggered by Job {}",
                 downstreamJob.getId(), downstreamJob.getCascadeDepth(), destWorkspace.getName(), sourceJob.getId());

        // Enqueue into Quartz execution context
        scheduleJobService.createJobContext(downstreamJob);
    }

    private boolean isPlanOnlyJob(Job job) {
        // Evaluate template definition / TCL or plan-only indicator
        return job.getTcl() != null && !job.getTcl().contains("type: \"terraformApply\"");
    }
}
```

---

## 6. Integration with Two-Tier Queue & Distributed Locking

A key architectural strength of this design is that **triggered downstream jobs do not require a separate execution engine**. Once `downstreamJob` is persisted and enqueued via `scheduleJobService.createJobContext(downstreamJob)`, it automatically obeys all existing Terrakube scheduling rules:

1. **Workspace Serialization (Tier 1)**:
   - If Workspace C has an ongoing run or another queued run, the new run backs off and waits for its turn.
   - Independent reactive triggers never cause race conditions in the workspace state.
2. **Global FIFO Queue & Admission Check (Tier 2)**:
   - Evaluated by `isJobNextInDispatchOrderExecutable`. Downstream jobs are allocated sequential integer primary keys and processed fairly across the cluster.
3. **Redis Distributed Execution Lock**:
   - `ScheduleJob.java` acquires `job-execution-lock:<jobId>` with 90s TTL upon tick, releases immediately upon HTTP 202 Accepted from the executor.

---

### 7.1 JSON:API Endpoints
- **Create Trigger**:
  `POST /api/v1/organization/{orgId}/workspace/{destWsId}/runTrigger`
  ```json
  {
    "data": {
      "type": "workspace_run_trigger",
      "attributes": {
        "enabled": true
      },
      "relationships": {
        "sourceWorkspace": {
          "data": { "type": "workspace", "id": "b3c2d1e0-..." }
        },
        "template": {
          "data": { "type": "template", "id": "f5e4d3c2-..." }
        }
      }
    }
  }
  ```
- **List Triggers for Workspace**:
  `GET /api/v1/organization/{orgId}/workspace/{destWsId}/runTrigger`
- **Delete Trigger**:
  `DELETE /api/v1/workspace_run_trigger/{triggerId}`

### 7.2 Elide GraphQL Dependency Graph Query
Clients can query the entire dependency graph in a single query:
```graphql
query GetWorkspaceDependencyGraph($orgId: String!) {
  organization(id: $orgId) {
    workspace {
      edges {
        node {
          id
          name
          lastJobStatus
          runTriggers {
            edges {
              node {
                id
                enabled
                sourceWorkspace {
                  id
                  name
                }
              }
            }
          }
          sourceRunTriggers {
            edges {
              node {
                id
                enabled
                destinationWorkspace {
                  id
                  name
                }
              }
            }
          }
        }
      }
    }
  }
}
```

### 7.3 CLI & SDK Extensions
- **`terrakube-cli`**:
  Uses `resource.Register[WorkspaceRunTrigger]()` to automatically provide:
  - `terrakube run-trigger list --workspace <ws-id>`
  - `terrakube run-trigger create --source-workspace <ws-id> --destination-workspace <ws-id>`
  - `terrakube run-trigger delete --id <trigger-id>`
- **`terrakube-go`**:
  Adds `WorkspaceRunTrigger` struct, `RunTriggerService` with `List()`, `Get()`, `Create()`, and `Delete()` methods.

---

---

## 8. Terrakube UI Architecture & React Flow Visualization

Terrakube's frontend already uses **React Flow** (`reactflow: ^11.11.3`) to render the Terraform state DAG in `domain/Workspaces/States.tsx` with custom nodes (`NodeResource.tsx`). We can directly reuse this exact library and architecture to build an interactive, visual representation of workspace dependencies.

### 8.1 UI Locations for Dependency Visualization

1. **Workspace Settings: Run Triggers Management Tab**
   - New section under `WorkspaceSettings.tsx`: `WORKSPACE_SETTINGS_SECTION_LABELS["run-triggers"] = "Run Triggers"`.
   - Contains:
     - A Table listing configured upstream sources and downstream consumers.
     - "Add Run Trigger" Modal with dropdown of workspaces in the same Organization.
     - A toggle switch to display the **Visual Dependency Graph** for this workspace.

2. **Workspace Details: Local Dependency Subgraph**
   - Shows the immediate upstream producers, the current workspace in the center, and downstream consumers:
     `[VPC Workspace] ────► [Current Workspace] ────► [App Workspace]`
   - Each node displays live workspace status (`WorkspaceStatusTag`), branch name, and last execution date.

3. **Organization / Project Level: Global Workspace DAG**
   - A dedicated "Dependencies" tab at the Project/Organization view that fetches all workspaces and active run triggers.
   - Renders the full multi-tier infrastructure execution graph across all workspaces.

4. **Run Details Lineage Banner (`DetailsJob.tsx`)**
   - When viewing a Job where `job.via === "RunTrigger"`, display an Ant Design Alert banner:
     *"Triggered automatically by upstream run #[Job ID] in workspace [Source Workspace Name]"* with direct navigation links.

### 8.2 Custom React Flow Component: `NodeWorkspace.tsx`

Patterned after `NodeResource.tsx`, `NodeWorkspace.tsx` renders an Ant Design Card representing a workspace node with handles for directed edges:

```tsx
import { memo } from "react";
import { Card, Space, Typography, Button } from "antd";
import { Handle, NodeProps, Position } from "reactflow";
import { Link } from "react-router-dom";
import WorkspaceStatusTag from "@/components/display/WorkspaceStatusTag/WorkspaceStatusTag";
import { HiOutlineExternalLink } from "react-icons/hi";

const { Text } = Typography;

export interface WorkspaceNodeData {
  id: string;
  name: string;
  orgId: string;
  status: string;
  branch: string;
  isCurrent?: boolean;
}

export default memo(({ data, isConnectable }: NodeProps<WorkspaceNodeData>) => {
  return (
    <>
      {/* Inbound Handle (from upstream producers) */}
      <Handle
        type="target"
        position={Position.Left}
        style={{ background: "#1677ff", width: 10, height: 10 }}
        isConnectable={isConnectable}
      />

      <Card
        size="small"
        style={{
          width: 260,
          border: data.isCurrent ? "2px solid #1677ff" : "1px solid #d9d9d9",
          borderRadius: 8,
          boxShadow: data.isCurrent ? "0 0 10px rgba(22, 119, 255, 0.35)" : "none",
        }}
      >
        <Space direction="vertical" style={{ width: "100%" }} size={4}>
          <Space style={{ justifyContent: "space-between", width: "100%" }}>
            <Link
              to={`/organizations/${data.orgId}/workspaces/${data.id}`}
              style={{ fontWeight: 600, fontSize: 14 }}
            >
              {data.name}
            </Link>
            <HiOutlineExternalLink />
          </Space>

          <Space size={8}>
            <WorkspaceStatusTag status={data.status} />
            <Text type="secondary" style={{ fontSize: 12 }}>
              {data.branch}
            </Text>
          </Space>
        </Space>
      </Card>

      {/* Outbound Handle (to downstream consumers) */}
      <Handle
        type="source"
        position={Position.Right}
        style={{ background: "#52c41a", width: 10, height: 10 }}
        isConnectable={isConnectable}
      />
    </>
  );
});
```

### 8.3 Graph Layout & Edge Construction

Following the pattern in `States.tsx`, we construct the nodes and edges:
- **Left-to-Right Topological Layout**:
  - Upstream workspaces at `x = 50`, `y = index * 100`
  - Current workspace at `x = 400`, `y = 150`
  - Downstream workspaces at `x = 750`, `y = index * 100`
- **Edges**:
  - Animated bezier edges with `MarkerType.ArrowClosed`:
  ```ts
  edges.push({
    id: `${trigger.sourceWorkspace.id}-${trigger.destinationWorkspace.id}`,
    source: trigger.sourceWorkspace.id,
    target: trigger.destinationWorkspace.id,
    animated: trigger.sourceWorkspace.lastJobStatus === "running",
    markerEnd: { type: MarkerType.ArrowClosed },
    style: { stroke: "#1677ff", strokeWidth: 2 },
  });
  ```

---

## 9. Failure Modes & Resilience Matrix

| Failure Scenario | Mitigation / Behavior |
|---|---|
| **Circular dependency created (A &rarr; B &rarr; A)** | Blocked statically at API creation/update time via DFS cycle detection in `WorkspaceRunTriggerHook`. Returns HTTP 400 with cycle path details. |
| **Runaway cascade / cascade storm** | Hard runtime limit (`cascadeDepth >= 10`) stops further propagation and emits a structured error log. |
| **Downstream workspace is locked** | Skipped gracefully with warning log; prevents stuck jobs and allows unlock before manual rerun. |
| **Downstream job fails execution** | The failure is isolated to the downstream workspace. It will NOT trigger its own downstream workspaces (since only `completed` runs qualify), naturally isolating the blast radius. |
| **Multiple upstream workspaces finish simultaneously** | Each event creates a pending job on the downstream workspace. Terrakube's per-workspace serialization queue serializes them, executing them sequentially without Terraform state locks colliding. |
| **API replica crashes during trigger dispatch** | Job creation and trigger dispatch occur within a transaction; if interrupted before commit, no partial state remains. If after commit, Quartz cluster recovery recovers the scheduled job context. |

---

## 10. Rollout & Phased Implementation Plan

1. **Phase 1: Persistence & Liquibase (Backend Foundation)**
   - Create Liquibase migration `changelog-2.34.0-workspace-run-trigger.xml`.
   - Register in `changelog.xml`.
   - Add `WorkspaceRunTrigger` Elide entity under `io.terrakube.api.rs.workspace.trigger`.
   - Update `Workspace`, `Job`, and `JobVia` entities.
2. **Phase 2: Validation & Cycle Detection**
   - Implement `WorkspaceGraphValidationService` (DFS cycle check).
   - Implement `WorkspaceRunTriggerHook` with security and cycle checks.
   - Unit tests for acyclic validation and self-trigger rejection.
3. **Phase 3: Reactive Dispatch Engine**
   - Implement `WorkspaceRunTriggerService`.
   - Hook into `ScheduleJob.java` and executor completion status callbacks (alongside `JobNotificationTrigger`).
   - Unit and integration tests for downstream job generation and hop limit enforcement.
4. **Phase 4: Client & UI Integration**
   - Expose run triggers in `terrakube-go` and `terrakube-cli`.
   - UI: Add "Run Triggers" tab in Workspace Settings with React Flow dependency graph (`NodeWorkspace.tsx`).
   - Add upstream run causation banner in `DetailsJob.tsx`.

### Anything else?

_No response_

klinux:
Implements the workspace dependency graph and run triggers designed in #3520, built and reviewed in four phases against that thread.

A run trigger is a directed edge between two workspaces of the same organization: when the **source** finishes a run that changed remote state, the **destination** gets a run enqueued. It closes the gap that otherwise leaves people applying a VPC and then remembering to go apply the VPN that reads its outputs.

### The edge (phase 1)

`WorkspaceRunTrigger` is exposed as `runTrigger`, with `sourceWorkspace`, `destinationWorkspace`, an optional `template` override and an `enabled` flag.

**Permissions.** Creating or changing an edge requires manage rights on the **destination** — the workspace whose runs start firing — plus visibility of the source. Without the second half, anyone with rights on a single workspace could attach it to any other and learn when that one applies. Deletion asks only for the destination, so losing access to a source cannot strand a trigger on a workspace you own. All three tiers of Terrakube's model are honoured: organization team, `ProjectAccess`, and workspace-level `Access`.

**Immutability.** Both endpoints carry `@UpdatePermission("user is a superuser")`. Repointing an edge is deleting one dependency and declaring another, and each deserves its own check; `enabled` and `template` stay editable.

**Tenancy.** The owning organization is derived from the destination and `@Exclude`d from the API surface rather than merely guarded — a client has no reason to set it, and accepting it would let an edge be attributed to a tenant owning neither workspace. It is derived in a JPA `@PrePersist` callback, not an Elide hook: `PRECOMMIT` fires after Hibernate has already flushed, which leaves the `NOT NULL` column unset.

### Cycle validation (phase 2)

A `PRECOMMIT` hook rejects any edge that would close a loop, with the offending path in the message.

Four decisions worth naming. Validation considers **disabled edges too** — otherwise you could create A→B disabled, then B→A, then enable A→B, and no validation ever saw the cycle. The edge being written is **excluded from the graph** it is validated against, since at `PRECOMMIT` its own row is already flushed. The traversal is **iterative, not recursive**, and visited-guarded, so a pre-existing cycle cannot hang it; there is a test with a 10,000 node chain. And the graph loads as an **id projection**, so validating never materialises workspaces.

Known limit, deliberately accepted: two concurrent creations can each pass validation and together close a loop. Closing that would need a lock over the whole organization. The runtime cascade limit is the net.

### Dispatch (phase 3)

Anchored in `JobReconciliationService`, not `ScheduleJob`. That service is the only routine that performs a job's transition to a terminal status, it takes the transition under `lockForUpdate`, and an already-terminal job short-circuits before reaching the dispatch call. **Exactly-once per completed run, across replicas, with no dedup table.** It also covers every completion path for free — inline, the 30s sweep, the admin endpoint, and CLI/TFE runs, which hand their job to the scheduler as pending and terminate through the same routine.

Dispatch runs **after commit, in its own transaction**, and never throws: a trigger that cannot be dispatched must not roll back the run that fired it.

**Qualification reads step outcomes, not job status.** A plan/apply template whose apply never executed still finishes as `completed`, so the engine requires a step that is both `completed` and of a state-changing flow type — `terraformApply`, `terraformDestroy` or `customScripts`, the set `ScheduleJob` already uses for queue admission. A plan on its own fires nothing.

**Each dependent commits in its own transaction** (`RunTriggerJobWriter`), and its Quartz context is created only afterwards. Both halves matter: `createJobContext` fires the trigger immediately and the worker reads on another connection, so an uncommitted row is invisible and `ScheduleJob` would treat the job as deleted and drop the trigger permanently — the trap Elide avoids by binding `JobManageHook` to `POSTCOMMIT`. And a failed insert marks its transaction rollback-only, which catching the exception does not clear, so a shared transaction would take every sibling down with one bad dependent.

**A locked destination enqueues rather than skips.** Skipping would leave the dependent silently drifted, which is the problem the feature exists to solve; `ScheduleJob` admits the pending job once the lock clears. **No coalescing**: consecutive runs evaluate quickly with no changes, and debouncing would add a timeout with failure modes of its own.

Bounds under `io.terrakube.run-trigger`: `enabled` (true), `max-cascade-depth` (10) as the runtime net for a cycle that slipped past validation, and `max-dependents-per-apply` (20), which truncates a stable prefix ordered by destination id so an over-wide graph dispatches the same subset on every apply rather than an arbitrary one.

### UI (phase 4)

The Run Triggers page splits the graph the way the **permission model** does, not the way the data does. "Runs after" — edges where this workspace is the destination — is configurable here, because this is the workspace you already have rights on. "Triggers" — where it is the source — is read-only and links to the other workspace, since removing one of those needs rights over there. The UI therefore never offers an action the API would refuse.

Both halves come from one request: a comma is `OR` in Elide's filter syntax, so `sourceWorkspace.id==X,destinationWorkspace.id==X` returns both directions, and the include block resolves names without a follow-up call.

Runs started by a trigger now say so, linking to the upstream run that changed state — `triggeredByJobId` and `cascadeDepth` have existed since phase 1 and nothing displayed either.

### Demo data

A `simple-trigger` organization with 38 workspaces and 28 edges seeds every canonical shape: fan-out, chain, fan-in, diamond, a disabled edge, a template override, standalone workspaces for the empty state, and a 12-step cascade that runs past the depth limit. Every workspace points at `terrakube-docker-compose`, whose root module is a `null_resource` and a 30 second sleep, so a cascade can be watched end to end with no cloud credentials.

## Two defects found and fixed during review

**Elide evaluates an inverse relationship's `UpdatePermission` on create.** The `runTriggers` collections on `Workspace` were guarded with `user is a superuser`; because Elide maintains the bidirectional relationship when a trigger is created, that rule fired on the ordinary create path. No non-superuser could create a trigger at all, and all three permission tiers were unreachable code. The inverse side now mirrors what `Workspace.job` already does. `cascade`/`orphanRemoval` went too: workspaces are soft deleted, so the cascade never fired for the case it appeared to cover, while orphanRemoval turned dropping an element from the collection into a row delete that bypassed `DeletePermission`.

**Elide answers 400 to a bodyless DELETE that declares a content type.** The page's delete was copied from Schedules, which sets one — so deleting a trigger from the UI would have failed, and nothing would have caught it, because the API test deletes without the header. There is now a test asserting the 400 with the reason written down.

## Testing

**1121 API tests, 0 failures. 470 UI tests, 0 failures.**

Beyond the unit coverage, `RunTriggerDispatchIntegrationTest` drives the real `reconcile` against the database — no executor needed — and reads the seeded graph: fan-out reaches all five dependents, a chain advances one step per run, a disabled edge dispatches nothing, and the diamond walks 0 → 1 → 2 runs on the join as each branch applies.

Verified in a browser on Codespaces against Postgres: both directions render on the correct side, and create, toggle and delete all work.

## Notes

- Rebased onto current `main`. `formatJobVia` from #3539 landed while this branch was open and fixes the same hardcoded heading, better; that version won the rebase and `RunTrigger` was added to it.
- `Schedules.tsx` and `Variables.tsx` set the same content type on their deletes. Left alone as out of scope, but the same 400 applies if axios ever stops dropping the header.
- The MinIO checksum conflict hit while setting up the devcontainer was fixed independently on `main` by #3529.

Closes #3520.

### DIFF
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/reconciliation/JobReconciliationService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/reconciliation/JobReconciliationService.java
@@ -2,6 +2,7 @@ package io.terrakube.api.plugin.scheduler.reconciliation;
 
 import io.terrakube.api.plugin.notification.JobNotificationTrigger;
 import io.terrakube.api.plugin.scheduler.ScheduleJobService;
+import io.terrakube.api.plugin.scheduler.trigger.RunTriggerDispatchService;
 import io.terrakube.api.plugin.scheduler.reconciliation.ReconciliationResult.ReconciliationDisposition;
 import io.terrakube.api.plugin.scheduler.reconciliation.ReconciliationResult.StepEvidence;
 import io.terrakube.api.repository.JobRepository;
@@ -55,6 +56,7 @@ public class JobReconciliationService {
     private final ScheduleJobService scheduleJobService;
     private final Scheduler scheduler;
     private final JobReconciliationMetrics metrics;
+    private final RunTriggerDispatchService runTriggerDispatchService;
 
     @Transactional
     public ReconciliationResult reconcile(int jobId, boolean dryRun) {
@@ -103,6 +105,9 @@ public class JobReconciliationService {
         log.info("Reconciled job {} from {} to {} (derived {})", jobId, from, target, outcome);
 
         deleteTriggerAfterCommit(jobId);
+        if (target == JobStatus.completed) {
+            dispatchRunTriggersAfterCommit(jobId);
+        }
         return result(jobId, job, outcome, target, ReconciliationDisposition.APPLIED, evidence);
     }
 
@@ -153,6 +158,20 @@ public class JobReconciliationService {
         workspaceRepository.save(workspace);
     }
 
+    /**
+     * Fans out to the workspaces that depend on this one. Deliberately anchored here rather
+     * than in ScheduleJob: this method is the only routine that performs the transition to a
+     * terminal status, it does so under a row lock, and an already-terminal job never reaches
+     * this point - so a completed run dispatches its triggers exactly once even with several
+     * API replicas competing for the same job.
+     *
+     * <p>After commit, like the Quartz cleanup above, so that a dispatch problem can never
+     * roll back the completion of the run that fired it.
+     */
+    private void dispatchRunTriggersAfterCommit(int jobId) {
+        afterCommit(() -> runTriggerDispatchService.dispatchFor(jobId));
+    }
+
     private void deleteTriggerAfterCommit(int jobId) {
         Runnable delete = () -> {
             try {
@@ -169,15 +188,23 @@ public class JobReconciliationService {
                 log.warn("Could not remove Quartz trigger for reconciled job {}: {}", jobId, e.getMessage());
             }
         };
+        afterCommit(delete);
+    }
+
+    /**
+     * Runs the action once this method's transaction has committed, or immediately when there
+     * is none to wait for (the admin endpoint calls in without one).
+     */
+    private void afterCommit(Runnable action) {
         if (TransactionSynchronizationManager.isSynchronizationActive()) {
             TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
                 @Override
                 public void afterCommit() {
-                    delete.run();
+                    action.run();
                 }
             });
         } else {
-            delete.run();
+            action.run();
         }
     }
 }
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/trigger/CyclicDependencyException.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/trigger/CyclicDependencyException.java
@@ -1,0 +1,15 @@
+package io.terrakube.api.plugin.scheduler.trigger;
+
+import com.yahoo.elide.core.exceptions.HttpStatusException;
+import org.apache.hc.core5.http.HttpStatus;
+
+/**
+ * Raised when a run trigger would close a loop in the organization's dependency graph.
+ * Answered as 400: the request is well formed, the resulting graph is not.
+ */
+public class CyclicDependencyException extends HttpStatusException {
+
+    public CyclicDependencyException(String message) {
+        super(HttpStatus.SC_BAD_REQUEST, message);
+    }
+}
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/trigger/RunTriggerDispatchService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/trigger/RunTriggerDispatchService.java
@@ -1,0 +1,223 @@
+package io.terrakube.api.plugin.scheduler.trigger;
+
+import io.terrakube.api.plugin.notification.JobNotificationTrigger;
+import io.terrakube.api.plugin.scheduler.ScheduleJobService;
+import io.terrakube.api.plugin.scheduler.job.tcl.TclService;
+import io.terrakube.api.plugin.scheduler.job.tcl.model.FlowType;
+import io.terrakube.api.repository.JobRepository;
+import io.terrakube.api.repository.StepRepository;
+import io.terrakube.api.repository.WorkspaceRunTriggerRepository;
+import io.terrakube.api.rs.job.Job;
+import io.terrakube.api.rs.job.JobStatus;
+import io.terrakube.api.rs.job.JobVia;
+import io.terrakube.api.rs.job.step.Step;
+import io.terrakube.api.rs.workspace.Workspace;
+import io.terrakube.api.rs.workspace.trigger.WorkspaceRunTrigger;
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.Comparator;
+import java.util.Date;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Enqueues the downstream runs of a workspace whose run just changed state.
+ *
+ * <p>Called by {@link io.terrakube.api.plugin.scheduler.reconciliation.JobReconciliationService}
+ * after it commits a job's transition to {@code completed}, which is the one routine in the
+ * system that performs that transition. That placement is what gives dispatch its
+ * exactly-once property: the transition is taken under a row lock and an already-terminal job
+ * short-circuits, so only the caller that actually moved the job reaches this service, even
+ * with several API replicas racing on the same job.
+ *
+ * <p>Runs in its own transaction after the commit rather than inside it. A trigger that
+ * cannot be dispatched must never roll back the completion of the run that fired it, and the
+ * Quartz context created for each downstream job is not transactional to begin with.
+ */
+@Slf4j
+@Service
+@AllArgsConstructor
+public class RunTriggerDispatchService {
+
+    /**
+     * Flow types that change remote state. A run made only of plans leaves nothing for a
+     * dependent to react to, so it does not fire triggers. Mirrors the qualification
+     * ScheduleJob.isActiveApplyOrDestroyRunning already applies for queue admission.
+     */
+    private static final Set<String> STATE_CHANGING_FLOWS = Set.of(
+            FlowType.terraformApply.toString(),
+            FlowType.terraformDestroy.toString(),
+            FlowType.customScripts.toString());
+
+    private final JobRepository jobRepository;
+    private final StepRepository stepRepository;
+    private final WorkspaceRunTriggerRepository workspaceRunTriggerRepository;
+    private final TclService tclService;
+    private final JobNotificationTrigger jobNotificationTrigger;
+    private final ScheduleJobService scheduleJobService;
+    private final RunTriggerProperties properties;
+
+    /**
+     * Fans out from a job that has just completed. Takes the id rather than the entity because
+     * the caller invokes this after its own transaction has committed: re-reading here keeps
+     * every association attached to a live session instead of relying on what happened to be
+     * initialised upstream.
+     *
+     * <p>Never throws. The upstream run is already finished and recorded; a dispatch problem
+     * is an operational event to be logged, not a reason to surface an error on a run that
+     * succeeded.
+     */
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
+    public void dispatchFor(int completedJobId) {
+        try {
+            dispatchInternal(completedJobId);
+        } catch (Exception e) {
+            log.error("Run trigger dispatch failed for job {}: {}", completedJobId, e.getMessage(), e);
+        }
+    }
+
+    private void dispatchInternal(int completedJobId) {
+        if (!properties.isEnabled()) {
+            return;
+        }
+
+        Job completedJob = jobRepository.findById(completedJobId).orElse(null);
+        if (completedJob == null || completedJob.getWorkspace() == null) {
+            return;
+        }
+
+        if (!changedState(completedJob)) {
+            log.debug("Job {} changed no state, no run trigger dispatched", completedJobId);
+            return;
+        }
+
+        // Checked before loading the edges: at the limit the answer is the same for every
+        // dependent, and saying so once is more useful than repeating it per edge.
+        int nextDepth = completedJob.getCascadeDepth() + 1;
+        if (nextDepth > properties.getMaxCascadeDepth()) {
+            log.warn("Job {} reached the run trigger cascade limit of {}, not dispatching further",
+                    completedJobId, properties.getMaxCascadeDepth());
+            return;
+        }
+
+        List<WorkspaceRunTrigger> triggers = workspaceRunTriggerRepository
+                .findEnabledBySourceWorkspaceId(completedJob.getWorkspace().getId());
+        if (triggers.isEmpty()) {
+            return;
+        }
+
+        List<WorkspaceRunTrigger> dispatchable = applyFanOutLimit(triggers, completedJob);
+
+        for (WorkspaceRunTrigger trigger : dispatchable) {
+            // Per-dependent isolation: one workspace missing a template, or failing to
+            // schedule, must not cost the dependents that come after it in the list.
+            try {
+                enqueue(trigger, completedJob, nextDepth);
+            } catch (Exception e) {
+                log.error("Could not enqueue run triggered by job {} on workspace {}: {}",
+                        completedJobId, destinationName(trigger), e.getMessage(), e);
+            }
+        }
+    }
+
+    /**
+     * True when at least one step that actually ran changed remote state. Steps are checked
+     * individually rather than trusting the job status: a plan/apply template whose apply was
+     * never executed still finishes as completed, and has nothing downstream to react to.
+     */
+    private boolean changedState(Job job) {
+        List<Step> steps = stepRepository.findByJobId(job.getId());
+        for (Step step : steps) {
+            if (step.getStatus() != JobStatus.completed) {
+                continue;
+            }
+            String flowType = tclService.getFlowTypeForStep(job, step.getStepNumber());
+            if (flowType != null && STATE_CHANGING_FLOWS.contains(flowType)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Caps the fan-out, keeping a stable prefix so that a graph over the limit dispatches the
+     * same dependents on every apply instead of a different arbitrary subset each time.
+     */
+    private List<WorkspaceRunTrigger> applyFanOutLimit(List<WorkspaceRunTrigger> triggers, Job completedJob) {
+        int limit = properties.getMaxDependentsPerApply();
+        if (triggers.size() <= limit) {
+            return triggers;
+        }
+
+        List<WorkspaceRunTrigger> ordered = triggers.stream()
+                .sorted(Comparator.comparing(t -> t.getDestinationWorkspace().getId()))
+                .toList();
+        List<WorkspaceRunTrigger> skipped = ordered.subList(limit, ordered.size());
+        log.warn("Workspace {} has {} enabled run triggers, above the limit of {}. Job {} will not "
+                        + "dispatch to: {}",
+                completedJob.getWorkspace().getName(), triggers.size(), limit, completedJob.getId(),
+                skipped.stream().map(RunTriggerDispatchService::destinationName).toList());
+        return ordered.subList(0, limit);
+    }
+
+    private void enqueue(WorkspaceRunTrigger trigger, Job completedJob, int depth) throws Exception {
+        Workspace destination = trigger.getDestinationWorkspace();
+
+        String templateReference = resolveTemplate(trigger, destination);
+        if (templateReference == null || templateReference.isBlank()) {
+            log.warn("Run trigger {} has no template and workspace {} has no default template, skipping",
+                    trigger.getId(), destination.getName());
+            return;
+        }
+
+        Date now = new Date(System.currentTimeMillis());
+        Job job = new Job();
+        job.setWorkspace(destination);
+        job.setOrganization(destination.getOrganization());
+        job.setTemplateReference(templateReference);
+        job.setStatus(JobStatus.pending);
+        job.setRefresh(true);
+        job.setPlanChanges(true);
+        job.setRefreshOnly(false);
+        job.setVia(JobVia.RUN_TRIGGER.getValue());
+        job.setTriggeredByJobId(completedJob.getId());
+        job.setCascadeDepth(depth);
+        job.setCreatedBy("serviceAccount");
+        job.setUpdatedBy("serviceAccount");
+        job.setCreatedDate(now);
+        job.setUpdatedDate(now);
+
+        Job savedJob = jobRepository.save(job);
+
+        // A plain repository save bypasses Elide, so neither JobNotificationHook nor
+        // JobManageHook fires for this job: the status event and the Quartz context both have
+        // to be raised here, exactly as ScheduleJobTrigger does for scheduled runs.
+        jobNotificationTrigger.notifyStatusChanged(savedJob);
+        scheduleJobService.createJobContext(savedJob);
+
+        log.info("Job {} triggered job {} on workspace {} (depth {})",
+                completedJob.getId(), savedJob.getId(), destination.getName(), depth);
+    }
+
+    /**
+     * The trigger's own template wins, then the destination's default. A locked destination is
+     * deliberately not special-cased: the job is enqueued as pending and ScheduleJob admits it
+     * once the lock clears, because skipping would leave the dependent silently drifted, which
+     * is the very thing run triggers exist to prevent.
+     */
+    private String resolveTemplate(WorkspaceRunTrigger trigger, Workspace destination) {
+        if (trigger.getTemplate() != null) {
+            return trigger.getTemplate().getId().toString();
+        }
+        return destination.getDefaultTemplate();
+    }
+
+    private static String destinationName(WorkspaceRunTrigger trigger) {
+        Workspace destination = trigger.getDestinationWorkspace();
+        return destination != null ? destination.getName() : "unknown";
+    }
+}
--- a/api/src/main/java/io/terrakube/api/plugin/scheduler/trigger/RunTriggerDispatchService.java
+++ b/api/src/main/java/io/terrakube/api/plugin/scheduler/trigger/RunTriggerDispatchService.java
@@ -9,18 +9,14 @@ import io.terrakube.api.repository.StepRepository;
 import io.terrakube.api.repository.WorkspaceRunTriggerRepository;
 import io.terrakube.api.rs.job.Job;
 import io.terrakube.api.rs.job.JobStatus;
-import io.terrakube.api.rs.job.JobVia;
 import io.terrakube.api.rs.job.step.Step;
 import io.terrakube.api.rs.workspace.Workspace;
 import io.terrakube.api.rs.workspace.trigger.WorkspaceRunTrigger;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
-import org.springframework.transaction.annotation.Propagation;
-import org.springframework.transaction.annotation.Transactional;
 
 import java.util.Comparator;
-import java.util.Date;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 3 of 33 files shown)


alfespa17 (approved):

alfespa17:
thank you @klinux

PR #3555, opened 2026-09-14
fix(api): lift Elide's implicit 500-row cap on workspace collections

### LINKED ISSUE #3552: Unable to list beyond 500 workspaces in UI
### Bug description 🐞

My terrakube installation has a single organisation that has ~1500 worksapces created.

These workspaces were defined en-masse via API (specifically the Terraform Provider). 

Once all these workspaces were created, the UI could only show 500 max.

Image

The workspaces seems to exist and I still search for them and access their individual pages. Except I cannot list them fully.

### Steps to reproduce

1. Create >550 Workspaces in a single organistion
2. Load Terrakube Workspace listing

### Expected behavior

All Workspaces are listed in the web UI and is not capped at 500

### Example repository

n/a

### Anything else?

Terrakube 2.33.1

It seems to imply the API has a 500 item limit with no pagination as I encounter the same issue when trying to list all my workspaces through the API

The only way I managed to get all my workspaces via the API was through this logic (in Python)

> The workspace collection endpoint silently caps at 500 records and rejects every `page[...]` parameter with "Cannot paginate workspace",  and returns no `meta`/`links`. 
> The organization record, however, lists every workspace ID, and RSQL filtering on the collection works. 
> So:
>  - take the ID list from the organization and fetch the records in batches with `filter[workspace]=id=in=(...)`, verifying each batch.

alfespa17:
I created 1 org with 1500 workspaces, 15 pages each with 100 workspaces

imagen

imagen

imagen

The PR is based on https://github.com/mechastorm/terrakube/pull/1 + a test case and some test information for the demo profile


alfespa17:
Closing the PR, maybe it is not a good idea to add the sample org with 1500 worspaces

PR #3564, opened 2026-09-15
feat(executor): make ephemeral job deadline, backoff limit, ttl, and grace period configurable

### LINKED ISSUE #3541: Expose activeDeadlineSeconds (and backoffLimit, ttlSecondsAfterFinished) for ephemeral jobs
### Problem

Ephemeral Terraform jobs have no wall-clock timeout, and there is no way to give them one.

In `EphemeralExecutorService`, the Job spec builder:

- hardcodes `.withTtlSecondsAfterFinished(30)` — cleanup of *finished* jobs, which is fine but not adjustable
- never sets `activeDeadlineSeconds`, so nothing bounds a *running* job
- never sets `backoffLimit`, so it takes the Kubernetes default of 6

https://github.com/terrakube-io/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/scheduler/job/tcl/executor/ephemeral/EphemeralExecutorService.java#L307-L335

The consequence is that a *hung* Terraform process — as opposed to a failed one — runs indefinitely. A wedged provider call, a stuck state lock, or a cloud API that never returns will hold its pod and its plugin-cache mount forever. On a cluster with per-pod billing that is a direct and unbounded cost; on any cluster it is capacity that never comes back without manual intervention.

The `backoffLimit` default compounds it: a step that fails consistently is retried six times before the job is marked failed, which for Terraform is rarely the desired behaviour.

None of the three is reachable from configuration. They are absent from the Java that builds the Job, so no Helm value or Terrakube setting can change them. A namespace `ResourceQuota` is the only lever available today, and it is a blunt one — it caps concurrency, not duration.

### Proposal

Add the three as fields on `EphemeralConfiguration` (`@ConfigurationProperties(prefix = "io.terrakube.executor.ephemeral")`), following the same pattern the multi-secret feature used in #3149:

| Field | Type | Default | Effect |
|---|---|---|---|
| `activeDeadlineSeconds` | `Long` | unset | Kubernetes terminates the job when exceeded |
| `backoffLimit` | `Integer` | unset | retries before the job is marked failed |
| `ttlSecondsAfterFinished` | `Integer` | 30 | unchanged from today, now adjustable |

Defaults preserve current behaviour exactly. `ttlSecondsAfterFinished` keeps its 30. The two new fields default to unset, and the builder omits the corresponding call entirely when they are null — so for anyone who does not opt in, the generated Job spec is byte-identical to what ships today and Kubernetes continues to apply its own `backoffLimit` default of 6.

`activeDeadlineSeconds` is the one that matters; the other two are cheap to include while the same file is open. `ttlSecondsAfterFinished` at 30 seconds is also quite short for debugging a failed run, so making it adjustable has its own small benefit.

### Separately

There appears to be no way to set a `securityContext` on ephemeral pods either. These pods execute user-authored Terraform, so they are arguably the ones that most want `runAsNonRoot` and dropped capabilities. Mentioning it here for visibility rather than folding it into this change — happy to raise it as its own issue if that is preferred.

### Version

Checked against `main` at e7d51f0b. I could not find an existing issue or PR covering this; apologies if I missed one.

putz612:
## Summary

Closes #3541.

Ephemeral Terraform jobs had no wall-clock bound: `activeDeadlineSeconds` was never set on the Kubernetes Job, `backoffLimit` was left at Kubernetes' own default of 6, `ttlSecondsAfterFinished` was hardcoded to 30, and `terminationGracePeriodSeconds` inherited Kubernetes' own default of 30s. A hung Terraform process had no operator-configurable cap, and a pod that did hit a deadline had too little grace period for terraform to release its state lock before SIGKILL.

This is the other half of the chain started in `terraform-spring-boot#101` (merged): that PR makes the library terminate the terraform process cleanly on `close()`/timeout, but nothing in the ephemeral path was triggering that close in a bounded way. Setting `activeDeadlineSeconds` on the Job gives Kubernetes a way to terminate the pod gracefully (`SIGTERM` → Spring shutdown hook → `TerraformClient.close()`), and raising `terminationGracePeriodSeconds` gives that shutdown path enough time to actually run before Kubernetes escalates to `SIGKILL`.

## Changes

- `EphemeralConfiguration`: add `activeDeadlineSeconds`, `backoffLimit`, and `terminationGracePeriodSeconds` (all new; `ttlSecondsAfterFinished` already existed, hardcoded), each bound to the existing `ExecutorEphemeral*` environment variable convention.
- `@Validated` with `@Positive` on `activeDeadlineSeconds` (Kubernetes rejects a Job whose deadline isn't greater than zero) and `@PositiveOrZero` on the other three.
- `terminationGracePeriodSeconds` defaults to 60 rather than inheriting Kubernetes' 30s default — a measured worst-case JVM/library shutdown budget of ~28s leaves too little margin against 30s, and it's still fully configurable.
- `activeDeadlineSeconds` and `backoffLimit` default to unset (today's behaviour: unbounded runtime, Kubernetes' own retry default).

A paired Helm chart PR exposing these as chart values is at terrakube-io/terrakube-helm-chart#297.

## Test plan

- [x] `mvn -pl api -Dtest='Ephemeral*Test' test` — 40/40 pass
- [x] `mvn -pl api test` (full suite) — 1063 tests, 0 failures (the 6 errors are pre-existing Testcontainers "no Docker" errors present on `main` too)
- [x] Manual mutation check on the new default: flipping 60→30 fails exactly the two tests asserting it, nothing else


alfespa17:
A few comments:

### Operational Considerations & Recommendations

#### 1. `backoffLimit` Default Behavior vs. Terraform State Safety (Important)
* **Context**: `backoffLimit` defaults to `null` (unset), and the pod template specifies `restartPolicy: "Never"`.
* **Issue**: When `spec.backoffLimit` is omitted on a Kubernetes `batch/v1` Job, **Kubernetes automatically defaults to `backoffLimit: 6`**.
* **Risk to Terraform**: If an ephemeral step fails mid-execution (e.g., Terraform error, node eviction, OOM, or transient network blip), the Kubernetes Job controller will automatically create a replacement pod to rerun the same job step:
  1. If the previous pod died while holding a state lock, the replacement pod will fail on `Error acquiring the state lock`.
  2. If the previous step partially applied non-idempotent cloud resources, blindly rerunning `terraform apply` can result in duplicate resources, race conditions, or state drift.
  3. Terrakube's own scheduler and reconciliation engine expects job steps to fail deterministically rather than having Kubernetes transparently retry them at the pod level.
* **Recommendation**:
  Consider defaulting `backoffLimit` to `0` (at-most-once step execution) or explicitly documenting in the Helm chart and sample configuration that operators should set `ExecutorEphemeralBackoffLimit=0`.

---

#### 2. Global Timeout vs. Workspace Workload Granularity
* **Context**: `activeDeadlineSeconds` is currently configured exclusively at the cluster/API level via `io.terrakube.executor.ephemeral.activeDeadlineSeconds`.
* **Observation**: In multi-tenant environments, Terraform runs vary significantly by workload:
  - A simple workspace plan or drift detection might take 30–60 seconds.
  - A large workspace apply (e.g., provisioning RDS clusters, Kubernetes clusters, or multi-region VPCs) can legitimately take 45–90 minutes.
  - Setting a global cluster deadline (e.g., 30 minutes) to catch runaway jobs will hard-terminate legitimate long-running applies with `DeadlineExceeded`.
* **Recommendation (Future / Follow-up)**:
  Consider following the pattern used by `EPHEMERAL_CPU_LIMIT` and `EPHEMERAL_CONFIG_NODE_SELECTOR_TAGS` by allowing workspaces to define `EPHEMERAL_ACTIVE_DEADLINE_SECONDS` in workspace environment variables, using the global setting as an optional upper ceiling.

---

#### 3. Kubernetes `activeDeadlineSeconds` Includes Scheduling and Image Pulling
* **Context**: In `application.properties`, the comment states:
  ```properties
  ## Maximum seconds a job's pod may run before Kubernetes terminates it as failed; unset means no deadline (unbounded)
  ```
* **Clarification**: In Kubernetes `batch/v1` Jobs, `activeDeadlineSeconds` is measured relative to the Job's `startTime` (when the Job controller activates it), **not** when the container process starts running.
* **Operational Gotcha**: In clusters using node autoscalers (Karpenter, Cluster Autoscaler) or where large container images (~1–2 GB) must be pulled to fresh nodes, pending and pull times eat into the deadline. If an operator configures an aggressive deadline (e.g. 120s), the Job may fail with `DeadlineExceeded` before the container boots.
* **Recommendation**: Update the comment in `application.properties` to clarify that the deadline includes pod scheduling and image pull duration.

---

#### 4. Forensic Log Retention on Pod Bootstrap Failure
* **Context**: `ttlSecondsAfterFinished` defaults to `30`.
* **Observation**: If a pod fails before the JVM boots (e.g., image pull failure, missing Kubernetes secret in `envFrom`, invalid PVC mount, or low memory limit causing early OOM), no logs or status updates are transmitted to Terrakube API.
* **Impact**: Kubernetes deletes the Job and its Pods exactly 30 seconds after failure, leaving a very tight window for operators to inspect `kubectl describe pod` or `kubectl logs`.
* **Recommendation**: In deployment documentation and Helm values, advise setting `ExecutorEphemeralTtlSecondsAfterFinished` to `300`–`600` (5–10 minutes) if operators need time to troubleshoot failed pod initializations.

---

#### 5. Minor Documentation Typo in `application.properties`
* In `api/src/main/resources/application.properties` and `application-demo.properties`:
  ```properties
  ## Seconds the pod gets to shut down cleanly after SIGTERM before Kubernetes SIGKILLs it; unset means
  ## Kubernetes' own default (30). Raise it if activeDeadlineSeconds is set: ...
  io.terrakube.executor.ephemeral.terminationGracePeriodSeconds=${ExecutorEphemeralTerminationGracePeriodSeconds:60}
  ```
  The comment states *"unset means Kubernetes' own default (30)"*, but the placeholder `${...:60}` and the field default in Java (`60L`) both default to `60`.

putz612:
Thanks for the review, all fair points.

1/3/5 fixed by expanding the doc comments in `application.properties`/`application-demo.properties` (and the matching values in the paired chart PR): `activeDeadlineSeconds` now notes it's counted from Job `startTime` and includes scheduling/image-pull time; `backoffLimit` now documents the retry-into-a-held-lock and duplicate-apply risk and that `0` gives at-most-once execution; `terminationGracePeriodSeconds`'s comment now correctly says 60 instead of the stale 30. Kept the defaults as-is (unset/6, unset/30, 60) rather than changing behavior — figured a documented default beats a silent one, and changing `backoffLimit`'s default to 0 felt like a separate call from making it configurable. Happy to flip it if you'd rather.

4: documented too (`ttlSecondsAfterFinished` comment now calls out the troubleshooting window and suggests 300-600 if you need more).

2: agreed, worth doing, but scoping it as follow-up work rather than folding into this PR.

PR #3565, opened 2026-09-16
feat(jobs): track and display plan approver

### LINKED ISSUE #3554: Track and display user who approved a plan in workspace executions
### Feature description 💡

Currently, when a workspace execution requires plan approval (added as a task to the template), the system does not track or display which user approved the plan before the apply phase. This makes it difficult to audit who authorized infrastructure changes and reduces accountability in team environments.

**Proposed Solution:**
- Add a new field to store the user who approved the plan
- Update the approval workflow to capture and save the user ID/email when a plan is approved
- Display the approver information in the job details UI, similar to how the job creator is currently shown
- Include this information in job history and audit logs

**Benefits:**
- Improved audit trail for infrastructure changes
- Better accountability in team environments
- Enhanced security and compliance tracking
- Clear visibility into who authorized specific changes

**Implementation Notes:**
- Backend changes to capture approver information
- Frontend updates to display approver details in the job details view

### Anything else?

- Consider whether to also track when the approval occurred (timestamp)
- Evaluate if similar tracking is needed for rejected plans
- Ensure the implementation works with existing RBAC permissions for approval

Macbet:
Records who approved a plan and when, then shows that information alongside the run creator in job details and workspace run history. Covers JSON:API/GraphQL status approvals and Terraform CLI apply requests.

Adds nullable `approvedBy` and `approvedAt` fields with a Liquibase migration. The server derives the approver from the authenticated token (email, falling back to subject); clients cannot set or overwrite these fields. Existing approval permissions still apply, and audit messages are emitted only after a successful commit. Repeated CLI requests preserve the original attribution. Historical runs without approval metadata continue to render normally.

Validation:
- `mvn -B -pl api -Dtest=JobTests,RemoteTfeServiceTest test`: 20 tests passed with Java 25.
- Job details and workspace data Jest suites: 30 tests passed.
- `tsc --noEmit` and `vite build`: passed.

Closes #3554.


ui/src/domain/Jobs/Details.tsx
```diff
@@ -765,6 +765,20 @@ export const DetailsJob = ({ jobId }: Props) => {
                     <Avatar size="small" shape="square" icon={<UserOutlined />} />{" "}
                     <b>{job.data.attributes.createdBy}</b> triggered a run from {formatJobVia(job.data.attributes.via)}{" "}
                     {job.data.attributes.createdDate ? relativeTime(job.data.attributes.createdDate) : ""}
+                    {job.data.attributes.approvedBy && (
+                      <span style={{ display: "block" }}>
```
jakegroves, line 769: could this be put in a shared component or helper, the main difference is `span` vs `div` between this and `Workspaces/details.tsx`

api/src/main/java/io/terrakube/api/plugin/security/audit/JobApprovalService.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+@Slf4j
+@Service
+public class JobApprovalService {
+    public void recordApproval(Job job, JwtAuthenticationToken user) {
+        String email = user.getToken().getClaimAsString("email");
```
jakegroves, line 16: we already have an existing helper https://github.com/terrakube-io/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/security/audit/dex/DexAuditorAwareImpl.java 
this is duplicating the helper

can we reuse the existing helper isntead of a second copy of the logic, so we can keep the mechanisms consistent:

https://github.com/terrakube-io/terrakube/blob/main/api/src/main/java/io/terrakube/api/plugin/security/audit/AuditConfig.java#L14

you should be able to do something like:
```
private final AuditorAware<String> auditorProvider;

public void recordApproval(Job job, JwtAuthenticationToken user) {
    String approver = auditorProvider.getCurrentAuditor().orElse(user.getName());
```

api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -1522,6 +1527,12 @@ RunsData runApply(int runId, JwtAuthenticationToken currentUser) {
 
             for (Step step : job.getStep()) {
                 if (step.getStepNumber() == 150) {
+                    if (step.getStatus() == JobStatus.completed) {
+                        break;
```
jakegroves, line 1531: this guard changes existing behaviour, but fixes a potential issue... as it prevent the job's status from moving backwards, and duplication execution

could we add it into the PR descrpition ... as this should be signed off as this is important cheap fix

ui/src/domain/Jobs/Details.tsx
Macbet: Addressed in 02c45bf6: extracted ApprovalAttribution and reused it in job details and workspace details. A block-level span works in both locations. Both relevant Jest suites pass (30 tests).

api/src/main/java/io/terrakube/api/plugin/security/audit/JobApprovalService.java
Macbet: Addressed in 02c45bf6: JobApprovalService now injects the existing AuditorAware<String> provider and falls back to user.getName() only when it returns empty. CLI tests cover both provider attribution and the fallback, including retries. Backend suites pass (22 tests).

api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
Macbet: Updated the PR description to explicitly call out this behavior change: once approval step 150 is completed, a repeated CLI apply request preserves the current job status and skips re-scheduling. Also noted that this guard does not provide locking against concurrent requests.

jakegroves (approved):
Reviewed the latest changes and confirmed the previous feedback has been addressed:
  - Approval attribution UI is now shared between job and workspace views.
  - The implementation reuses the configured `AuditorAware<String>` provider, with an authenticated-user fallback.
  - The Terraform CLI retry guard is documented and prevents completed approvals from resetting a run to `pending`, emitting duplicate notifications, or scheduling duplicate execution.

this is backward-compatible for existing workflows:
  - Existing jobs remain valid because the new database fields are nullable.
  - API changes are additive and approval metadata cannot be forged by clients.
  - Normal first-time Terraform CLI apply behaviour is unchanged.
  - The only behaviour change is the intentional protection against sequential duplicate CLI apply requests.

 Performance impact is minimal: no additional database lookup or query path is introduced, and the attribution fields are written alongside the existing job update

One known limitation remains: simultaneous, concurrent CLI apply requests are not serialized. This is acknowledged in the PR and is outside this change’s scope

**approved**

PR #3566, opened 2026-09-16
fix(api): map workspace.lastJobStatus by name instead of ordinal

Macbet:
## Why

`Workspace.lastJobStatus` is a `JobStatus` enum with no `@Enumerated`, so Hibernate mapped it by ordinal and stored the ordinal as text in the `varchar(36)` column (`"12"` for `NeverExecuted`). Reads survived through JDBC coercion, but any SQL comparison did not: PostgreSQL refuses `varchar = smallint`, so every `lastJobStatus` filter failed with

```
ERROR: operator does not exist: character varying = smallint
STATEMENT: select count(distinct w1_0.id) ... where w1_0.last_job_status in ($1) ...
```

The paginated workspace list (#3514) issues seven facet counts with that filter on every page load, which is how this surfaced. H2 coerces the types, so the existing tests passed.

## What

- `@Enumerated(EnumType.STRING)` on `lastJobStatus`, matching `Job.status` and `policyComplianceStatus`.
- Liquibase changeset `2-34-0-workspace-last-job-status-enum-name` rewrites stored ordinals to names in `JobStatus` declaration order.
- `WorkspaceLastJobStatusChangeSetTest` asserts the changeset's mapping table matches `JobStatus.values()`, so reordering the enum without touching the changeset fails a test instead of silently relabeling workspaces.

## Testing

- `WorkspaceLastJobStatusChangeSetTest` and `WorkspaceGraphQlPaginationTest` (status filters on H2 with the changeset applied) pass.
- Verified the failing query and the `"12"` column values against a PostgreSQL 17 devcontainer database.


jakegroves:
thanks for this, and sorry: I missed this PR and fixed the same issue independently in #3586 (`@Enumerated(EnumType.STRING)` on `Workspace.lastJobStatus`) and #3593 (the ordinal → name changeset). Both are now merged, so the `varchar = smallint` failure is fixed on `main`. Your diagnosis matched exactly.

I diffed the two changesets and they're functionally identical: the same 13 `'0'..'12'` → name updates, in `JobStatus` declaration order. Only the changeset id differs. This PR now conflicts with `main`:

```
CONFLICT (content): api/src/main/resources/db/changelog/changelog.xml
CONFLICT (add/add): api/src/main/resources/db/changelog/local/changelog-2.34.0-workspace-last-status-enum.xml
```

So I'd suggest either closing this PR or rebasing it down to the parts `main` doesn't have yet:

- **`WorkspaceLastJobStatusChangeSetTest`: worth keeping.** `main` only covers the migration end-to-end for one ordinal (`'5'` → `completed` in `PostgreSQLStartupTests`). Your test guards all 13 values against someone reordering `JobStatus`. I ran it unchanged against `main`'s changelog and it passes.
- **The real-Redis change to `JobReconciliationSweepIntegrationTest`: probably better as its own PR.** It passed 3/3 locally on top of `main`, and so did `main`'s current mocked version, so a link to a CI run where it flaked would help make the case. The added `lastJobStatus == completed` assertion is a nice extra.

If you rebase, please drop the `Workspace.java` / `changelog.xml` / changeset changes and keep `main`'s changelog file as-is. Bringing this PR's changeset id across would make Liquibase run the same migration a second time on every database that already has #3593. That's harmless because it's idempotent, but it adds a duplicate changelog entry.

Macbet:
upgraded

PR #3568, opened 2026-09-16
perf(ui): load only the Monaco languages and features the editors use

Macbet:
## Why

`monacoConfig.ts` imported the root `monaco-editor` entry, which registers ~80 language definitions, the CSS/HTML/JSON/TypeScript language services, an LSP client and every editor feature. The app uses the editor in three places (read-only JSON on the workspace States page, YAML for templates, JavaScript for actions) and each only reads and writes text.

## What

The editor is now assembled from `editor.api`, a curated set of editor features and the three languages actually used. The TypeScript language service is dropped on purpose: its worker alone is 1.45 MB gzipped and its diagnostics misfire on the JSX-style action code. Syntax highlighting is unchanged. JSON validation is disabled because JSON is only ever shown read-only.

## Measured (production build)

| | before | after |
|---|---|---|
| CodeEditor chunk, gzipped | 326 KB | 107 KB |
| workers emitted | editor, json, css, html, ts | editor, json |
| JS in `build/assets` on disk | 24.0 MB | 15.0 MB |

## Testing

tsc, eslint and jest pass. Please open a workspace state, a template and an action to confirm highlighting, find (Ctrl+F), folding and the context menu still work.


ui/src/config/monacoConfig.ts
```diff
@@ -1,18 +1,48 @@
...
+import { jsonDefaults } from "monaco-editor/languages/features/json/register.js";
 import editorWorker from "monaco-editor/editor/editor.worker.js?worker";
 import jsonWorker from "monaco-editor/language/json/json.worker.js?worker";
-import tsWorker from "monaco-editor/language/typescript/ts.worker.js?worker";
 import { ThemeMode } from "./themeConfig";
 
 self.MonacoEnvironment = {
   getWorker(_workerId: string, label: string) {
     if (label === "json") return new jsonWorker();
-    if (label === "javascript" || label === "typescript") return new tsWorker();
```
jakegroves, line 11: The JS language service (ts.worker) being dropped for bundle-size reasons makes sense — it's ~1.4MB gzipped and misfires on the JSX-style action code anyway. Good tradeoff

However, dropping it removes the only safety net that caught syntactically invalid Action scripts before save — right now a broken script (unbalanced parens, etc.) saves silently and only fails at execution time, which is a worse debugging experience than what we had

This would be a rgeression, so we probably want a lightweight client-side syntax check on save as a replacement rather than leaving that gap open... something like `acorn.parse()`
As this affects `Actions.tsx`
We could have a small validatore module that parses it:
```
import { Parser } from "acorn";
import jsx from "acorn-jsx";

const jsxParser = Parser.extend(jsx());

export function validateActionSyntax(code: string): string | null {
  try {
    jsxParser.parse(code, { ecmaVersion: "latest", sourceType: "script" });
    return null;
  } catch (err) {
    return err instanceof SyntaxError ? err.message : "Invalid syntax";
  }
}
```
and use this in `Actions.tsx` for `onCreate` or `onUpdate`

PR #3580, opened 2026-09-20
feat(api): support key/value workspace tags in the TFE-compatible API

hirokimatsueda:
## Summary

Workspaces can now carry key/value tags. The Terraform CLI `cloud` block can select them by key/value deterministically:

```hcl
terraform {
  cloud {
    organization = "aws"
    workspaces {
      tags = { env = "dev", app = "myapp" }
    }
  }
}
```

This PR covers the database and the TFE-compatible API (`/remote/tfe/v2/`). Exposing the value through Elide (JSON:API / GraphQL) and the UI will follow in a separate PR.

## Problem

`workspacetag` had no column for a value, and the TFE API did not implement what the CLI relies on for key/value tags:

- The CLI lists workspaces with `filter[tagged][N][key]` / `filter[tagged][N][value]`, repeating the keys in `search[tags]`. Only `search[tags]` was understood, so `env=dev` and `env=prod` workspaces with the same keys both matched. Depending on `.terraform/environment` and `-input`, `terraform init` then picked a different workspace, asked the user to choose, or failed.
- `GET workspaces/{id}/tag-bindings` returned 404, which the CLI treats as "no key/value support" and aborts (`backend does not support key/value tags`).
- Workspaces had no `tag-names` attribute, and tags sent on workspace creation were ignored.

## Changes

**Database**
- `workspacetag.tag_value varchar(255)`, nullable. `NULL` means a key-only tag, so existing rows keep working unchanged. The column is named `tag_value` because `VALUE` is reserved in H2.
- `tag` still holds the organization's keys, and the existing unique constraint on `(tag_id, workspace_id)` keeps one value per key and workspace.

**TFE API**

| Endpoint | Behavior |
|---|---|
| `GET organizations/{org}/workspaces` | `search[tags]`, `filter[tagged]`, `search[name]` and `filter[project][id]` are combined with AND, without duplicates. `filter[tagged]` indexes may arrive in any order and with gaps; an empty value matches any value of the key. Returns `meta.pagination`. With no filter at all, it still returns an empty list, as before. |
| `GET workspaces/{id}/tag-bindings` (new) | Returns every binding. Key-only tags have `value: ""`. Access denied is 403, never 404, so the CLI does not conclude that key/value tags are unsupported. |
| `PATCH workspaces/{id}/tag-bindings` (new) | Upserts by key, like HCP Terraform. Bindings missing from the request are kept, so tags added in the UI survive `terraform init`. |
| Workspace responses | Add `tag-names`, which includes the keys of key/value tags. |
| `POST organizations/{org}/workspaces` | Applies `relationships.tags` and `relationships.tag-bindings` when a new workspace is created. |
| `POST workspaces/{id}/relationships/tags` (legacy) | Still adds key-only tags, and no longer overwrites the value of an existing key/value tag. |

An empty key, a key over 64 characters (`tag.name` is `varchar(64)`) or a value over 255 characters is rejected with 400 instead of failing in the database.

## Testing

- **Unit tests** (`RemoteTfeServiceTest`, `WorkspaceListQueryTest`): filter parsing and matching, the AND of all filters, pagination, upsert semantics (including a key repeated in one request), `tag-names`, creation with tags, legacy tags API, permissions and validation.
- **REST tests** (`TfcApiTests`): a create → tag-bindings → PATCH → filtered list flow using the same payloads as go-tfe, plus 400/403/404 cases; and two workspaces sharing the keys but differing in a value, where filtering by key alone matches both and `filter[tagged]` picks exactly one.
- **Wire compatibility with go-tfe v1.110.0** (the version Terraform uses):
  - `Workspaces.Create` with tags and with tag bindings
  - `Read` (`TagNames`)
  - `List` with `TagBindings` + `Tags`
  - `ListTagBindings`, `AddTagBindings`, `AddTags`
- **Manual `terraform init` with Terraform 1.15.3**, against `myapp-dev` / `myapp-prod` / `myapp-stg`, which all have `app=myapp` and differ only in `env`:
  - `tags = { env = "dev", app = "myapp" }` selects `myapp-dev` every time, including in fresh directories and with `-input=false`.
  - A second `init` sends no PATCH (`does not require tag update`).
  - Switching the config to `env = "prod"` moves to `myapp-prod`.
  - `TF_WORKSPACE` pointing at a workspace outside the tags fails with `Invalid workspace selection`.
  - With no match, a workspace created from the prompt gets its tags at creation.
- `mvn verify`: all related tests pass.

## Known limitations

- If two requests create the same new tag key for the organization at the same time, one fails with a 500. The unique constraint keeps the data consistent, so the failed `terraform init` only needs to be re-run. This comes from the existing `searchOrCreateTagOrganization`, whose retry-on-failure never runs because the insert only fails at flush or commit.
- Tag filtering in the workspace list is done in memory, as the existing `search[tags]` filter was, and every match is returned on the first page.
- Out of scope: `effective-tag-bindings` (tags inherited from projects), `search[exclude-tags]`, and policies attached to a key/value pair (policies still attach to the key).

Implemented with assistance from Claude Code. Design decisions, review and the manual `terraform init` verification were done by me.


api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -627,52 +649,262 @@ StateConsumerList getWorkspaceStateConsumers(String workspaceId, JwtAuthenticati
...
+        List<Workspace> matches = candidates.stream()
+                .filter(workspace -> matchesProject(workspace, query.projectId()))
+                .filter(workspace -> matchesTags(workspace, query, organizationTagNames))
+                .filter(workspace -> seen.add(workspace.getId()))
+                .sorted(Comparator.comparing(Workspace::getName))
+                .toList();
+        log.info("Workspaces matching tags {} tag bindings {} project {}: {}", query.searchTags(), query.tagged(),
+                query.projectId(), matches.size());
+
+        // Every match fits in the first page; the CLI stops paging once current-page reaches total-pages
```
jakegroves: You are deliberately returning every matching workspace on page 1 and reports `total-pages: 1`, but the TFE workspace-list API supports `page[number]` and `page[size]` (defaulting to 20). `WorkspaceListQuery` currently parses only`page[number]`, so a broad tag filter can return an unbounded response; requesting page 2 then returns no results.

Please parse `page[size]`, apply it to `matches` before populating `data`, and calculate the pagination metadata from the full match count

Can we remediate this rather than emit misleading pagination metadata?

See: https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#list-workspaces

api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -126,6 +129,9 @@ public class RemoteTfeService {
     private RbacService rbacService;
 
     private static final String LATEST_TERRAFORM_VERSION = "latest";
+    // tag.name is varchar(64); HCP Terraform allows 255 characters for values
```
jakegroves: https://developer.hashicorp.com/terraform/cloud-docs/api-docs/workspaces#add-update-tag-bindings-on-a-workspace

TFE specifies:
tag keys = `128` chracaters
tag values = `256` characters

and up to 10 direct tags, can you make it match the TFE api please and correct the db migration also

api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
```diff
@@ -627,52 +649,262 @@ StateConsumerList getWorkspaceStateConsumers(String workspaceId, JwtAuthenticati
...
+            log.warn("User does not have permission to update tags for workspace {}", workspace.get().getName());
+            throw new org.springframework.security.access.AccessDeniedException(
+                    "User does not have permission to manage tags for this workspace");
+        }
+
+        upsertTagBindings(workspace.get(), tagBindingList);
+        return toTagBindingList(workspaceTagRepository.findByWorkspace(workspace.get()));
+    }
+
+    private void upsertTagBindings(Workspace workspace, TagBindingList tagBindingList) {
```
jakegroves, line 831: TFE permits up to 10 direct tag bindings per workspace, but this PATCH path currently permits an unbounded number. Could we calculate the effective set of keys (`existing keys U requested keys`) before persisting and return a JSON:API 400 when it would exceed 10?

This needs to account for PATCH updating an existing key, rather than only checking `bindings.size()`

jakegroves (changes requested):
Nice implementation overall—the nullable migration approach is safe for existing key-only tags, and the parsing/upsert behaviour is thoughtfully covered by tests.

I’ve left three compatibility comments around workspace-list pagination and tag-binding limits. Once those are addressed then i'm happy to approve

api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
hirokimatsueda: Good point — fixed. The list now honours `page[size]` (default 20, max 100) and `page[number]`, returns only the requested page, and computes the pagination metadata from the full match count. I also checked with Terraform 1.15.3 that the CLI follows `next-page` across two pages.

api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
hirokimatsueda: Updated to match TFE: keys up to 128 characters, values up to 256. The migration now creates `tag_value` as `varchar(256)` and widens `tag.name` to `varchar(128)`; I checked the widening on PostgreSQL and SQL Server with the existing unique constraint in place.

api/src/main/java/io/terrakube/api/plugin/state/RemoteTfeService.java
hirokimatsueda: Makes sense — done. PATCH and workspace creation now count `existing ∪ requested` keys and return a JSON:API 400 above 10, so updating an existing key doesn't count twice.

One judgement call: workspaces that already have more than 10 tags (possible via the UI today) can still update their existing keys, so `terraform init` doesn't start failing for them — only adding new keys beyond the limit is rejected. The legacy `POST relationships/tags` endpoint is left unchanged. Happy to tighten either if you'd prefer.

hirokimatsueda:
Thanks for the thorough review, @jakegroves! I've addressed all three points in 8d487391 and merged the latest `main` to resolve the conflicts. Short notes are in each thread.

jakegroves (approved):

